PR #3, opened 2018-11-02
Trm 20180 query string

xwatkins:
Added support for query string generation as well as test cases.


dlrice:
I made a few commits with minor edits. If these look sensible to you, I am happy to merge.

xwatkins:
In the case the provided strings are undefined, I'm wondering if it would be best to either:
- return an empty string for the subquery
- or throw an exception
as the queries wouldn't be valid?

xwatkins:
Added exception throwing. Please merge if ready @dlrice

PR #7, opened 2018-12-03
Trm 21141 handle search button click

dlrice:
* Create `utils/utils.js` with functions: `serializableDeepCopy`, `serializableDeepEquals`, `removeProperty`.
* Create tests for `utils/utils.js`
* Within ResultsContainer and in order of precedence:
   * Check for presence of user-touched query clauses in props and if present fetch results from these
   * Check for presence of query in URL and if present fetch results from these
   * If nothing, run a query on all
* Handle AdvancedSearch search click and navigate to ResultsPage and pass a deep-copy of query clauses list as a prop


src/advanced-search/__tests__/AdvancedSearchContainer.spec.jsx
```diff
@@ -44,10 +44,10 @@ describe('AdvancedSearch component', () => {
     expect(props.dispatchfetchEvidencesIfNeeded).toHaveBeenCalledWith('annotation');
   });
 
-  test('should submit a query', () => {
-    wrapper.find('#submit-query').simulate('click');
-    expect(props.dispatchSubmitQuery).toHaveBeenCalled();
-  });
+  // test('should submit a query', () => {
```
xwatkins: You could test it by creating the mock below and checking it was called:
`const historyMock = { push: jest.fn() };`

src/results/ResultsContainer.jsx
```diff
@@ -1,31 +1,55 @@
...
+import { serializableDeepCopy } from '../utils/utils';
+import { isClauseTouched } from '../advanced-search/utils/clause';
 
 export class Results extends Component {
   componentDidMount() {
-    const { query, dispatchFetchResults } = this.props;
-    dispatchFetchResults(query);
+    console.log('here');
+    const { queryClauses: queryClausesFromProps, dispatchFetchResults, location } = this.props;
+    const touchedQueryClausesFromProps = queryClausesFromProps.filter(isClauseTouched);
```
xwatkins: Likewise, maybe 'changedQueryClausesFromProps'?

src/state/reducers.js
```diff
@@ -0,0 +1,40 @@
+import { combineReducers } from 'redux';
+import query from '../advanced-search/state/reducers';
+import search from '../results/state/reducers';
+// import { COPY_QUERY_CLAUSES_OBJECT_TO_SEARCH } from './actions';
```
xwatkins: Do you want to keep a reference to these or can we remove them?

src/advanced-search/utils/clause.js
```diff
@@ -14,4 +15,9 @@ const createEmptyClause = () => ({
   queryInput: {},
 });
 
-export default createEmptyClause;
+export const isClauseTouched = (clause) => {
```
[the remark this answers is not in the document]
dlrice: Logic refactored so that detecting if the user has touched a clause is no longer needed.

src/results/ResultsContainer.jsx
dlrice: Logic refactored so that detecting if the user has touched a clause is no longer needed.

src/state/reducers.js
dlrice: I used them in the end.

dlrice:
Fixed all comments. Something that wasn't mentioned before: this branch's results page has an advanced search element above it.

PR #8, opened 2018-12-05
Trm 21051 default columns

xwatkins:
Adding basic default views for Results table columns.

### DIFF
--- a/src/results/views/FieldToViewMappings.jsx
+++ b/src/results/views/FieldToViewMappings.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import SimpleView from './SimpleView';
+import NameView from './NameView';
+
+const FieldToViewMappings = {
+  accession: row => <SimpleView termValue={row.accession} />,
+  id: row => <SimpleView termValue={row.id} />,
+  protein_name: (row) => {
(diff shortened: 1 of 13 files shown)


src/results/views/FieldToViewMappings.jsx
```diff
@@ -0,0 +1,39 @@
...
+    if (row.protein.recommendedName.ecNumber) {
+      alternativeNames.push(row.protein.recommendedName.ecNumber.value);
+    }
+    if (row.protein.recommendedName.alternativeName) {
+      alternativeNames.push(row.protein.recommendedName.alternativeName.value);
+    }
+    const props = {
+      name: row.protein.recommendedName.fullName
+        ? row.protein.recommendedName.fullName.value
+        : undefined,
```
dlrice: would `name: row.protein.recommendedName.fullName.value` be equivalent?

src/results/views/NameView.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+const NameView = (props) => {
```
dlrice: Could do the props destructing in the function argument:
```const NameView = ({ name, shortName, alternativeNames }) => {```

PR #12, opened 2018-12-13
Trm 21234 quicksearch

xwatkins:


xwatkins:
Changes done as requested

PR #13, opened 2018-12-19
Use DataTable from franklin; allow rows to be selectable and store st…

dlrice:
…ate in ResultsContainer.

Will need to use franklin-sites version which incorporates https://github.com/ebi-uniprot/franklin-sites/pull/8


dlrice:
Now getting columns from state and also added the organisms render function to `FieldToViewMappings.jsx`

PR #14, opened 2018-12-21
Trm 21143 facet list

xwatkins:
Displays the list of facets for a given results set:
- backend is currently not sending back all labels. creating a Jira for this
- Currenlty only possible to add filters, not remove them until we have the query string parsing - see TODO in `state/reducers.js`


src/results/state/reducers.js
```diff
@@ -11,9 +11,26 @@ const results = (state = [], action) => {
...
         ...state,
         results: action.data.results,
+        facets: action.data.facets,
         lastUpdated: action.receivedAt,
         isFetching: false,
       };
+    case TOGGLE_FACET: {
+      const { selectedFacets } = state;
+      if (!(action.facetName in selectedFacets)) {
+        selectedFacets[action.facetName] = [...Array(1)].map(() => action.facetValue);
```
dlrice: Could `[...Array(1)].map(() => action.facetValue)` become `[action.facetValue]` or `[serializableDeepCopy(action.facetValue)]`?

PR #16, opened 2019-02-12
Model update

xwatkins:
These are model updates to fit the new backend. They haven't switched to `wwwdev` yet so in the meantime I've hardcoded the url in `apiUrls`


src/results/views/__tests__/__snapshots__/FieldToViewMappings.spec.tsx.snap
```diff
@@ -34,17 +34,8 @@ exports[`FieldToViewMappings should render gene_names 1`] = `
...
-  My prot
    
-  <span>
-    <span>
-       
-      ec1234
-    </span>
-  </span>
 </div>"
 `;
```
dlrice: Seems to lack an opening `<strong>` tag and has no content.

PR #17, opened 2019-02-12
Page intro

xwatkins:
Use the page intro pattern, creating a generic system which allows us to define the content based on namespace.

### DIFF
--- a/src/info/InfoMappings.tsx
+++ b/src/info/InfoMappings.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { Namespace } from '../search/types/searchTypes';
+import UniProtKBInfo from './UniProtKBInfo';
+
+const infoMappings: {
+  [index in Namespace]: {
+    name: string;
+    info: JSX.Element;
+    links: Array<{ title: string; destination: string }>;
+  }
+} = {
+  uniprotkb: {
+    name: 'UniProtKB',
+    info: <UniProtKBInfo />,
+    links: [
+      { title: 'Help', destination: '' },
+      { title: 'Video', destination: '' },
+    ],
+  },
+};
+
+export default infoMappings;
--- a/src/layout/SideBarLayout.tsx
+++ b/src/layout/SideBarLayout.tsx
@@ -4,13 +4,15 @@ import BaseLayout from './BaseLayout';
 import './styles/SideBarLayout.scss';
 
 type SideBarLayoutProps = {
+  title: JSX.Element;
   sidebar: JSX.Element;
   content: JSX.Element;
 };
 
-const SideBarLayout: React.FC<SideBarLayoutProps> = ({ sidebar, content }) => (
+const SideBarLayout: React.FC<SideBarLayoutProps> = ({ title, sidebar, content }) => (
   <BaseLayout>
     <section id="sidebar-layout" data-layout="left-sidebar-layout">
+      <section id="sidebar-layout__title">{title}</section>
       <section id="sidebar-layout__sidebar">{sidebar}</section>
       <section id="sidebar-layout__content">{content}</section>
     </section>
--- a/src/layout/SideBarLayout.tsx
+++ b/src/layout/SideBarLayout.tsx
@@ -9,12 +9,16 @@ type SideBarLayoutProps = {
   content: JSX.Element;
 };
 
-const SideBarLayout: React.FC<SideBarLayoutProps> = ({ title, sidebar, content }) => (
+const SideBarLayout: React.FC<SideBarLayoutProps> = ({
+  title,
+  sidebar,
+  content,
+}) => (
   <BaseLayout>
-    <section id="sidebar-layout" data-layout="left-sidebar-layout">
-      <section id="sidebar-layout__title">{title}</section>
-      <section id="sidebar-layout__sidebar">{sidebar}</section>
-      <section id="sidebar-layout__content">{content}</section>
+    <section className="sidebar-layout" data-layout="left-sidebar-layout">
+      <section className="sidebar-layout__title">{title}</section>
+      <section className="sidebar-layout__sidebar">{sidebar}</section>
+      <section className="sidebar-layout__content">{content}</section>
     </section>
   </BaseLayout>
 );
--- a/src/layout/styles/SideBarLayout.scss
+++ b/src/layout/styles/SideBarLayout.scss
@@ -1,7 +1,7 @@
 @import '../../styles/settings';
 
 #sidebar-layout {
-  padding-left: 0.5em;
+  padding-left: 0.5rem;
   display: grid;
   grid-gap: $gutter-size;
   grid-template-columns: 1fr 6fr;
--- a/src/layout/styles/SideBarLayout.scss
+++ b/src/layout/styles/SideBarLayout.scss
@@ -1,11 +1,16 @@
-@import "../../styles/settings";
+@import '../../styles/settings';
 
 #sidebar-layout {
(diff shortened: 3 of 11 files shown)


src/info/InfoMappings.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { Namespace } from '../search/types/searchTypes';
+import UniProtKBInfo from './UniProtKBInfo';
+
+const infoMappings: {
+  [index in Namespace]: {
+    name: string;
+    info: JSX.Element;
+    links: Array<{ title: string; destination: string }>;
```
its-mahdi: I don't have a solid proof for this, but I think the so called angle-bracket type definitions are discouraged in type script e.g.  `Array<...>`

I'm using a TSLint configuration called [Airbnb TS config](https://github.com/progre/tslint-config-airbnb/#readme), but it's not anything official from Airbnb, but it does have a good mix of community accepted rules. I think Microsoft coding-styles also discourages this type of type definitions, except for Generic and Iterator types.

I think it's worth look into this more in depth.

src/layout/styles/SideBarLayout.scss
```diff
@@ -1,11 +1,16 @@
-@import "../../styles/settings";
+@import '../../styles/settings';
 
 #sidebar-layout {
   padding-left: 0.5em;
```
its-mahdi: The `em` units are inherited and relative to the parent. This can cause styling issues when a component is nested inside itself or another component. This is a very controversial topic, but I suggest using `rem` instead of `em` as it is only relative to the root element of the page.

src/layout/SideBarLayout.tsx
```diff
@@ -4,13 +4,15 @@ import BaseLayout from './BaseLayout';
...
 type SideBarLayoutProps = {
+  title: JSX.Element;
   sidebar: JSX.Element;
   content: JSX.Element;
 };
 
-const SideBarLayout: React.FC<SideBarLayoutProps> = ({ sidebar, content }) => (
+const SideBarLayout: React.FC<SideBarLayoutProps> = ({ title, sidebar, content }) => (
   <BaseLayout>
     <section id="sidebar-layout" data-layout="left-sidebar-layout">
```
its-mahdi: The BEM guideline explicitly bans the usage of any IDs or Tag names for the CSS identifiers e.g. `id="sidebar-layout"` should be changed to `className="sidebar-layout"` and the related CSS code should change too.

src/results/ResultsContainer.tsx
```diff
@@ -20,6 +20,7 @@ import {
 
 interface IResultsProps extends RouteComponentProps {
```
its-mahdi: There are a few old-ish coding conventions that still prefix interfaces with `I` but these days it is considered a form of [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) and it is discouraged by many. This one is also slightly controversial, but I have seen examples of `ResultsPropsInterface` or `ResultsPropsType` more often than not -- in TypeScript interfaces are basically types.

src/results/ResultsContainer.tsx
```diff
@@ -120,38 +121,48 @@ export class Results extends Component<IResultsProps, ResultsContainerState> {
...
       dispatchRemoveFacet,
       dispatchUpdateColumnSort,
+      namespace,
       sort,
     } = this.props;
     const { selectedRows } = this.state;
     if (isFetching) {
       return <h3>Loading...</h3>;
     }
+    const info = infoMappings[namespace];
```
its-mahdi: Here maybe `info` can be further destructed to `{ name, links }`. If we stick to this pattern it will help with more complicated object, etc.

src/results/ResultsContainer.tsx
```diff
@@ -120,38 +121,48 @@ export class Results extends Component<IResultsProps, ResultsContainerState> {
...
-            selectedRows={selectedRows}
-            handleHeaderClick={dispatchUpdateColumnSort}
-            sort={sort}
-          />
-        }
-      />
+      <Fragment>
+        <SideBarLayout
+          title={
+            <PageIntro title={info.name} links={info.links}>
```
its-mahdi: I think it will be good to have further discussion about passing children via props like this. I think the community convention here is to prepare the child element in the function scope and then pass it like `<PageIntroElement />` rather creating the child element inline when it is passed as a prop.

src/info/InfoMappings.tsx
xwatkins: I *think* they are only discouraged when it's a single type, e.g. `Array<string>` instead of `string[]`.

src/layout/styles/SideBarLayout.scss
xwatkins: True

src/layout/SideBarLayout.tsx
xwatkins: Woops!

src/results/ResultsContainer.tsx
xwatkins: Renamed

src/search/types/searchTypes.ts
```diff
@@ -1,8 +1,5 @@
 export enum Namespace {
```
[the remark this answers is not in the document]
xwatkins: Tricky... I named them Namespaces as this is what they are referenced as in the current codebase, but that obviously doesn't mean it's right. They represent any of the resources we have (UniProtKB, UniRef, UniParc, proteomes etc...) but also tools (BLAST, Align, etc...). Any ideas?

src/results/ResultsContainer.tsx
xwatkins: true, done

src/results/ResultsContainer.tsx
xwatkins: Let's have a chat

src/search/types/searchTypes.ts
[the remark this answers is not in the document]
its-mahdi: Maybe something like 'Database' or 'Datastore'?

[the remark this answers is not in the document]
xwatkins: But what about tools?

src/info/InfoMappings.tsx
its-mahdi: Maybe we should dig deeper into this and have a chat about it later. To me, the TypeScript convention seems to be:

    type Link = { title: string; destination: string };
    links: Link[];    // and later in your code...

src/results/ResultsContainer.tsx
```diff
@@ -128,13 +128,13 @@ export class Results extends Component<ResultsProps, ResultsContainerState> {
     if (isFetching) {
       return <h3>Loading...</h3>;
     }
-    const info = infoMappings[namespace];
+    const { name, links, info } = infoMappings[namespace];
```
its-mahdi: I think instead of:

    const { name, links, info } = infoMappings[namespace];

There should be either:

     const { info } = infoMappings[namespace];
     const { name, links } = info;

or:

     const { name, links } = infoMappings[namespace].info;

src/search/types/searchTypes.ts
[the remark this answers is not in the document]
its-mahdi: I think we may need to have a chat with others as well to come up with something good.

src/results/ResultsContainer.tsx
xwatkins: No it's actually fine - the object returned by `infoMappings[namespace]` contains ` name, links, info`, see `infoMappings.tsx`

PR #18, opened 2019-02-12
Add root reducer to reset state when results component is unmounted

xwatkins:
Reset the query (actually the whole state at the moment) when the user navigates away from the results page (unmount)

### DIFF
--- a/src/state/reducers.ts
+++ b/src/state/reducers.ts
@@ -1,10 +1,27 @@
-import {
-  combineReducers, Action, Reducer, ReducersMapObject,
-} from 'redux';
-import query from '../search/state/reducers';
-import results from '../results/state/reducers';
+import { combineReducers, Action, Reducer, ReducersMapObject } from 'redux';
+import query, { SearchAction } from '../search/state/reducers';
+import results, { ResultAction } from '../results/state/reducers';
+import { SearchState } from '../search/state/initialState';
+import { ResultsState } from '../results/state/initialState';
 
-export default combineReducers({
+type RootState =
+  | {
+      query: SearchState;
+      results: ResultsState;
+    }
+  | undefined;
+
+const appReducer = combineReducers({
   query,
   results,
 });
+
+const rootReducer = (state: RootState, action: SearchAction | ResultAction) => {
+  if (action.type === 'RESET') {
+    state = undefined;
+  }
+
+  return appReducer(state, action);
+};
+
+export default rootReducer;
--- a/src/state/reducers.ts
+++ b/src/state/reducers.ts
@@ -4,19 +4,20 @@ import results, { ResultAction } from '../results/state/reducers';
 import { SearchState } from '../search/state/initialState';
 import { ResultsState } from '../results/state/initialState';
 
-type RootState =
-  | {
-      query: SearchState;
-      results: ResultsState;
-    }
-  | undefined;
+type RootState = {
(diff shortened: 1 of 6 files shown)


src/state/reducers.ts
```diff
@@ -1,10 +1,27 @@
...
+      results: ResultsState;
+    }
+  | undefined;
+
+const appReducer = combineReducers({
   query,
   results,
 });
+
+const rootReducer = (state: RootState, action: SearchAction | ResultAction) => {
```
dlrice: Rather than wrapping the combined reducers (`appReducer`), we can pick up `action.type === 'RESET'` in both the search and results reducers instead.

src/state/reducers.ts
```diff
@@ -1,10 +1,27 @@
...
+  if (action.type === 'RESET') {
+    state = undefined;
```
dlrice: `null` is best used when we want to set a variable to no value

src/state/reducers.ts
```diff
@@ -1,10 +1,27 @@
...
-import results from '../results/state/reducers';
+import { combineReducers, Action, Reducer, ReducersMapObject } from 'redux';
+import query, { SearchAction } from '../search/state/reducers';
+import results, { ResultAction } from '../results/state/reducers';
+import { SearchState } from '../search/state/initialState';
+import { ResultsState } from '../results/state/initialState';
 
-export default combineReducers({
+type RootState =
+  | {
```
dlrice: is this extra `|` there on purpose?

src/state/reducers.ts
xwatkins: Isn't it better to have a single place to do this though, as it affects both parts of the state (and might in the future be needed to reset other parts of the state)?

src/state/reducers.ts
xwatkins: The Dan Abramov explanation for doing this is with `undefined`, presumably because the variable is declared:
https://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store/35641992#35641992

src/state/reducers.ts
xwatkins: For some reason, Prettier adds this, trying to find a workaround....

xwatkins: It was in the wrong place.

src/state/reducers.ts
dlrice: I suppose that depends if we would ever want some specificity in the future - will always want all parts of the app state reset or will there be a time one just some subset reset?

xwatkins: We might not want to always reset everything, which is why it's nice to keep our options open.

src/state/reducers.ts
dlrice: Defining something as undefined seems counterintuitive to me but I'm sure it's all legit if Dan says it's ok.

src/state/reducers.ts
dlrice: just want to clarify: you're saying leave it as is for now (reset all) and in the future if needed we will replace this PR's code with a state slice specific reset (ie RESET is caught in each of the separate reducers)?

xwatkins: No, I mean that in the future we are likely to have more/different RESETS across separate parts of the state so it's better to keep it in a cross slice reducer

PR #21, opened 2019-03-01
Adding FUNCTION section with basic evidence tags

xwatkins:
- Define the basis for evidence tags, mapping ECO codes to definitions as per old website.
- Add Catalytic activity (will need visualisation added in the future)
- Add Pathway section


src/model/CatalyticActivity.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return null;
+  }
+  return (
+    <Fragment>
+      <h4>Catalytic Activity</h4>
+      {catalyticActivityData.map(catalyticActivity => (
+        <p key={`catalytic-${catalyticActivity.reaction.ecNumber}`}>
+          <strong>{catalyticActivity.reaction.ecNumber}</strong>{' '}
+          {catalyticActivity.reaction.name}
+          {catalyticActivity.reaction.evidences.map(evidence => (
```
dlrice: This will be coming from the server - do we want to do any checking to make sure that all of the keys in this path exist?

PR #23, opened 2019-03-06
Trm 21654 autocomplete without choice fix (MERGE THIS ONE FIRST PLZ)

dlrice:
Update input value as user types on fields with the suggester API so user doesn't have to press enter.

### DIFF
--- a/src/components/AutocompleteWrapper.tsx
+++ b/src/components/AutocompleteWrapper.tsx
@@ -26,7 +26,7 @@ type State = {
   data: Array<Suggestions>;
 };
 
-type Chosen = {
+type SelectedValue = {
   id: string;
   itemLabel: string;
   pathLabel: string;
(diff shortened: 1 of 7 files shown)


src/components/AutocompleteWrapper.tsx
```diff
@@ -54,16 +55,20 @@ class AutocompleteWrapper extends Component<Props, State> {
...
-          id: v1()
+          id: v1(),
         }))
       )
       .then(data => this.setState({ data }))
       .catch(e => console.error(e));
   };
 
-  handleSelect = (chosen: Chosen) => {
+  handleSelect = (chosen: Chosen | string) => {
```
xwatkins: I know it's not part of the PR, but maybe the type name `chosen` is a bit ambiguous? `selectedValue` maybe?

PR #24, opened 2019-03-08
Fix null types

xwatkins:


src/model/FreeText.tsx
```diff
@@ -55,7 +55,7 @@ export const FreeText: React.FC<FreeTextDataProps> = ({
   type,
   includeTitle = false,
 }) => {
-  if (!data.comments) {
+  if (!data.comments || !data.comments.find(d => d.commentType === type)) {
```
dlrice: `some` might fit a better here.

src/model/CatalyticActivity.tsx
```diff
@@ -24,11 +24,14 @@ type CatalyticActivityProps = {
 export const CatalyticActivity: React.FC<CatalyticActivityProps> = ({
   data,
 }) => {
-  if (!data.comments) {
+  if (
+    !data.comments ||
+    !data.comments.find(d => d.commentType === FreeTextType.CATALYTIC_ACTIVITY)
```
dlrice: `some` might fit a better here.

dlrice:
Could be worth adding some model data (`src/model/__tests__/modelData.json`) that has some missing information to make sure it's failing gracefully / rendering nothing (ie no unnecessary commas).

xwatkins:
Using `some`, made a Jira for test cases

PR #26, opened 2019-03-22
Trm 21676/add sequence section

xwatkins:
Adding the sequence section to display the canonical and isoform sequences.
- canonical sequence is displayed by default
- isoform sequences are loaded on demand (see react hook usage)

Caveats:
- need to discuss styling with UX (button level definition, potentially horizontal `InfoList` component)
- enzyme doesn't support `hooks` (yes?) so the interaction with button is not tested...
- I'm wondering about having the entry type definition in a separate file reused by all viewers.


src/model/SequenceView.tsx
```diff
@@ -0,0 +1,222 @@
...
+      molWeight: number;
+      crc64: string;
+    };
+    comments?: AlternativeProduct[];
+  };
+};
+
+export const SequenceInfo: React.FC<{
+  isoformId: string;
+  isoformSequence?: SequenceViewProps['data']['sequence'];
```
dlrice: Could be read easier if `sequence` was it's on type and is used to define `SequenceInfo`'s `isoformSequence` and just added to `SequenceViewProps` ie:
```
SequenceViewProps = {
...
sequence?: sequenceType;
...
```

src/model/SequenceView.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const [data, setData] = useState(null);
+  const [isoformToFetch, setIsoformToFetch] = useState('');
+
+  useEffect(() => {
+    if (!isoformToFetch) {
+      return;
+    }
+    const fetchIsoformData = async () => {
+      const result = await fetchData(`${apiUrls.entry(isoformToFetch)}`);
+      setData(result.data.sequence);
```
dlrice: Would check for existence of the path `result.data.sequence` before accessing it.

src/model/SequenceView.tsx
```diff
@@ -0,0 +1,222 @@
...
+}) => {
+  const infoListData = [
+    {
+      title: 'Name',
+      content: isoformData.name.value,
+    },
+    {
+      title: 'Synonyms',
+      content:
+        isoformData.synonyms &&
```
dlrice: Could use `idx` since it's already imported.

src/model/SequenceView.tsx
```diff
@@ -0,0 +1,222 @@
...
+        isoformData.synonyms.map(syn => syn.value).join(', '),
+    },
+    {
+      title: 'Note',
+      content:
+        isoformData.note &&
+        isoformData.note.texts.map(note => note.value).join(', '),
```
dlrice: Could use `idx` since it's already imported.

src/model/SequenceView.tsx
```diff
@@ -0,0 +1,222 @@
...
+  }
+  return (
+    <Fragment>
+      <InfoList infoData={sequenceInfoData} />
+      <p>
+        This entry describes{' '}
+        <strong>{alternativeProducts.isoforms.length}</strong> isoforms produced
+        by <strong>{alternativeProducts.events.join(' & ')}</strong>.
+      </p>
+      <p>{alternativeProducts.note.texts.map(text => text.value).join(' ')}</p>
```
dlrice: Need some existence checks here in the path `alternativeProducts.note.texts`. An error can be observed at entry `Q9HC29`.

src/model/SequenceView.tsx
```diff
@@ -0,0 +1,222 @@
...
+      {alternativeProducts.isoforms.map(isoform => {
+        let isoformSequence, lastUpdateDate;
+        // Canonical
+        if (isoform.isoformSequenceStatus === 'Displayed') {
+          isoformSequence = data.sequence;
+          if (data.entryAudit) {
+            lastUpdateDate = `${data.entryAudit.lastSequenceUpdateDate} v${
```
dlrice: Is all of `data.entryAudit.lastSequenceUpdateDate` guaranteed to exist?

src/model/SequenceView.tsx
xwatkins: > If an intermediate property is either `null` or `undefined`, it is instead returned. The purpose of this function is to simplify extracting properties from a chain of maybe-typed properties.

We need to check that the map is not `null` or `undefined` anyway so I don't think `idx` gives us anything here?

src/model/SequenceView.tsx
xwatkins: Same as above

src/model/SequenceView.tsx
xwatkins: Done

src/model/SequenceView.tsx
xwatkins: Good spot. Sorted

src/model/SequenceView.tsx
xwatkins: It is if `entryAudit` is present. I've changed the type def to reflect this.

src/model/SequenceView.tsx
dlrice: `isoformSequence?: SequenceViewProps['data']['sequence'];` can use the new type you've made to become `isoformSequence?: Sequence;`

xwatkins: d'oh! done.

PR #29, opened 2019-04-10
Xref trm 21745

dlrice:
* Maps used throughout as these data structures allow enum types to be used as keys.
* Uses Franklin's ExternalLink component from `v0.0.3-dev.32`
* `databaseToDatabaseInfo.json` is a copy of backend's `drlineconfiguration.json` which contains the information needed to create hrefs to the database resource. It is assumed this will eventually become an endpoint. `DatabaseToDatabaseInfo` defines the interface for this JSON file.


src/model/XRef.tsx
```diff
@@ -0,0 +1,207 @@
...
+};
+
+const XRefCategoryTable: React.FC<XRefCategoryTableProps> = ({
+  databaseCategory,
+  databases,
+  xRefData,
+  accession,
+}) => {
+  const databaseCategoryString = databaseCategoryToString.get(databaseCategory);
+  const title = databaseCategoryString && <h4>{databaseCategoryString}</h4>;
```
its-mahdi: `title` might be `undefined` in some cases. Better to have this as a ternary with a `null` in case `databaseCategoryString` didn't have a truthy value.

src/data/DatabaseCategoryToDatabases.ts
```diff
@@ -0,0 +1,194 @@
+import DatabaseCategory from './DatabaseCategory';
+import Database from './Database';
+
+const databaseCategoryToDatabases = new Map<DatabaseCategory, Database[]>([
```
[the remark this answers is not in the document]
dlrice: Yes, it looks much nicer.

src/data/EntrySectionToDatabaseCategories.ts
```diff
@@ -0,0 +1,60 @@
+import DatabaseCategory from './DatabaseCategory';
+import EntrySection from './EntrySection';
+
+const databaseCategoryToDatabases = new Map<EntrySection, DatabaseCategory[]>([
```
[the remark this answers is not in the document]
dlrice: Yes, it looks much nicer.

src/model/XRef.tsx
dlrice: Short-circuit evaluation seems to be a common pattern in React and keeps things concise. How about putting the evaluation in the return - would this seem more natural to you?

```
<div>
      <hr />
      {databaseCategoryString && <h4>{databaseCategoryString}</h4>}
      {databases.sort().map((database: string) => (
...
```

its-mahdi: Yes, that looks better and cleaner!

dlrice: Done.

PR #31, opened 2019-04-16
Entry expression trm 21672

dlrice:
Create an Expression section on the entry view consisting of two simple components:
* TissueSpeficity
* Induction

Test coverage is low because `modelData.json` is missing TissueSpeficity and Induction comment types. Creating a separate Jira to update this file as it is maintained by backend and we want to keep in sync with their test files.

### DIFF
--- a/src/model/Induction.tsx
+++ b/src/model/Induction.tsx
@@ -1,0 +1,62 @@
+import React, { Fragment } from 'react';
+import v1 from 'uuid';
+import UniProtEvidenceTag from '../components/UniProtEvidenceTag';
+import { EvidenceType } from './types/modelTypes';
+import { FreeTextType } from './FreeText';
+
+type Comments = {
+  commentType: FreeTextType;
+  texts: [
+    {
+      value: string;
+      evidences?: EvidenceType[];
+    }
+  ];
+};
+
+type InductionData = {
+  comments?: [Comments];
+};
+
+type InductionProps = {
+  data: InductionData;
+};
+
+export const Induction: React.FC<InductionProps> = ({ data }) => {
+  const isValid = (d: Comments) =>
+    d.commentType === FreeTextType.INDUCTION && d.texts && d.texts.length > 0;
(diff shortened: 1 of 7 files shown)


src/model/Induction.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const Induction: React.FC<InductionProps> = ({ data }) => {
+  const isValid = (d: Comments) =>
+    d.commentType === FreeTextType.INDUCTION && d.texts && d.texts.length > 0;
+  if (!data.comments || !data.comments.some(d => isValid(d))) {
+    return null;
+  }
+  const inductionData = data.comments.filter(d => isValid(d));
+  if (!inductionData || inductionData.length <= 0) {
```
its-mahdi: I think it's better to check the length with a positive condition. Either with `inductionData.length === 0` or `!inductionData.length > 0`, because the return value of the `length` function is an `unsigned int`, so it can't be negative.

src/model/TissueSpeficity.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const TissueSpeficity: React.FC<TissueProps> = ({ data }) => {
+  const isValid = (d: Comments) =>
+    d.commentType === FreeTextType.TISSUE_SPECIFICITY &&
+    d.texts &&
+    d.texts.length > 0;
+  if (!data.comments || !data.comments.some(d => isValid(d))) {
+    return null;
+  }
+  const tissueSpecifityData = data.comments.filter(d => isValid(d));
+  if (!tissueSpecifityData || tissueSpecifityData.length <= 0) {
```
its-mahdi: Same as previous comment above. I think it's better to check for a positive condition in the `length` check.

src/model/Induction.tsx
dlrice: yep, totally agree. changed.

src/model/TissueSpeficity.tsx
dlrice: likewise as above. changed.

PR #32, opened 2019-04-29
Add xref to Expression section; refactor Xref so that InfoList receiv…

dlrice:
…es an array of objects rather than creating separate InfoList instances.

https://www.ebi.ac.uk/panda/jira/browse/TRM-21745
Yes, 83.72% lines covered
- [x] Is it possible to visually evaluate this PR in the browser? If so, what are the steps to do this (eg 
specific URLs or sections)?
Any entry page. Eg http://localhost:XXXXX/uniprotkb/Q9HC29
No
Yes


src/model/XRef.tsx
```diff
@@ -100,13 +100,12 @@ export const XRefList: React.FC<XRefListProps> = ({
...
           <XRefItem xRefEntry={xref} accession={accession} />
         </ul>
       );
     });
-  const infoData = [{ title: database, content }];
-  return <InfoList infoData={infoData} />;
+  return <Fragment>{nodes}</Fragment>;
 };
 
 export const XRefCategoryTable: React.FC<XRefCategoryTableProps> = ({
```
xwatkins: This is returning an InfoList, not a Table so XrefCategoryInfoList would be more fitting.

xwatkins:
All fine except a naming comment. Also wondering whether some of the types should have `?`?

src/model/XRef.tsx
dlrice: It's returning more than an InfoList (also an hr and a title). Now refactored so that is returning only an InfoList and updated component name to `XRefCategoryInfoList`.

dlrice:
In previous commit, I have also made all of the backend-related types optional.

src/model/__tests__/__snapshots__/Xref.spec.tsx.snap
```diff
@@ -16,29 +16,25 @@ exports[`XRef should render for ProteinProcessing section 1`] = `"<Fragment />"`
...
 "<Fragment>
-  <XRefCategoryTable databaseCategory=\\"GENOME\\" databases={{...}} xRefData={{...}} accession=\\"P21802\\" />
-</Fragment>"
-`;
-
-exports[`XRef should render for Structure section 1`] = `"<Fragment />"`;
-
-exports[`XRefCategoryTable should render for database category GENOME and Ensembl 1`] = `
-"<div>
   <hr />
```
xwatkins: The `<hr/>` also displays when there's nothing above (see O00311 Expression for instance). Would it be better to put it at the end of the `<Fragment>` block only if there's more `foundDatabaseCategory` to display?

PR #33, opened 2019-04-29
Keyword trm 21844

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-21844
Yes, 86.67% of lines covered.
Any entry (eg http://localhost:55295/uniprotkb/P05067)
No
Yes


src/entry/Entry.tsx
```diff
@@ -12,6 +12,9 @@ import { SequenceViewEntry } from '../model/SequenceView';
 import TissueSpeficity from '../model/TissueSpeficity';
 import Induction from '../model/Induction';
 import EntrySectionType from '../model/types/EntrySection';
+import { Keyword } from '../model/Keyword';
+// import entrySectionToKeywordCategories from '../data/keywords';
```
xwatkins: Remove line if not needed

src/entry/Entry.tsx
```diff
@@ -12,6 +12,9 @@ import { SequenceViewEntry } from '../model/SequenceView';
...
+import KeywordCategoryType from '../model/types/keywordTypes';
```
xwatkins: This doesn't seem to get used in this class?

src/model/Keyword.tsx
```diff
@@ -0,0 +1,97 @@
...
+type KeywordItempProps = {
+  id?: string;
+  value?: string;
+};
+
+export const KeywordItem: React.FC<KeywordItempProps> = ({ id, value }) => {
+  if (!id || !value) {
+    return null;
+  }
+  return <a href={apiUrls.keyword(id)}>{value}</a>;
```
xwatkins: This will eventually be a `Link` pointing to `/keywords/KW-0418` rather than an anchor. I know the url won't be working but I wonder if it would be better to link to "what will be" so we don't have to change these links again later?

dlrice: Updated to use `Link` point to `/keywords/KW-XXXX`.

src/entry/Entry.tsx
dlrice: Removed.

src/entry/Entry.tsx
dlrice: Removed.

src/model/Keyword.tsx
```diff
@@ -0,0 +1,97 @@
...
+  }
+  const infoData = Object.keys(foundCategoryToKeywords)
+    .sort()
+    .map(foundCategory => ({
+      title: foundCategory,
+      content: (
+        <KeywordList keywords={foundCategoryToKeywords[foundCategory]} />
+      ),
+    }));
+  const node = infoData.length && (
```
xwatkins: In the case where `infoData.length = 0`, `node` takes the value 0 (see expression for O00311). Maybe the function should return null if `infoData.length = 0` and the infoList Fragment otherwise?

dlrice: tots agree.

PR #35, opened 2019-04-30
Trm 21889/protein features

xwatkins:
https://www.ebi.ac.uk/panda/jira/browse/TRM-21889

Entry page

This is the first stab at integrating the Nightingale components with the Entry page. Add to add `react-testing-library` and I think going forward it will eventually have to replace `enzyme` in our codebase.


src/entry/Entry.tsx
```diff
@@ -36,15 +37,32 @@ const Entry: React.FC<EntryProps> = ({ match }) => {
           type={FreeTextType.PATHWAY}
           includeTitle={true}
         />
+        <FeaturesView
+          data={entryData}
+          types={[
```
dlrice: Can we convert `types` to an enum

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+import ProtvistaNavigation from 'protvista-navigation';
+import { loadWebComponent } from '../utils/utils';
+
+type FeatureLocation = {
+  value: number;
+  modifier: string;
+};
+
+type Feature = {
+  type: string;
```
dlrice: These data will will come from a backend request so best to be paranoid and make them optional types.

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+type ProtvistaFeature = {
+  type: string;
+  description: string;
+  start: number;
+  end: number;
+};
+
+type FeaturesViewProps = {
+  data: {
+    sequence: {
```
dlrice: Should all of these be optional attributes too?

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+  },
+  description: {
+    label: 'Description',
+    resolver: (d: ProtvistaFeature) => d.description,
+  },
+};
+
+const processData = (data: Feature[], types: string[] = []) =>
+  data
+    .filter(d => types.length <= 0 || types.includes(d.type))
```
dlrice: I don't mind single character variable names but I remember previous comments requesting more descriptive variable names when I have done this. Could use something like `datum`?

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+    .map(d => {
+      return {
+        accession: d.featureId,
```
dlrice: In this return object there are several attempts at traversing a nested objected without asserting their existence.

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+  featureId: string;
+  description: string;
+  location: {
+    start: FeatureLocation;
+    end: FeatureLocation;
+  };
+};
+
+type ProtvistaFeature = {
+  type: string;
```
dlrice: Can `ProtvistaFeature.type` become an enum?

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+  loadWebComponent('protvista-navigation', ProtvistaNavigation);
+
+  if (!data.features) {
+    return null;
+  }
+
+  const processedData = processData(data.features, types);
+
+  const setTrackData = useCallback(node => {
+    if (node !== null) {
```
dlrice: Do we have to check specifically for `null` or would `!!node` be more encompassing?

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+      node.data = processedData;
+    }
+  }, []);
+
+  const setTableData = useCallback(node => {
+    if (node !== null) {
```
dlrice: Do we have to check specifically for null or would !!node be more encompassing?

package.json
```diff
@@ -36,11 +36,16 @@
   "dependencies": {
     "axios": "^0.18.0",
     "axios-mock-adapter": "^1.15.0",
+    "d3": "^5.9.2",
     "franklin-sites": "0.0.3-dev.33",
     "idx": "^2.5.2",
     "npm-run-all": "^4.1.2",
     "path": "^0.12.7",
     "prop-types": "^15.6.2",
+    "protvista-manager": "^2.0.3",
```
dlrice: Missing `protvista-datatable` from package.

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+  }, []);
+
+  return (
+    <Fragment>
+      <h4>Features</h4>
+      <protvista-manager attributes="highlight displaystart displayend">
+        <protvista-navigation length={data.sequence.value.length} />
+        <protvista-track
+          ref={setTrackData}
+          length={data.sequence.value.length}
```
dlrice: Is all of `data.sequence.value` guaranteed to exist?

src/model/FeaturesView.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const setTableData = useCallback(node => {
+    if (node !== null) {
+      node.data = processedData;
+      node.columns = columns;
+    }
+  }, []);
+
+  return (
+    <Fragment>
+      <h4>Features</h4>
```
dlrice: Will it ever the case that there is nothing for `protvista-track` to show? In which case `null` should be returned rather than empty feature viewer.

dlrice: Accession P05067 appears to have no features to view.

src/entry/Entry.tsx
xwatkins: Fair point! I think we need to find a consistent way to name files. Most under /types start with  lower case and have the `types` suffix so I adopted that, but `EntrySections.ts` doesn't follow that pattern

src/model/FeaturesView.tsx
xwatkins: According to Leo, `sequence` is one of the only things we can assume is always there. No entry without sequence.

src/model/FeaturesView.tsx
xwatkins: Good point, `feature` is a more fitting name.

src/model/FeaturesView.tsx
xwatkins: The type is coming from the API and there is always scope for new feature types being added/changed, so I would rather not.

src/model/FeaturesView.tsx
xwatkins: 👍

src/model/FeaturesView.tsx
xwatkins: 👍

src/model/FeaturesView.tsx
xwatkins: yes

src/model/FeaturesView.tsx
xwatkins: Good catch

src/model/FeaturesView.tsx
xwatkins: Had a chat with Jie - a feature will always have a type and a location, although there are 3 different modifiers for the position (added an enum) and if the modifier is of a certain value we can't assume the FeatureLocation will have a value, so now checking for that.

xwatkins:
Ready for re-review

PR #36, opened 2019-05-09
Add feature viewer to the different sections

xwatkins:
Adding the feature viewer to the relevant sections as described here http://ebi-uniprot.github.io/ProtVista/userGuide.html#feature-categories-and-types
Just needed snapshot update
Go to entry view, check for presence of feature viewer/feature table


src/model/FeaturesView.tsx
```diff
@@ -63,13 +66,18 @@ const columns = {
   },
   description: {
     label: 'Description',
-    resolver: (d: ProtvistaFeature) => d.description,
+    resolver: (d: ProtvistaFeature) => {
+      return `${d.description} ${d.evidences &&
```
dlrice: `(x) => {return y}` blocks can become `x => y`

src/model/FeaturesView.tsx
```diff
@@ -63,13 +66,18 @@ const columns = {
...
+        d.evidences.map(evidence => `(${evidence.evidenceCode}) `)}`;
```
dlrice: Embedding too much inside of template strings makes it difficult read.

src/model/FeaturesView.tsx
```diff
@@ -63,13 +66,18 @@ const columns = {
...
+    },
   },
 };
 
 const processData = (data: Feature[], types: string[] = []) =>
   data
-    .filter(feature => types.length <= 0 || types.includes(feature.type))
+    .filter(feature => {
+      return types.length <= 0 || types.includes(feature.type);
```
dlrice: `(x) => {return y}` blocks can become `x => y`

src/model/FeaturesView.tsx
xwatkins: This is temporary as we need a strategy for evidence tags to work both in website and in Nightingale, so using it as a placeholder.

PR #37, opened 2019-05-13
Trm 21670/involvement in disease section

dlrice:
# Description
- Created CommentType enum & FreeTextType as a subset of this.
- DiseaseInvolvement which is an array of DiseaseInvolvementEntry components
- Changed the v1 mocking

# Checklist
https://www.ebi.ac.uk/panda/jira/browse/TRM-21670
Yes: 90.63%
Yes, any entry
No
Yes


__mocks__/uuid.js
```diff
@@ -1 +1,6 @@
-module.exports = { v1: jest.fn(() => 1) };
+module.exports = {
+  v1: () => {
+    let value = 0;
+    return value++;
```
xwatkins: I don't think this is working, all ids resolve to `0`. I don't think it's worth changing the value of the id anyway?

src/model/DiseaseInvolvement.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </Fragment>
+    );
+  }
+
+  let referenceNode;
+  if (reference) {
+    const { databaseType: database, id } = reference;
+    if (database && id && databaseToDatabaseInfo[database]) {
+      const info = databaseToDatabaseInfo[database];
+      const uri = info.uriLink.replace(/%value/g, id);
```
xwatkins: I think I've seen this line of code somewhere else, would it be worth creating a wrapper component around `ExternalLink` to deal with this?

src/model/DiseaseInvolvement.tsx
```diff
@@ -0,0 +1,125 @@
...
+    evidences &&
+    evidences.map(evidence => (
+      <UniProtEvidenceTag evidence={evidence} key={evidence.id} />
+    ));
+
+  let notesNode;
+  if (note) {
+    const { texts } = note;
+    if (texts) {
+      notesNode = texts.map(text => text.value).join(' ');
```
xwatkins: I don't know if it's best to join them or to have them enclosed within <p> tags?

src/model/DiseaseInvolvement.tsx
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  let descriptionNode;
+  if (description) {
+    descriptionNode = (
+      <Fragment>
+        <u>Disease description</u> {description}
```
xwatkins: Use the `InfoList` pattern, with "Disease description" as the title.

src/model/DiseaseInvolvement.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <ExternalLink url={uri}>
+          See also {info.displayName}:{id}
+        </ExternalLink>
+      );
+    }
+  }
+  return (
+    <Fragment>
+      <h5>
+        {diseaseId} {acronym && `(${acronym})`} {evidenceNodes}
```
xwatkins: Put {evidenceNodes} in a <p> after the <h5>

src/model/DiseaseInvolvement.tsx
```diff
@@ -0,0 +1,125 @@
...
+      </h5>
+      <p>
+        {notesNode}
```
xwatkins: Use `InfoList` with "Disease note" as a title. See below for "Disease description" as the other InfoList item.

src/model/DiseaseInvolvement.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <br />
+        {descriptionNode}
+        <br />
+        {referenceNode}
```
xwatkins: Use InfoList - "See also" could be the title?

src/model/DiseaseInvolvement.tsx
dlrice: Created `XRefExternalLink` in `XRef.tsx`.

src/model/DiseaseInvolvement.tsx
dlrice: Is the blank space template string back ticks ` `` `? If so, could you expand on what you're thinking.

src/model/DiseaseInvolvement.tsx
dlrice: ?

src/model/DiseaseInvolvement.tsx
xwatkins: `texts` is an array of strings, which I think you are converting into a long string, with each array item separated by a space? If so I was wondering if it would be better to have each array item as a `<p/>` instead (maybe not!)

src/model/DiseaseInvolvement.tsx
xwatkins: oh the suspense... looks like my connection crapped out...
Put `{evidenceNodes}` in a `<p/>` after the `<h5/>`

dlrice: got it

src/model/DiseaseInvolvement.tsx
dlrice: I've made these unordered list with no bullets

PR #41, opened 2019-05-23
Trm 20286/in page nav

xwatkins:
- Add InPageNav component
- Decouple data transformation from UI components
- Write new tests for data converters


src/model/utils/XrefUtils.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+  sectionCategories.forEach(category => {
+    // Get the database relevant to the given category
+    const databases = databaseCategoryToDatabases.get(category);
+    const categoryDatabases: XrefsGoupedByDatabase[] = [];
+    if (databases) {
+      databases.forEach(database => {
+        // Filter the xref data to only return the ones for
+        // the given database
+        const databaseXrefs = xrefs.filter(
```
dlrice: This refactoring is more intuitive than the previous version. However, it will be slightly more CPU intensive as all of the  `databaseCrossReferences` will be filtered `|sectionCategories| * |databases for category|` times. For example, for P05067 there are 14 + 34 + 19 = 67 `sectionCategories`/`databases for category` loops and `databaseCrossReferences` has 782 entries which will mean a total of 52,394 equality checks: `xref.databaseType === database`. We haven't finalized performance testing but this block of code may be something to consider in the future if we need to shave off some ms off the loading time.

src/model/uniprotkb/sections/SequenceConverter.ts
```diff
@@ -0,0 +1,124 @@
...
+      ? data.proteinDescription.flag
+      : 'Complete';
+
+    sequenceData.processing = [
+      Flag.PRECURSOR,
+      Flag.FRAGMENTS_PRECURSOR,
+      Flag.FRAGMENT_PRECURSOR,
+    ].includes(data.proteinDescription.flag)
+      ? 'The displayed sequence is further processed into a mature form.'
+      : undefined;
```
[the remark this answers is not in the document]
xwatkins: the `processing` attribute type is defined as optional `?` which works with `undefined` but not `null`

src/model/utils/XrefUtils.tsx
xwatkins: Very good point. Might be worth looking at creating maps at load time to speed up the data retrieval. For now I've added a comment `// TODO Performance improvement`, would be good to add these wherever we spot a potential bottleneck on which we can improve later.

xwatkins:
Addressed most comments, replied to the others.

xwatkins:
> For each of the comments, features and keywords in the converters there are some arrays/enums which are "configuration" like. Eg:
> 
> ```
> if (data.features) {
>     const features = data.features.filter(feature => {
>       return [
>         FeatureTypes.INIT_MET,
>         FeatureTypes.SIGNAL,
>         FeatureTypes.TRANSIT,
>         FeatureTypes.PROPEP,
>         FeatureTypes.CHAIN,
>         FeatureTypes.PEPTIDE,
>         FeatureTypes.MOD_RES,
>         FeatureTypes.LIPID,
>         FeatureTypes.CARBOHYD,
>         FeatureTypes.DISULFID,
>         FeatureTypes.CROSSLNK,
>       ].includes(feature.type);
>     });
> ```
> 
> I am inclined to have these arrays defined at the top of file and then referenced later. What are your feelings?

Missed this comment, I agree, doing this now...

PR #42, opened 2019-06-10
Trm 22352/add action buttons

xwatkins:
- [ x] What is the link to the Jira that this PR is for? 
Trm 22352
- [ x] Have you written tests and do these have >= 75% coverage of the code that you are contributing?
Just snapshots
- [ x] Is it possible to visually evaluate this PR in the browser? If so, what are the steps to do this (eg specific URLs or sections)?
Should see action buttons on results and entry pages


src/results/ResultsContainer.tsx
```diff
@@ -273,19 +279,38 @@ export class Results extends Component<ResultsProps, ResultsContainerState> {
...
+                  <button className="button link-button">
+                    <StatisticsIcon />
+                    Statistics
+                  </button>
+                  <button className="button link-button">Map to</button>
+                </div>
+                <ResultsTable
+                  results={results}
+                  columnNames={columns}
+                  handleRowSelect={this.handleRowSelect}
```
its-mahdi: Since this is not wrapped inside an arrow function to capture `this`, it may cause some problems in some situations due to the `this` value being different than `this` of the instance of the class at the call time or inside the invoked method.

As `.bind()` is less popular these days, I think putting this inside an arrow function would make it a little bit more robust.

src/results/ResultsContainer.tsx
```diff
@@ -273,19 +279,38 @@ export class Results extends Component<ResultsProps, ResultsContainerState> {
...
+                  selectedRows={selectedRows}
+                  handleHeaderClick={this.updateColumnSort}
```
its-mahdi: Same as line `303`

src/results/ResultsContainer.tsx
xwatkins: This should be ok as both `handleRowSelect` and `updateColumnSort` are defined as arrow functions (see https://daveceddia.com/avoid-bind-when-passing-props/#way-4-property-initializers-good-esnext)

its-mahdi: Perfect, I totally missed that!

PR #48, opened 2019-06-13
Extract common code from converters

xwatkins:
Don't got one.
No but I'm going to look into it.


src/model/uniprotkb/sections/PathologyAndBiotechConverter.ts
```diff
@@ -1,59 +1,21 @@
...
-    }
-  }
-  if (data.features) {
-    const features = data.features.filter(feature => {
-      return pathologyAndBiotechFeatures.includes(feature.type);
-    });
-    pathologyAndBiotechData.featuresData = features;
-  }
-  return pathologyAndBiotechData;
+const pathologyAndBiotechComments = [Comment.DISEASE];
```
dlrice: What do you think to `pathologyAndBiotechKeywords` becoming `keywords` and likewise for these config variables?

PR #51, opened 2019-07-05
Xrefs trm 22396

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22396
100% coverage
Any entry.
No.
N/A

The databases which are shown at each entry page section are configured by both categories (ie a set of databases) and singly-picked databases. This PR allows for this degree of specification/configuration.

- Several mappings are created (upon initial page load) to speed up the sorting and filtering of the entry's `databaseCrossReferences` data:
    - `databaseNameToCategory`
    - `entrySectionToDatabaseNames`
    - `entrySectionToDatabaseCategoryOrder`
- Replaced typo ValueWitEvidence with ValueWithEvidence
- Removed Database enum as these change frequently on the backend
- Removed src/data/databaseToDatabaseInfo.json as this is now derived from an endpoint


dlrice:
Forgot to mention, `src/data/databaseInfo.json` will soon become an endpoint which is why I've tried to pass this as an argument rather than importing at the top of `src/utils/database.ts`.

src/data/database.ts
```diff
@@ -1,215 +1,139 @@
...
-  DatabaseCategory.PROTOCOL,
-  DatabaseCategory.OTHER,
-]);
+export const {
+  databaseCategoryToNames,
+  databaseNameToCategory,
+  databaseToDatabaseInfo,
+} = getDatabaseInfoMaps(databaseInfo);
+
+const select = selectDatabases(databaseCategoryToNames);
```
xwatkins: Would it be possible to add context to the name of the `select` variable? eg `databaseSelector`

src/view/uniprotkb/components/XRefView.tsx
```diff
@@ -1,15 +1,19 @@
...
+  databaseToDatabaseInfo,
+} from '../../../data/database';
 import {
   XrefUIModel,
   Xref,
   XrefsGoupedByDatabase,
 } from '../../../model/utils/XrefUtils.js';
 import { Property } from '../../../model/types/modelTypes.js';
 
+const a = [1];
```
xwatkins: Wassat?

PR #63, opened 2019-08-08
Update keyword component

xwatkins:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22668 - see comments on Jira

Only changed snapshots

### DIFF
--- a/src/view/uniprotkb/components/UniProtCard.tsx
+++ b/src/view/uniprotkb/components/UniProtCard.tsx
@@ -1,13 +1,18 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
-import React, { FC } from 'react';
+import React, { FC, Fragment } from 'react';
 import { Link } from 'react-router-dom';
 import idx from 'idx';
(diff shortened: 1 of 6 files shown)


src/view/uniprotkb/components/UniProtCard.tsx
```diff
@@ -17,6 +22,45 @@ const UniProtCard: FC<{ data: UniProtkbAPIModel }> = ({
...
+    ) {
+      infoListData.push({
+        title: 'Gene',
+        content: <GeneNamesView {...convertedGeneNames} />,
+      });
+    }
+  }
+
+  if (data.keywords) {
+    const categorisedKewywords = getKeywordsForCategories(data.keywords, [
```
dlrice: I noticed that the jira suggested "go with the order: function, biological process, disease, followed by the rest alphabetically?" Whereas the cards have only used the function, biological process, disease. Is that intentional?

xwatkins: Yes, just checked with @SangyaP these are the ones selected by curators so we're trying with these first.

PR #66, opened 2019-09-20
Trm 22641 suggester update

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22641
93.67% of lines covered
Generate queries with the following terms:
  - `keyword`
  - `cc_cofactor_chebi`
  - `cc_scl_term`
  - `go`
No
Checks already existed


xwatkins:
It seems GO doesn't use the GO term as it should:

`(keyword:KW-0053) AND (go:apoptotic process [0006915]) AND (cc_cofactor_chebi:"CHEBI:33913")`

dlrice:
To address your previous comment: the go terms now work `(go:"0006915")`.

dlrice:
There is still an issue with quoting which will be fixed with in a separate jira:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22877

src/search/utils/__tests__/QueryStringGenerator.spec.ts
```diff
@@ -1,12 +1,22 @@
 import createQueryString from '../QueryStringGenerator';
-import testData from '../__fixtures__/clauseQueryTestData';
+import {
+  errorFreeTestData,
+  errorThrowingTestData,
+} from '../__fixtures__/clauseQueryTestData';
 
 describe('QueryHelper', () => {
-  testData.forEach(({ description, queryString, clauses }) => {
+  errorFreeTestData.forEach(({ description, queryString, clauses }) => {
```
xwatkins: Splitting hairs here, but I would call `errorFreeTestData` `testData` and `errorThrowingTestData` `erroneousTestData`

src/search/utils/__fixtures__/clauseQueryTestData.ts
```diff
@@ -20,6 +20,102 @@ const testData = [
...
+        },
+        logicOperator: 'AND',
+        queryInput: {
+          stringValue: 'blah',
+        },
+      },
+    ],
+  },
+  {
+    description: 'should handle enzyme classification [EC] search',
```
xwatkins: Wondering whether it would be worth also testing for cases where the id is not present for ec, etc...

src/search/utils/QueryStringGenerator.ts
```diff
@@ -56,16 +48,10 @@ const createValueString = (
...
-    if (id) {
-      return id;
-    }
-    throw new Error('Value not provided in query');
+    throw new Error('term is undefined');
   }
-  if (['organism', 'taxonomy', 'host'].includes(term) && id) {
-    return id;
+  if (id) {
+    return `"${id}"`;
```
xwatkins: I'm wondering if it would be worth doing the same with `stringValue` a few lines below (or maybe it also needs the prefix!?) and wrapping it in `"` to escape special characters. Check with Jie

src/search/utils/__tests__/QueryStringGenerator.spec.ts
dlrice: Yeah, I agree on the `testData`. I'm hesitant about `erroneousTestData` in case a developer thinks the test data isn't right. How about `exceptionThrowingTestData` ?

src/search/utils/__fixtures__/clauseQueryTestData.ts
dlrice: Good idea. Added.

src/search/utils/QueryStringGenerator.ts
dlrice: Yeah, will take that on in https://www.ebi.ac.uk/panda/jira/browse/TRM-22877

PR #67, opened 2019-09-23
Trm 22899 flat card with function

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22899
Any card section
No
Yes

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -25,3 +25,13 @@ export const loadWebComponent = (name: string, className: Function) => {
 
 export const flattenArrays = (arrays: any[][]) =>
   [].concat(...(arrays as any[]));
+
+export const truncateStringWithEllipses = (
+  string: string,
+  maxLength: number
+) => {
+  if (string.length <= maxLength) {
+    return string;
+  }
+  return `${string.slice(0, maxLength)}...`;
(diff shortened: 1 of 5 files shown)


src/utils/utils.ts
```diff
@@ -25,3 +25,13 @@ export const loadWebComponent = (name: string, className: Function) => {
 
 export const flattenArrays = (arrays: any[][]) =>
   [].concat(...(arrays as any[]));
+
+export const truncateStringWithEllipses = (
```
xwatkins: typo, should be `Ellip**sis**`

src/view/uniprotkb/components/GeneNamesView.tsx
```diff
@@ -18,6 +18,20 @@ const GeneNamesView: React.FC<GeneNamesDataProps> = ({
   return <NameView {...props} />;
 };
 
+export const GeneNamesViewFlat = ({
+  name,
+  alternativeNames = [],
```
xwatkins: Wondering if it would be worth adding a default to `GeneNamesListView` for consistency

xwatkins:
If it's possible, could you add it to the summary as well just above **Similarity**?

PR #69, opened 2019-09-25
Trm 22877 double quote query

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22877
94.2% coverage
Generate a query where the input value has spaces.
N/A
N/A

### DIFF
--- a/src/search/utils/QueryStringGenerator.ts
+++ b/src/search/utils/QueryStringGenerator.ts
@@ -51,7 +51,9 @@ const createValueString = (
     throw new Error('term is undefined');
   }
   if (id) {
-    return `"${id}"`;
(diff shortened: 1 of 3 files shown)


src/search/utils/QueryStringGenerator.ts
```diff
@@ -51,17 +53,20 @@ const createValueString = (
     throw new Error('term is undefined');
   }
   if (id) {
-    return `"${id}"`;
+    // We need to not quote ec numbers because this will cause the
```
xwatkins: If there are wildcards it means we don't have an id as it wasn't selected from suggester? If so this is unnecessary.

dlrice: fixed

PR #72, opened 2019-10-01
Trm 22884 advanced search xref

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22884
94.74% of lines
Yes, query builder > search for embl > select "Cross-references / Sequence databases / EMBL" > enter "*" for value and click search
N/A
N/A

### DIFF
--- a/src/search/utils/QueryStringGenerator.ts
+++ b/src/search/utils/QueryStringGenerator.ts
@@ -37,7 +37,7 @@ const createTermString = (
   if (term === undefined) {
     throw new Error('term is undefined');
   }
-  if (term === 'xref' && stringValue.trim() === '*') {
+  if (term === 'xref' && stringValue === '*') {
     return 'database:';
   }
   if (termSuffix) {
@@ -53,7 +53,6 @@ const createValueString = (
   stringValue: string,
   id: string | undefined
 ) => {
-  const stringValueTrimmed = stringValue.trim();
   if (term === undefined) {
     throw new Error('term is undefined');
   }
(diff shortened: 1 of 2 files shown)


src/search/utils/QueryStringGenerator.ts
```diff
@@ -31,11 +31,15 @@ const createTermString = (
...
   itemType: string,
   id: string | undefined,
-  termSuffix: boolean | undefined
+  termSuffix: boolean | undefined,
+  stringValue: string | undefined
 ) => {
   if (term === undefined) {
     throw new Error('term is undefined');
   }
+  if (term === 'xref' && stringValue === '*') {
```
xwatkins: I don't know if whitespace is trimmed from `stringValue` anywhere but if not it probably should.

src/search/utils/QueryStringGenerator.ts
```diff
@@ -56,12 +60,22 @@ const createValueString = (
...
   let valueString = stringValue;
-  if (valuePrefix && !(term === 'xref' && valuePrefix === 'any')) {
-    valueString = `${valuePrefix}-${stringValue}`;
+  if (term === 'xref') {
+    if (!valuePrefix) {
+      throw new Error('valuePrefix not provided in xref query');
+    }
+    // The API will run more quickly when all database entries are
+    // requested by the user if xref-X:* becomes database:X
+    else if (term === 'xref' && stringValue === '*') {
```
xwatkins: you're already checking `term === 'xref'` in the block

dlrice: good catch

src/search/utils/QueryStringGenerator.ts
dlrice: Totally agree. I've just a commit that does all the trimming at the point of query string generation as the query builder form elements are controlled.

xwatkins:
I'm wondering if for the sake of simplicity it might be better to trim the string in the `UPDATE_INPUT_VALUE` reducer? That way we only have to do it once ever.

dlrice:
> I'm wondering if for the sake of simplicity it might be better to trim the string in the `UPDATE_INPUT_VALUE` reducer? That way we only have to do it once ever.

That would have been much nicer and I did try this but unfortunately because the advanced search/query builder input elements are controlled components doing this will prevent users from entering spaces (eg "homo sapiens") as the value will constantly be trimmed.

PR #75, opened 2019-10-04
Trm 22818/annotation score visualisation

dlrice:
# Notes

- This needs this Franklin PR https://github.com/ebi-uniprot/franklin-sites/pull/43 to be merged first as there is an error about `div`s being children of `p`s.
- Provided two different IDs to UniProtEvidenceTag to avoid encountering 'two children with the same key' warning during tests.

# Checklist
https://www.ebi.ac.uk/panda/jira/browse/TRM-22818
100%
In results and protein summary
N/A
N/A

### DIFF
--- a/src/view/uniprotkb/components/ProteinOverviewView.tsx
+++ b/src/view/uniprotkb/components/ProteinOverviewView.tsx
@@ -44,11 +44,7 @@ export const ProteinOverview: FC<{
     },
     {
       title: 'Annotation score',
-      content: (
-        <AnnotationScoreDoughnutChart>
-          {annotationScore}
-        </AnnotationScoreDoughnutChart>
-      ),
+      content: <AnnotationScoreDoughnutChart score={annotationScore} />,
     },
   ];
 
--- a/src/view/uniprotkb/components/ProteinOverviewView.tsx
+++ b/src/view/uniprotkb/components/ProteinOverviewView.tsx
@@ -55,7 +55,9 @@ export const ProteinOverview: FC<{
     {
       title: 'Annotation score',
       content: (
-        <DoughnutChart percent={annotationScoreToPercentage(annotationScore)} />
+        <DoughnutChart percent={annotationScoreToPercentage(annotationScore)}>
+          {annotationScore > 99 ? '99+' : annotationScore}
+        </DoughnutChart>
       ),
     },
   ];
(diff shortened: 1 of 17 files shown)


src/view/uniprotkb/components/ProteinOverviewView.tsx
```diff
@@ -18,6 +19,7 @@ export const ProteinOverview: FC<{
     _ => _.recommendedName.fullName.value
```
its-mahdi: I do realise you are following the `idx` examples/patterns, but two things comes to my mind when I see the `_` here:

1) In most FP languages, `_` is where the data is ignored. Basically a placeholder for something that we don't need to access.

2) A utility function from `Lodash` library.

Does it make sense to use a better parameter name here e.g. `o` or `obj`.

src/view/uniprotkb/components/ProteinOverviewView.tsx
```diff
@@ -42,7 +44,11 @@ export const ProteinOverview: FC<{
     },
     {
       title: 'Annotation score',
-      content: <Bubble value={annotationScore} />,
+      content: (
+        <AnnotationScoreDoughnutChart>
+          {annotationScore}
```
its-mahdi: Would it be cleaner and more succinct if `{annotationScore}` was passed as `prop` to `AnnotationScoreDoughnutChart` e.g.

    <AnnotationScoreDoughnutChart score={annotationScore} />

src/view/uniprotkb/components/ProteinOverviewView.tsx
dlrice: Interesting point, I'll raise this as a broader question with you both (and @xwatkins) as throughout our code base (and in Franklin) we follow this pattern. Would prefer to fix - if that is the consensus - in a separate jira/issue and not as part of this PR.

src/view/uniprotkb/components/ProteinOverviewView.tsx
dlrice: Good point. Updated.

src/view/uniprotkb/components/ProteinOverviewView.tsx
dlrice: Updated.

PR #81, opened 2019-10-23
fixes TRM-22820: populating advance search form with pre-selected fields

its-mahdi:
TRM-22820
No. The coverage seems to be good.
Going to the query builder page should display a few pre-loaded clauses.
No.
No.

### DIFF
--- a/src/search/AdvancedSearchContainer.tsx
+++ b/src/search/AdvancedSearchContainer.tsx
@@ -122,7 +122,7 @@ const mapDispatchToProps = (dispatch: Dispatch<RootAction>) =>
       dispatchUpdateQueryString: queryString =>
         searchActions.updateQueryString(queryString),
       dispatchSetPreSelectedClauses: () =>
-        searchActions.addPreSelectedClauses(),
+        searchActions.setPreSelectedClauses(),
     },
     dispatch
   );
--- a/src/search/AdvancedSearchContainer.tsx
+++ b/src/search/AdvancedSearchContainer.tsx
@@ -38,7 +38,7 @@ type Props = {
   ) => void;
   handleLogicChange: (clauseId: string, value: Operator) => void;
   handleRemoveClause: (clauseId: string) => void;
-  dispatchInitWithPreSelectedClauses: () => void;
+  dispatchSetPreSelectedClauses: () => void;
 } & RouteComponentProps;
 
 type State = {
@@ -121,7 +121,7 @@ const mapDispatchToProps = (dispatch: Dispatch<RootAction>) =>
       dispatchUpdateClauses: clauses => searchActions.updateClauses(clauses),
       dispatchUpdateQueryString: queryString =>
         searchActions.updateQueryString(queryString),
-      dispatchInitWithPreSelectedClauses: () =>
+      dispatchSetPreSelectedClauses: () =>
         searchActions.addPreSelectedClauses(),
     },
     dispatch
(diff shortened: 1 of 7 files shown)


src/search/AdvancedSearchContainer.tsx
```diff
@@ -120,6 +121,8 @@ const mapDispatchToProps = (dispatch: Dispatch<RootAction>) =>
       dispatchUpdateClauses: clauses => searchActions.updateClauses(clauses),
       dispatchUpdateQueryString: queryString =>
         searchActions.updateQueryString(queryString),
+      dispatchInitWithPreSelectedClauses: () =>
```
dlrice: To be inline with the rest of the `dispatchX` props this would become `dispatchAddPreSelectedClauses ` (or `dispatchSetPreSelectedClauses`)

src/search/utils/clause.ts
```diff
@@ -48,6 +21,71 @@ export const createEmptyClause = (): Clause => ({
   queryInput: {},
 });
 
+export const createPreSelectedClauses = () : Clause[] => [
+  {
+    id: v1(),
+    logicOperator: Operator.AND,
+    searchTerm: {
```
dlrice: I'm aware that we're creating some work for ourselves if the `searchTerms` changes. As I think we're testing this pre-filled advanced search with users it's fine to leave as is but I'll create a jira that states if we're keeping this it's best to use the entries from the fetched `searchTerms` rather than using the static values used here.

src/search/types/searchTypes.ts
```diff
@@ -9,13 +9,31 @@ export enum Operator {
   NOT = 'NOT',
 }
 
+export enum itemType {
```
dlrice: Nice! Good to have types defined just once and in the right place.

src/search/state/searchActions.ts
```diff
@@ -19,6 +19,7 @@ export const UPDATE_RANGE_VALUE = 'UPDATE_RANGE_VALUE';
 export const UPDATE_LOGIC_OPERATOR = 'UPDATE_LOGIC_OPERATOR';
 export const HANDLE_FIELD_SELECT = 'HANDLE_FIELD_SELECT';
 export const SUBMIT_ADVANCED_QUERY = 'SUBMIT_ADVANCED_QUERY';
+export const ADD_PRE_SELECTED_CLAUSES = 'ADD_PRE_SELECTED_CLAUSES';
```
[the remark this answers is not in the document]
its-mahdi: done!

src/search/AdvancedSearchContainer.tsx
its-mahdi: Done! Didn't notice the pattern!

src/search/utils/clause.ts
its-mahdi: Great, that's fine!

src/search/state/searchActions.ts
```diff
@@ -75,7 +75,7 @@ export const updateLogicOperator = (clauseId: string, value: Operator) =>
 
 export const submitAdvancedQuery = () => action(SUBMIT_ADVANCED_QUERY);
 
-export const addPreSelectedClauses = () => action(ADD_PRE_SELECTED_CLAUSES);
+export const addPreSelectedClauses = () => action(SET_PRE_SELECTED_CLAUSES);
```
dlrice: The action creators also follow the pattern of converting the action string constant to camelCase so `addPreSelectedClauses` would be `setPreSelectedClauses`

its-mahdi: Sorry, totally missed that. fixed now!

PR #82, opened 2019-10-31
Trm 22707 variation table

xwatkins:
Adding the nightingale data table component to the variation section. As there was a lot of data to display in some cases, I had to modify the `protvista-datatable` component to add a "collapse/expand" functionality for each row.


src/view/uniprotkb/components/VariationView.tsx
```diff
@@ -46,17 +53,120 @@ interface ChangeEvent extends Event {
...
+};
+
 const VariationView: FC<{ primaryAccession: string }> = ({
   primaryAccession,
 }) => {
   loadWebComponent('protvista-variation', ProtvistaVariation);
   loadWebComponent('protvista-navigation', ProtvistaNavigation);
   loadWebComponent('protvista-sequence', ProtvistaSequence);
   loadWebComponent('protvista-manager', ProtvistaManager);
+  loadWebComponent('protvista-datatable', ProtvistaDatatable);
```
dlrice: This will be loaded each time the component is evaluated/re-rendered (eg clicking on the evidence tag within an table row causes `VariationView` to run three times. As `loadWebComponent` needs to only happen once I am wondering if this could be taken outside of the component and placed at the top of the file.

src/view/uniprotkb/components/VariationView.tsx
```diff
@@ -46,17 +53,120 @@ interface ChangeEvent extends Event {
...
+                          source: evidence.source.name,
+                          id: evidence.source.id,
+                        } as unknown) as Evidence;
+                      }),
+                      evidenceTagCallback
+                    )}
+                </p>
+              `;
+            })
+          : '',
```
dlrice: There are several lines between the `?` and the `:` in this ternary expression. Could be a little easier to read if the null-case is returned immediately as you have done with the `description`'s resolver.

src/view/uniprotkb/components/VariationView.tsx
```diff
@@ -46,17 +53,120 @@ interface ChangeEvent extends Event {
...
+    },
+    consequence: {
+      label: 'Consequence',
+      child: true,
+      resolver: (d: ProtvistaVariant) => d.consequenceType,
+    },
+    sift: {
+      label: 'SIFT prediction',
+      child: true,
+      resolver: (d: ProtvistaVariant) => `${d.siftPrediction} (${d.siftScore})`,
```
dlrice: Sift prediction is sometimes rendered as "undefined"

src/view/uniprotkb/components/VariationView.tsx
```diff
@@ -46,17 +53,120 @@ interface ChangeEvent extends Event {
...
+    },
+    polyphen: {
+      label: 'Polyphen prediction',
+      child: true,
+      resolver: (d: ProtvistaVariant) =>
+        `${d.polyphenPrediction} (${d.polyphenScore})`,
```
dlrice: Polyphen prediction is sometimes rendered as "undefined"

src/view/uniprotkb/components/VariationView.tsx
```diff
@@ -46,17 +53,120 @@ interface ChangeEvent extends Event {
...
+  const evidenceTagCallback = (
+    evidenceData: EvidenceData,
+    references: Evidence[] | undefined
+  ) => {
+    setSelectedEvidenceData(evidenceData);
+    setSelectedReferences(references);
+    setShowEvidenceTagData(true);
+  };
+
+  const columns = {
```
dlrice: The `columns` variable doesn't change between successive evaluations of `VariationView` but takes up several of the component's lines. A suggestion is to take this out of the component:
```
const getColumns = (
  evidenceTagCallback: (
    evidenceData: EvidenceData,
    references: Evidence[] | undefined
  ) => void
) => ({
  positions: {
    label: 'Position(s)',
    resolver: (d: ProtvistaVariant) =>
      d.begin === d.end ? d.begin : `${d.begin}-${d.end}`,
  },
...
```
and then `setTableData` will only depend on `data`:
```
const setTableData = useCallback(
    (node): void => {
      if (node) {
        // eslint-disable-next-line no-param-reassign
        node.data = data.features;

        const evidenceTagCallback = (
          evidenceData: EvidenceData,
          references: Evidence[] | undefined
        ) => {
          setSelectedEvidenceData(evidenceData);
          setSelectedReferences(references);
          setShowEvidenceTagData(true);
        };

        // eslint-disable-next-line no-param-reassign
        node.columns = getColumns(evidenceTagCallback);
      }
    },
    [data]
  );
```

src/view/uniprotkb/components/VariationView.tsx
```diff
@@ -46,17 +53,120 @@ interface ChangeEvent extends Event {
...
+      label: 'Polyphen prediction',
+      child: true,
+      resolver: (d: ProtvistaVariant) =>
+        `${d.polyphenPrediction} (${d.polyphenScore})`,
+    },
+    description: {
+      label: 'Description',
+      resolver: (d: ProtvistaVariant) => {
+        if (!d.description) {
+          return html``;
```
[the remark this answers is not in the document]
xwatkins: Returning string. Returning null causes issues with `lit-html`

src/view/uniprotkb/components/FeaturesTableView.tsx
```diff
@@ -0,0 +1,79 @@
...
+      evidenceData: EvidenceData,
+      references: Evidence[] | undefined
+    ) => void
+  ) => FeatureColumns;
+}> = ({ data, getColumnConfig }) => {
+  const [showEvidenceTagData, setShowEvidenceTagData] = useState(false);
+  const [selectedEvidenceData, setSelectedEvidenceData] = useState();
+  const [selectedReferences, setSelectedReferences] = useState();
+
+  const evidenceTagCallback: FeaturesTableCallback = (
```
dlrice: As `evidenceTagCallback` is only used inside the `useCallback` this could move inside and will prevent redefinition on subsequent passthroughs of `FeaturesTableView`

src/view/uniprotkb/components/VariationView.tsx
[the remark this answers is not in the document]
dlrice: Could `html''` become `''` then?

src/view/uniprotkb/components/FeaturesTableView.tsx
xwatkins: But if it's defined under `evidenceTagCallback` it will be redefined every time the data or column config change? I remember discussing it this morning but now I don't think it would be much of an improvement?

dlrice: I agree, I don't think there is much in it.

PR #83, opened 2019-11-05
Trm 22575/column selection component

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22362
In the results page.
N/A
N/A


xwatkins:
A few functional issues:
- when searching for a field, selecting one re-renders all the accordions unfiltered and opened.
- the "customise data" button should not be visible on the cards view
- have we discussed persisting the selections on the client, can't remember if we said we wanted to leave that for later
- I'm not sure why but I see WebGL issues in the console when going to the column configuration panel and click in the search input or select another field.
- What gets re-rendered when a field gets selected? There seems to be a bit of a lag.

dlrice:
> * when searching for a field, selecting one re-renders all the accordions unfiltered and opened.

Good catch. Had to fix this in Franklin (PR: https://github.com/ebi-uniprot/franklin-sites/pull/50)

> * the "customise data" button should not be visible on the cards view

Fixed this.

> * have we discussed persisting the selections on the client, can't remember if we said we wanted to leave that for later

I don't think we've discussed saving stuff on the client's machine yet.

> * I'm not sure why but I see WebGL issues in the console when going to the column configuration panel and click in the search input or select another field.

Yeah, I've seen a few WebGL errors/warnings in the console too and also before implementing the column select functionality. I've had a google around a while ago and didn't come up with anything solid and have chosen to ignore them since. It is worrying though and the warnings are not always consistently displayed either.

> * What gets re-rendered when a field gets selected? There seems to be a bit of a lag.

As you know, using memo to prevent renders produces some odd behaviour. Will create an issue to optimise performance of the column selection and create reproducible demo of the bug to report to react (if it actually is a bug).

PR #84, opened 2019-11-14
Add interaction visualisation and table

xwatkins:
For an example, head over to P05067
- the visualisation will be updated later (too big for some entries). On the current website this is just not displayed over a given threshold.
- the table is missing NOTE which is missing in the data. Will talk to Jie

### DIFF
--- a/src/view/uniprotkb/InteractionSection.tsx
+++ b/src/view/uniprotkb/InteractionSection.tsx
@@ -79,11 +79,15 @@ const InteractionSection: FC<{
 }> = ({ data, primaryAccession }): JSX.Element | null => {
   const datatableContainer = useRef<HTMLInteractionDatatable>(null);
   useEffect(() => {
-    if (datatableContainer && datatableContainer.current) {
-      const interactionComment = data.commentsData.get(
-        CommentType.INTERACTION
-      ) as InteractionComment[];
+    const interactionComment = data.commentsData.get(
+      CommentType.INTERACTION
+    ) as InteractionComment[];
 
+    if (
+      datatableContainer &&
+      datatableContainer.current &&
+      interactionComment[0]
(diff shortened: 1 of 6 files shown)


src/view/uniprotkb/InteractionSection.tsx
```diff
@@ -1,24 +1,109 @@
...
 }> = ({ data, primaryAccession }): JSX.Element | null => {
+  const datatableContainer = useRef<HTMLInteractionDatatable>(null);
+  useEffect(() => {
+    if (datatableContainer && datatableContainer.current) {
+      const interactionComment = data.commentsData.get(
+        CommentType.INTERACTION
+      ) as InteractionComment[];
+
+      // eslint-disable-next-line no-param-reassign
+      datatableContainer.current.data = interactionComment[0].interactions;
```
dlrice: is `interactionComment[0]` always guaranteed to exist?

src/view/uniprotkb/InteractionSection.tsx
```diff
@@ -1,24 +1,109 @@
...
+import { loadWebComponent } from '../../utils/utils';
+import {
+  getIntActQueryUrl,
+  getIntActQueryForAccessionUrl,
+} from '../../utils/externalUrls';
+
+loadWebComponent('interaction-viewer', InteractionViewer);
+loadWebComponent('protvista-datatable', ProtvistaDatatable);
+
+const getInteractionColumns = (primaryAccession: string) => {
```
dlrice: can be `=> (` (ie without needing return)

src/view/uniprotkb/InteractionSection.tsx
xwatkins: Added a check

PR #85, opened 2019-11-14
More columns

xwatkins:
More columns added...but also had to add missing function sections to the entry page.


src/view/uniprotkb/FunctionSection.tsx
```diff
@@ -12,11 +12,97 @@ import {
...
+  BioPhysicoChemicalProperties,
+  Absorption,
+  KineticParameters,
+} from '../../model/uniprotkb/sections/FunctionConverter';
+
+export const AbsorptionView: FC<{ data: Absorption }> = ({ data }) => {
+  return (
+    <Fragment>
+      <section className="text-block">
+        Abs(max) = {data.approximate && '~'}
```
dlrice: wondering what you think of:
```
{`Abs(max) = ${data.approximate && '~'}${data.max}nm`}
```

src/view/uniprotkb/FunctionSection.tsx
```diff
@@ -12,11 +12,97 @@ import {
...
+export const KineticsView: FC<{ data: KineticParameters }> = ({ data }) => {
+  return (
+    <Fragment>
+      <section className="text-block">
+        {data.michaelisConstants && (
+          <ul className="no-bullet">
+            {data.michaelisConstants.map(km => (
+              <li key={km.constant}>
+                K<sub>M</sub>={km.constant}
+                {km.unit} for {km.substrate}{' '}
```
dlrice: wondering what you think of:
```
{`=${km.constant}${km.unit} for ${km.substrate} `}
```

src/view/uniprotkb/SequenceSection.tsx
```diff
@@ -23,12 +28,30 @@ const SequenceSection: FC<{
...
           sequence={data.sequence.value}
         />
-        {data.sequenceCaution && (
+        {data.sequenceCaution && data.sequenceCaution.length > 0 && (
           <Fragment>
             <h4>Sequence caution</h4>
             <SequenceCautionView data={data.sequenceCaution} />
           </Fragment>
         )}
+        {data.massSpectrometry && data.massSpectrometry.length > 0 && (
```
dlrice: Do we strictly need the `>0` check?

src/view/uniprotkb/components/FreeTextView.tsx
```diff
@@ -1,43 +1,39 @@
...
+  );
 
   return (
     <Fragment>
-      {includeTitle && (
-        <h4 style={{ textTransform: 'capitalize' }}>
-          {comments[0].commentType.toLowerCase()}
-        </h4>
-      )}
+      {title && <h4 style={{ textTransform: 'capitalize' }}>{title}</h4>}
```
dlrice: We have so far not used put styles inline like this - is this intentional?

src/view/uniprotkb/components/NumberView.tsx
```diff
@@ -5,7 +5,7 @@ export enum Unit {
   AA = 'AA',
 }
 
-const numberView = ({ value, unit }: { value: number; unit?: Unit }): string =>
-  `${formatLargeNumber(value)}${unit ? ` ${unit}` : ''}`;
+const numberView = ({ value, unit }: { value?: number; unit?: Unit }): string =>
+  value ? `${formatLargeNumber(value)}${unit ? ` ${unit}` : ''}` : '';
```
dlrice: If `value` is `0` this will return `''` - is that intended?

src/view/uniprotkb/components/SequenceView.tsx
```diff
@@ -145,6 +151,62 @@ export const SequenceCautionView: React.FC<{ data: SequenceCaution[] }> = ({
   );
 };
 
+export const MassSpectrometryView: React.FC<{ data: MassSpectrometry[] }> = ({
+  data,
+}) => (
+  <Fragment>
+    {data.map(item => (
+      <section className="text-block" key={`${item.molWeight}${item.method}`}>
+        Molecular mass is {numberView({ value: item.molWeight, unit: Unit.DA })}{' '}
```
dlrice: What do you think of:
```
{`Molecular mass is ${numberView({ value: item.molWeight, unit: Unit.DA })} from positions `}
```

src/model/ColumnConfiguration.tsx
```diff
@@ -234,49 +264,174 @@ ColumnConfiguration.set(Column.ftVarSeq, {
     );
   },
 });
+ColumnConfiguration.set(Column.fragment, {
+  label: 'Fragment',
+  render: data => {
+    const { flag } = data[EntrySection.Sequence];
+    const isFragment =
+      flag &&
```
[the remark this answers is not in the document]
xwatkins: It doesn't in this case because the expected type in the array is `Flag`

src/model/ColumnConfiguration.tsx
```diff
@@ -302,45 +457,102 @@ ColumnConfiguration.set(Column.errorGmodelPred, {
...
+  render: data => {
+    const { entryAudit } = data[EntrySection.Sequence];
+    return entryAudit && entryAudit.lastSequenceUpdateDate;
+  },
+});
+ColumnConfiguration.set(Column.version, {
+  label: 'Version',
+  render: data => {
+    const { entryAudit } = data[EntrySection.Sequence];
+    return entryAudit && <span>{entryAudit.entryVersion}</span>;
```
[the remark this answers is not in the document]
xwatkins: entryVersion is a number and JSXElement is expected so I wrapped it in a Fragment instead

src/model/types/CommentTypes.ts
```diff
@@ -119,14 +157,14 @@ export type Isoform = {
   isoformSequenceStatus: string;
   isoformIds: string[];
   synonyms: { value: string }[];
-  note: { texts: { value: string }[] };
+  note: { texts: TextWithEvidence[] };
```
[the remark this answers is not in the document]
xwatkins: Yes, although I didn't find an example with evidences it's better to be generic.

src/model/uniprotkb/SectionConverter.ts
```diff
@@ -10,12 +10,12 @@ import KeywordCategory from '../types/KeywordCategory';
 import FeatureType from '../types/FeatureType';
 import { UniProtkbAPIModel } from './UniProtkbConverter';
 
-export type UIModel = {
+export interface UIModel {
```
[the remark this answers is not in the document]
xwatkins: Inheritance! You can inherit from `interface` but not `type`

src/model/types/CommentTypes.ts
```diff
@@ -152,6 +226,13 @@ type Comment =
   | InteractionComment
   | AlternativeProducts
   | SequenceCaution
-  | SubcellularLocation;
+  | SubcellularLocation
+  | MassSpectrometry
+  | RNAEditing
+  | AbsorptionComment
```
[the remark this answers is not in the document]
xwatkins: Good idea. Wondering whether to use the suffix `Comment` or `CC` which is what it's referred to internally?

src/model/uniprotkb/SectionConverter.ts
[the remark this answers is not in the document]
xwatkins: So use `interface` if you need to extend, otherwise `type` is more efficient I think?

src/view/uniprotkb/FunctionSection.tsx
xwatkins: Much nicer

src/view/uniprotkb/FunctionSection.tsx
xwatkins: Much nicer

src/view/uniprotkb/SequenceSection.tsx
xwatkins: I don't think we'd want to display the titles if these were defined but empty

src/view/uniprotkb/components/FreeTextView.tsx
xwatkins: Nope! thanks for the heads up

src/view/uniprotkb/components/NumberView.tsx
xwatkins: Good catch, replaced with type check.

xwatkins:
Addressed most comments, commented on others.

src/model/types/CommentTypes.ts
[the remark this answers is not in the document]
dlrice: If a new dev was to come along i think `Comment` would be clearer. A uniprot pro would be happy with both.

src/model/ColumnConfiguration.tsx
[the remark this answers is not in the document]
dlrice: I see your point.

PR #95, opened 2020-01-13
Lost query builder trm 23336

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23336
1. Goto query builder
2. Enter query
3. Click search
4. Push back
5. The search will still be there
6. Click search
7. Click query builder in the header
8. The query will not be there
**...to see the query in the search bar:**
9. Click search
10. Select an result
11. The query will be in the search input
n/a
n/a

### DIFF
--- a/src/search/AdvancedSearchContainer.tsx
+++ b/src/search/AdvancedSearchContainer.tsx
@@ -14,6 +14,7 @@ import {
 } from './types/searchTypes';
 import AdvancedSearch from './AdvancedSearch';
 import createQueryString from './utils/QueryStringGenerator';
+import { queryBuilderPath } from '../App';
 
 import './styles/SearchContainer.scss';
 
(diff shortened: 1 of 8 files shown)


src/search/AdvancedSearchContainer.tsx
```diff
@@ -54,6 +55,18 @@ export class Search extends Component<Props, State> {
...
 
+  componentDidMount() {
+    const {
+      dispatchSetPreSelectedClauses,
+      location: { pathname },
+      history,
+    } = this.props;
+    if (pathname === `${queryBuilderPath}/reset`) {
+      dispatchSetPreSelectedClauses();
+      history.replace(queryBuilderPath);
```
xwatkins: I'm not sure if we need to do this is this path is not defined in the router?

PR #96, opened 2020-01-13
Fixed issue when missing data

xwatkins:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23379

### DIFF
--- a/src/view/uniprotkb/FunctionSection.tsx
+++ b/src/view/uniprotkb/FunctionSection.tsx
@@ -104,29 +104,34 @@ const BioPhysicoChemicalPropertiesView: FC<{
 };
 
 export const CofactorView: FC<{
-  cofactors: CofactorComment[];
+  cofactors?: CofactorComment[];
   title?: string;
-}> = ({ cofactors, title }) => (
-  <Fragment>
-    {title && <h3>{title}</h3>}
-    {cofactors.map(cofactorComment => (
-      <section className="text-block" key={v1()}>
-        {cofactorComment.cofactors &&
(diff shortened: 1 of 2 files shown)


src/view/uniprotkb/FunctionSection.tsx
```diff
@@ -104,29 +104,34 @@ const BioPhysicoChemicalPropertiesView: FC<{
...
-          ))}
-        {cofactorComment.note && (
-          <TextView comments={cofactorComment.note.texts} />
-        )}
-      </section>
-    ))}
-  </Fragment>
-);
+}> = ({ cofactors, title }) => {
+  if (!cofactors || cofactors.length <= 0) {
```
dlrice: `cofactors.length` can never be < 0 so I would suggest `!cofactors.length`

Here's a nice SO answer about testing for array length: https://stackoverflow.com/questions/24403732/check-if-array-is-empty-or-does-not-exist-js

src/view/uniprotkb/components/SubcellularLocationView.tsx
```diff
@@ -5,8 +5,11 @@ import UniProtEvidenceTag from '../../../components/UniProtEvidenceTag';
 import { TextView } from './FreeTextView';
 
 const SubcellularLocationView: FC<{
-  comments: SubcellularLocationComment[];
+  comments?: SubcellularLocationComment[];
 }> = ({ comments }) => {
+  if (!comments || comments.length <= 0) {
```
dlrice: similarly to previous comment I would suggest: `!comments.length`

PR #98, opened 2020-01-21
Franklin changes

xwatkins:
Changes required by new styles (typography and colours)


src/view/uniprotkb/components/styles/UniProtTitle.scss
```diff
@@ -4,9 +4,9 @@
     display: inline-flex;
     align-self: center;
     svg {
-      width: 2rem;
-      height: 2rem;
-      top: 0.125em;
+      width: 0.75em;
+      height: 0.75em;
+      margin-right: 0.25rem;
```
its-mahdi: Is the mixing of `em` and `rem` units intentional here?

xwatkins: Yes it 's so the icon size reflects the title font size

PR #99, opened 2020-01-27
Trm 22361 download

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22361
Yes, goto the download section from the results page.
N/A
N/A


src/results/DownloadContainer.tsx
```diff
@@ -0,0 +1,168 @@
...
+        setDownloadAll(e.target.value === 'true')
+      }
+      onFileFormatChange={(e: React.ChangeEvent<HTMLSelectElement>) =>
+        setFileFormat(e.target.value as FileFormat)
+      }
+      onCompressedChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+        setCompressed(e.target.value === 'true')
+      }
+      preview={
+        compareDownloadUrlsDisregardSize(preview.url, downloadUrl) &&
```
xwatkins: For clarity, it would be nice to have this evaluated into a variable, maybe `resetPreview`.

src/results/DownloadContainer.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { withRouter, RouteComponentProps } from 'react-router-dom';
+import idx from 'idx';
+import { RootState } from '../state/state-types';
+import DownloadView from './DownloadView';
+import { Column } from '../model/types/ColumnTypes';
+import { FileFormat, fileFormatToContentType } from './types/resultsTypes';
+import { getDownloadUrl } from '../utils/apiUrls';
+import fetchData from '../utils/fetchData';
+
+export const replaceExcelWithTsv = (fileFormat: FileFormat) =>
```
xwatkins: This could be made more generic by dissociating it from the name of data types, renaming it to `getPreviewFileFormat`. This means that if file format changes later and there's another file format which is not supported by preview we can just add it to that method.

src/results/ResultsContainer.tsx
```diff
@@ -306,8 +309,21 @@ export class Results extends Component<ResultsProps, ResultsContainerState> {
                 Align
               </button>
               <button type="button" className="button tertiary">
-                <DownloadIcon />
-                Download
+                <Link
+                  to={{
+                    pathname: '/download',
+                    state: {
+                      query,
```
xwatkins: As a I think most of these parameters are present in the store and `DownloadContainer` is connected to it, I wonder if it would be nicer to get these from the store directly instead of passing them as props?

src/results/DownloadContainer.tsx
```diff
@@ -0,0 +1,168 @@
...
+    state: {
+      query,
+      selectedFacets,
+      sortColumn,
+      sortDirection,
+      selectedEntries,
+    },
+  },
+}) => {
+  const [selectedColumns, setSelectedColumns] = useState(tableColumns);
```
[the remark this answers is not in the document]
dlrice: That's a good question. I passed in the table columns because I assumed that a user would want to see download what they're seeing (assuming they're in table mode). If a user is using card mode and they want to download the results it's not clear which columns exactly they want to download so it would make sense to store these in a separate place as you've suggested. We might even need to know which view the user was using when showing the download columns. Shall we poll the team for their perspectives?

src/results/DownloadContainer.tsx
dlrice: Yes, would be much clearer.

src/results/DownloadContainer.tsx
dlrice: Makes sense. Nice future proofing.

src/results/ResultsContainer.tsx
dlrice: Yes, no point passing things in the link's state if it is available elsewhere. I'll take these out.

src/results/DownloadContainer.tsx
```diff
@@ -0,0 +1,168 @@
...
+import DownloadView from './DownloadView';
+import { Column } from '../model/types/ColumnTypes';
+import { FileFormat, fileFormatToContentType } from './types/resultsTypes';
+import { getDownloadUrl } from '../utils/apiUrls';
+import fetchData from '../utils/fetchData';
+
+export const replaceExcelWithTsv = (fileFormat: FileFormat) =>
+  fileFormat === FileFormat.excel ? FileFormat.tsv : fileFormat;
+
+export const compareDownloadUrlsDisregardSize = (
```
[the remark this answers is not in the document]
dlrice: Nice idea.

src/results/ResultsContainer.tsx
dlrice: Looking at each of the bits of data passed in the `Link`:

Taken from the URL within the `ResultsContainer` and then updated in the store:
* `query`

Taken from the URL within the `ResultsContainer` but not updated in the store:
* `selectedFacets`
* `sortColumn`
* `sortDirection`

Only exists as local state to `ResultsContainer`:
* `selectedEntries`

It looks like the only thing I could take is the `query` - would you prefer this to be taken from the store rather than passed via the `Link`?

src/results/DownloadContainer.tsx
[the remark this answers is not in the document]
xwatkins: Would you mind making a Jira about this and assigning it to Sangya?

src/results/ResultsContainer.tsx
xwatkins: No that's fine. Had forgotten that the main state came from the url

src/results/DownloadContainer.tsx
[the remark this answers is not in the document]
dlrice: done: https://www.ebi.ac.uk/panda/jira/browse/TRM-23576

PR #104, opened 2020-02-26
Add translated xref descriptions

xwatkins:


src/model/utils/XrefUtils.tsx
```diff
@@ -183,6 +179,32 @@ export const getUnconditionalImplicitXrefs = () => {
   return foundXrefs;
 };
 
+export const getJoinedXrefs = (xrefs: Xref[]) => {
+  const joinedXrefs = [...xrefs];
+  const scheduledForRemoval: string[] = [];
+  joinedXrefs
+    .filter(xref => xref.properties && xref.properties.Status === 'JOINED')
+    .forEach(xref => {
+      const masterXref = joinedXrefs.find(
```
dlrice: The reference to `masterXref` is getting updated by side effect here. Is it possible to move towards a more functional approach without modifying references?

src/model/utils/XrefUtils.tsx
```diff
@@ -183,6 +179,32 @@ export const getUnconditionalImplicitXrefs = () => {
   return foundXrefs;
 };
 
+export const getJoinedXrefs = (xrefs: Xref[]) => {
+  const joinedXrefs = [...xrefs];
+  const scheduledForRemoval: string[] = [];
+  joinedXrefs
+    .filter(xref => xref.properties && xref.properties.Status === 'JOINED')
+    .forEach(xref => {
```
dlrice: The combination of `forEach` and `find` will make this O(n^2). I am wondering if it is possible to make this O(n) ie a single pass through the xrefs?

src/view/uniprotkb/components/XRefView.tsx
```diff
@@ -85,24 +90,84 @@ type XRefProps = {
...
+      {additionalIds &&
+        additionalIds.map(additionalId => (
+          <ExternalLink url={externalUrls.ENA(additionalId)} key={additionalId}>
+            {additionalId}
+          </ExternalLink>
+        ))}
+      {properties &&
+        properties.MoleculeType &&
+        `${
+          ((EMBLXrefProperties as unknown) as { [key: string]: string })[
```
dlrice: can avoid the use of `uknown` with:
```
EMBLXrefProperties[properties.MoleculeType as keyof typeof EMBLXrefProperties
```
and likewise at L145

xwatkins: Nice.

src/model/utils/XrefUtils.tsx
xwatkins: I guess we could make a deep copy of xrefs at the top of the function?

dlrice: It probably doesn't really matter if the xrefs are modified in place (and as an added bonus this will be a quick operation). My only issue really here is really just readability as it's not immediately obvious what happens to `masterXref` until you realise it's being modified in place. I'm happy to leave as is though.

PR #110, opened 2020-03-06
Additional xrefs trm 23380

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23380
We have >=75% but are not testing this as the mock api model doesn't use the database.
P05067
N/A
N/A

### DIFF
--- a/src/view/uniprotkb/components/XRefView.tsx
+++ b/src/view/uniprotkb/components/XRefView.tsx
@@ -7,6 +7,7 @@ import {
   databaseCategoryToString,
   databaseToDatabaseInfo,
   PDBMirrors,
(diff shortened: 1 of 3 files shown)


src/view/uniprotkb/components/XRefView.tsx
```diff
@@ -234,21 +235,39 @@ export const DatabaseList: FC<{
...
-          primaryAccession={primaryAccession}
-          crc64={crc64}
-        />
-      ),
-    }))}
-  </ExpandableList>
-);
+}) => {
+  let viewItem;
+  const viewLink = viewProteinLinkDatabases.get(database);
```
xwatkins: Please add comment

PR #114, opened 2020-03-11
Refactor layout

xwatkins:


src/layout/styles/BaseLayout.scss
```diff
@@ -0,0 +1,19 @@
...
+.base-layout {
+  display: grid;
+  grid-template-areas:
+    'main-header'
+    'content'
+    'footer';
+
+  grid-template-rows: 4rem auto auto;
+
+  .main-header {
```
dlrice: Do we want to make these BEM like ie `.base-layout__main-header`?

src/layout/styles/SideBarLayout.scss
```diff
@@ -1,28 +1,38 @@
...
-  display: grid;
-  column-gap: $gutter-size;
-  grid-template-columns: 16.6vw auto;
-  grid-template-rows: auto 100vh;
-  grid-template-areas: 'title title' 'sidebar content';
-  &--inverted {
-    grid-template-areas: 'title title' 'content sidebar';
-    grid-template-columns: 6fr 1fr;
-  }
+  .base-layout {
```
dlrice: Is this the same name as in `BaseLayout.scss` to override it?

PR #119, opened 2020-04-14
Trm 23951 return more than data

aurel-l:
  - Jira [TRM-23951](https://www.ebi.ac.uk/panda/jira/browse/TRM-23951)
  - updated tests for visual components affected, added tests for hook logic and for fetch logic
  - entry page, variation section
  - in the custom hook (consistent-return, and default-case)

note: the interaction-viewer is only here because I branched off of master instead of this sprint's branch.


src/model/__tests__/__snapshots__/ColumnConfiguration.spec.tsx.snap
```diff
@@ -904,7 +904,62 @@ exports[`ColumnConfiguration component should render all columns 20`] = `
 </DocumentFragment>
 `;
 
-exports[`ColumnConfiguration component should render all columns 21`] = `<DocumentFragment />`;
+exports[`ColumnConfiguration component should render all columns 21`] = `
```
xwatkins, line 907: Wondering why there is a loading message displayed here. Is there an async call happening in the columns? Can't remember...

src/hooks/useDataApi.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (!url) {
+      dispatch({ type: ActionType.SUCCESS });
+      return;
+    }
+
+    dispatch({ type: ActionType.INIT });
+
+    // variables to handle cancellation
+    const source = axios.CancelToken.source();
+    let didCancel = false;
```
xwatkins, line 68: I'm wondering whether you need that flag as you could check whether the thrown error is a cancelation with `axios.isCancel(thrown)`. Unless the promise can be resolved as a success even with cancelation?

src/hooks/useDataApi.ts
```diff
@@ -0,0 +1,95 @@
...
+      // catch error
+      (error: AxiosError) => {
+        if (didCancel) return;
+        dispatch({ type: ActionType.ERROR, error });
+      }
+    );
+
+    // handle unmounting of the hook
+    // eslint-disable-next-line consistent-return
+    return () => {
```
xwatkins, line 86: Does this run if the user navigates away? Does it need to be handled in the component that uses the HOC or is it automatic?

src/hooks/__tests__/useDataApi.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  test('cancellation', async () => {
+    mock.onGet(url).reply(200, 'some data');
+    const { result, unmount } = renderHook(() => useDataApi(url));
+
+    expect(result.current).toEqual({ loading: true });
+
+    unmount();
+    // not sure how to test cancellation, but at least make sure there's no
```
xwatkins, line 81: Wondering if there are also cases where we'd want to actively cancel a request (say something where we could make requests in short successions, for instance autocompletion)

src/model/__tests__/__snapshots__/ColumnConfiguration.spec.tsx.snap
aurel-l: This is because the VariationView is used [here](https://github.com/ebi-uniprot/uniprot-website/blob/425faa6d2f27720690ff43965adf575a86faebe2/src/model/ColumnConfiguration.tsx#L386). Is it not supposed to be used here?

src/hooks/useDataApi.ts
aurel-l: In an ideal world I would have wanted `source` to be able to tell me if it had been cancelled (through something like `source.canceled` for example) but I don't think this is possible.
I did this because I remembered doing something similar in previous code, but using native fetch. In this case you have 2 levels of async, one for getting the response, and one for parsing the JSON (`fetch()` and then `response.json()`) and I was worried that if the cancellation (due to the component unmounting) happened during the 2nd step, the AbortController would not be involved anymore and I wouldn't be able to catch it.
I'm not sure how this happens in axios logic though.
Not sure if I'm making sense 😬 
So since I'm not sure how to test my theory I leant on the safe side here by using my own flag.

src/hooks/useDataApi.ts
aurel-l: It automatically runs whenever the component using this hook unmounts, and whenever `url` changes

src/hooks/__tests__/useDataApi.spec.ts
aurel-l: Same than previous comment, the cancellation happens everytime the URL changes. So the autocomplete logic should be quite straightforward to implement with it since it's just a matter of replacing the URL and the previous fetch will automatically be discarded.
Just making sure that the change of URL is debounced and that's it.

src/hooks/useDataApi.ts
aurel-l: So if you want me to use `isCancel` I'm fine with it if you think the way axios handles it is OK.

src/model/__tests__/__snapshots__/ColumnConfiguration.spec.tsx.snap
xwatkins: It should just be the table from Variation view but it looks like there's no options for having the table only. Leave it with me, I'll make the change on my PR as it's linked.

src/hooks/useDataApi.ts
xwatkins: Makes sense. I don't mind either way.

PR #121, opened 2020-04-17
Message manager

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23898
https://www.ebi.ac.uk/panda/jira/browse/TRM-23950
* Add the following to the `messagesInitialState.ts` to see the in-page messages:
```
const messagesInitialState = {
  active: [
    {
      id: 'in-page-message-1',
      content: 'in-page message 1',
      format: MessageFormat.IN_PAGE,
      level: MessageLevel.FAILURE,
    },
    {
      id: 'in-page-message-2',
      content: 'in-page message 2',
      format: MessageFormat.IN_PAGE,
      level: MessageLevel.SUCCESS,
      locations: [Location.UniProtKB_Results],
    },
  ],
  deleted: {},
};
```
(you'll also need to some imports of types and the `Location` enum)
* Add the following to the body of `MessageManagerComponent` to see the pop-up messages:
```
// Temporary add messages to test pop ups are working
  useEffect(() => {
    const interval = setInterval(() => {
      addMessage({
        id: v1(),
        content: `New message ${v1()}`,
        format: MessageFormat.POP_UP,
        level: MessageLevel.INFO,
      });
    }, 3000);
    return () => clearInterval(interval);
  }, [addMessage]);
```
(you'll also need to some imports of types)
n/a
n/a


src/messages/styles/popupMessageHub.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/settings';
+
+.popup-message-container {
+  z-index: $z-index-high;
```
aurel-l, line 4: Still in a issue here with the content boxes on the home page having a higher z-index

src/layout/BaseLayout.tsx
```diff
@@ -33,6 +33,9 @@ const BaseLayout: React.FC<BaseLayoutProps> = props => {
       <section className="main-header">
         <UniProtHeader isHomePage={isHomePage} isSearchPage={isSearchPage} />
       </section>
+      <section className="in-page-messages">
```
aurel-l, line 36: Either the BaseLayout should alway be rendered, or this should be placed at a higher level (near the root of the app), because at the moment it is unmounted and remounted when browsing through the pages of the website and it's strange (appear/disappear, reactivate the apparition transition for already existing messages)

xwatkins: Created https://www.ebi.ac.uk/panda/jira/browse/TRM-23983

src/messages/styles/popupMessageHub.scss
xwatkins: Maybe I'm looking at the wrong thing, but it seems ok?

image

aurel-l: Somehow my setup doesn't work anymore so can't confirm right now. But yeah, the messages were going below those boxes. If what you see now is fine I guess I had old code and you can resolve that.

PR #122, opened 2020-04-27
Full page errors

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23886
Please read the long comment there.

- [x] Have you written tests and do these have >= 75% coverage of the code that you are contributing? Yes.
for no results page, search a non-existing accession e.g. '/uniprotkb?query=sangya'
for entry not found, go to a page with non-existing accession e.g. '/uniprotkb/sangya'
for page not found, go to '/page-not-found'
for service unavailable, go to '/service-unavailable'
for inactive entries, go to 'uniprotkb/Q00007' and '/uniprotkb/I8FBX0'
- [ ] Are there any linting errors that required you to add exceptions? No.
- [ ] If there are backend requests, have you checked for the existence of the resource's attributes before attempting access? NA.


src/App.tsx
```diff
@@ -56,6 +58,30 @@ const App = () => (
               </BaseLayout>
             )}
           />
+          <Route
+            path="/page-not-found"
```
xwatkins, line 62: During this sprint, @dlrice defined routes in as an enum (see all routes above). I think we should be consistent and use this system for all routes.

src/utils/browserHistory.ts
```diff
@@ -0,0 +1,5 @@
+import { createBrowserHistory } from 'history';
```
xwatkins, line 1: I'm a bit confused by this file, does it just re-export `createBrowserHistory` ?

src/entry/state/entryActions.ts
```diff
@@ -51,8 +52,27 @@ export const fetchEntry = (accession: string) => async (dispatch: Dispatch) => {
...
-    }) /* eslint-disable no-console */
-    .catch(error => console.error(error));
+    })
+    .catch((error) => {
+      const { status } = error.response;
+
+      switch (status) {
+        case 400:
+        case 404:
+          history.push('/page-not-found');
```
xwatkins, line 62: and here you can re-use the page name as set in the enum (see `App.tsx` comment)

src/styles/ErrorPages.scss
```diff
@@ -0,0 +1,17 @@
...
+  flex-direction: column;
+  width: 100%;
+  text-align: center;
+  margin: 10% 0;
+  min-height: 200px;
+  align-items: center;
+
+  &__art-work {
+    width: 400px;
+    height: 400px;
```
xwatkins, line 14: I wonder if to make this responsive, we could add a container around the SVG which uses `vh` as unit (the svg being 100% width and height inside)?

src/styles/ErrorPages.scss
```diff
@@ -0,0 +1,17 @@
+.error-page-container {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-direction: column;
+  width: 100%;
+  text-align: center;
+  margin: 10% 0;
+  min-height: 200px;
```
xwatkins, line 9: If possible, we should avoid `px`.

src/model/uniprotkb/UniProtkbConverter.ts
```diff
@@ -97,6 +98,33 @@ export type UniProtkbUIModel = {
...
+
+export type UniProtkbAPIInactiveEntryModel = {
+  annotationScore: number;
+  entryType: EntryType.INACTIVE;
+  inactiveReason: InactiveEntryReason;
+  primaryAccession: string;
+  uniProtId: string;
+}
+
+export type UniProtKBUIInactiveEntryModel = {
```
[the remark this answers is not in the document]
its-mahdi: Good point, I have renamed to `UniProtkbInactiveEntryModel` now.

src/pages/errors/__tests__/ObsoleteEntryPage.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <ObsoleteEntryPage
+        accession={primaryAccession}
+        details={deletedEntryData.inactiveReason}
+      />
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test('should render demereged page', () => {
```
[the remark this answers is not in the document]
its-mahdi: I had that all over the place, fixed now!

src/App.tsx
its-mahdi: I don't if that was there when I initially added my routes, but I have now moved all of them to the `Location` enum.

src/utils/browserHistory.ts
its-mahdi: No, it doesn't simply re-exports `createBrowserHistory`; It invokes it and then exports the reference to the returned object of the function call. Since imported modules are cached internally, this means multiple files can have access to the same -- in this case, `history` object. Which is what was needed for the page redirects.

I think this is actually very widely used like that.

src/entry/state/entryActions.ts
its-mahdi: I tried to do that but `LocationToPath.get()` which is internally a `Map.get()` does not guarantee a return value, therefore `history.push()` is rejecting it.

Any suggestions?

src/pages/errors/ObsoleteEntryPage.tsx
```diff
@@ -0,0 +1,75 @@
...
+}) => (
+  <Message level="info">
+    <h4>This entry is obsolete</h4>
+    <p data-testid="demerged-entry-message">
+      It can now be found as secondary accession in {' '}
+      {demergedTo
+        .reduce((a: (string | JSX.Element)[], c, i) => {
+          if (i > 0) {
+            if (i === demergedTo.length - 1) {
+              a.push(<span key={v1()}> and </span>);
```
aurel-l, line 41: Let's just stop using uuids for keys shall we 😅.
Maybe the string value, or the even here I'd be OK using the index

src/styles/ErrorPages.scss
aurel-l: It is not not the same width and height inside. I guess that is why this is needed, so that all the illustrations look the same for regarding their position in the page

xwatkins: Yes, but we don't really care about the width, only the height should be consistent. I think we can make the width scale based on a fixed height.

src/pages/errors/ObsoleteEntryPage.tsx
its-mahdi: Oh, I should have used  a simple ` and ` string there! But I agree with you, I don't like using the `uuid` neither. I think when I did it there I had a different logic there and I did initially compose a key like `${c}-and`, but thought using `uuid` might be cleaner, specially because this is not an element which would get any updates.

Maybe we should come up with a few rules-of-thumb to know when to use `uuid` and when shouldn't. I think there are cases that it can be useful, so I'd vote against banning it completely.

aurel-l: I mean, that shouldn't ever be used for React keys. Because across render they would change so trigger an update. At least here it's not dynamic content so using index should be fine enough.
Apart of that, outside of React keys, yes, using uuid wherever needed is fine.

its-mahdi: I think react is going to complain if you ever use any numerical `index` value as the key, but I will look into it. I agree the extra render, however minor, should not happen if it can be prevented easily.

aurel-l: I think the only time React complains would be duplicate keys (and by the way I already so that warning somewhere, I'll need to investigate it next time I see it). If there is any complaint about index, that would be a linting rule that would be easy enough to override locally.

its-mahdi: Yeah, you might be right. From React docs: `When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort`.

src/entry/state/entryActions.ts
dlrice: I see the issue with using a `Map`. I'm going to commit onto your branch with a rework of the urls.

dlrice: I have just pushed a commit to resolve this.

src/styles/ErrorPages.scss
its-mahdi: Changed this to `rem`.

src/styles/ErrorPages.scss
its-mahdi: Changed this to `rem` just so pixel values aren't used, but the simple `ViewPort` approach didn't work. I need to work a little bit more on this to make this responsive.

its-mahdi:
also added a generic catch-all page and created the `ErrorPage` so it can be used by other pages, whenever possible.

src/pages/errors/ObsoleteEntryPage.tsx
its-mahdi: Addressed this too.

PR #126, opened 2020-05-01
Caching

aurel-l:
- [x] What is the link to the Jira that this PR is for? [TRM-23954](https://www.ebi.ac.uk/panda/jira/browse/TRM-23954)
- [ ] Have you written tests and do these have >= 75% coverage of the code that you are contributing? Nope... Not really sure if I can/should test service workers
- [x] Is it possible to visually evaluate this PR in the browser? If so, what are the steps to do this (eg specific URLs or sections)? Need to browse from a deployed version. `yarn run build:dev` to build, then serve the build folder and browse
- [x] Are there any linting errors that required you to add exceptions? yes, just `console.log` for developers
- [ ] If there are backend requests, have you checked for the existence of the resource's attributes before attempting access? no

Added a service-worker, using the [workbox](https://developers.google.com/web/tools/workbox/guides/get-started) library.
We are intercepting most of the fetch done from the website (even from external libraries), and responding using different strategies from the service-worker, as if it was a proxy.
Caching is done from inside the service-worker.

Some resources use the [stale-while-revalidate approach](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#stale-while-revalidate). But ideally, if the resource provides a version information (version header), we could add keep the resource forever in the cache, and add further logic to drop the whole cache on version change.

Additionnally, some optimisations have been done on the webpack side to generate more chunks but of smaller size, to hopefully invalidate only parts of them on code change or on library update. CSS has been separated into there own CSS files (instead of being inside the JS). Also, this chunking effort needs to be continued across the codebase to have a smaller quantity of assets to be loaded on first load and thus improve loading performance even if the network bottleneck has been removed through caching.


tsconfig.json
```diff
@@ -17,8 +17,9 @@
     "esModuleInterop": true,
     "module": "esnext",
     "jsx": "react",
-    "types": ["jest", "node"]
+    "types": ["jest", "node"],
+    "skipLibCheck": true
```
xwatkins, line 21: Is this safe?

aurel-l: Well, I had to do that not because of performance, but because service worker context is incompatible with normal browser context. And since it was checking the content of all the workbox libraries as if they were executed in browser context it was failing.
Apparently, this doesn't stop checking the codebase types against the libraries types, it just stop checking the libraries themselves internally.
Ideally we would need to have 2 typescript configs for the 2 different contexts, but I didn't manage to have that working.

xwatkins: I see, didn't think about the context issue. I guess not checking the library definitions we create is fairly low risk anyway

aurel-l:
OK, so somehow the squash was ignored or it didn't let me do it even though I asked for it...🤷🏽‍♂️

PR #127, opened 2020-05-05
useLocalStorage

aurel-l:
- [ ] What is the link to the Jira that this PR is for? none
- [x] Have you written tests and do these have >= 75% coverage of the code that you are contributing? yes
- [ ] Is it possible to visually evaluate this PR in the browser? If so, what are the steps to do this (eg specific URLs or sections)? no
- [ ] Are there any linting errors that required you to add exceptions? yes (no-empty-function)
- [ ] If there are backend requests, have you checked for the existence of the resource's attributes before attempting access? NA

added a new custom hook to access localStorage through react hooks logic like this:
```js
const [value, setValue, deleteValue] = useLocalStorage('key', 'default value');
```

It will *eventually* persist to localStorage. `value` is updated right away though.
This should be usable the same way than react's `useState` (compatible API).
One limitation is that values need to be JSON-serialisable.
As localStorage and JSON.parse and JSON.stringify are blocking operation, we try to execute as much of it as we can whenever the browser has time to do so through `requestIdleCallback`.

### DIFF
--- a/src/hooks/useLocalStorage.ts
+++ b/src/hooks/useLocalStorage.ts
@@ -1,0 +1,57 @@
+import { useState, useRef } from 'react';
+
+declare global {
+  interface Window {
+    requestIdleCallback: (callback: () => void) => any;
+    cancelIdleCallback: (handle: any) => void;
+  }
+}
+
+// fallback: noop
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const cancelIdleCallback = window.cancelIdleCallback || (() => {});
+// fallback: execute rightaway
+const requestIdleCallback =
+  window.requestIdleCallback || ((fn: () => any) => fn());
+
+const useLocalStorage = (key: string, initialValue: any = null) => {
+  const handle = useRef();
+
+  const [state, setState] = useState(() => {
+    const stored = window.localStorage.getItem(key);
+    let parsed = null;
+    try {
+      parsed = JSON.parse(stored || '');
+      return parsed;
+    } catch {
+      window.localStorage.setItem(key, JSON.stringify(initialValue));
+      return initialValue;
+    }
+  });
+
+  const setValue = (value: any) => {
+    const valueToStore = typeof value === 'function' ? value(state) : value;
+    setState(valueToStore);
+
+    cancelIdleCallback(handle.current);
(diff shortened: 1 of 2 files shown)


src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,57 @@
...
+    requestIdleCallback: (callback: () => void) => any;
+    cancelIdleCallback: (handle: any) => void;
+  }
+}
+
+// fallback: noop
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const cancelIdleCallback = window.cancelIdleCallback || (() => {});
+// fallback: execute rightaway
+const requestIdleCallback =
```
xwatkins, line 14: Is this polyfilled by babel for Safari?

src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,57 @@
+import { useState, useRef } from 'react';
+
+declare global {
+  interface Window {
```
xwatkins, line 4: wondering if this should go to `src/types` as we might want to define more Window items later?

src/hooks/useLocalStorage.ts
aurel-l: babel doesn't polyfill that, since it's web plateform feature, not JavaScript. We could add a polyfill for that if we want to, or just use the fake polyfill I've added here just because this behaviour would be an improvement, but not really necessary.

src/hooks/useLocalStorage.ts
aurel-l: Well, the types are not correct. It's just to make typescript happy in this file with the fake polyfill. So I don't know

xwatkins: Oh I missed that. So if it's not there it just runs the function instead 👍

src/hooks/useLocalStorage.ts
xwatkins: Oh I missed that. So if it's not there it just runs the function instead 👍

src/hooks/useLocalStorage.ts
xwatkins: Ok.

PR #128, opened 2020-05-06
Restructuring

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23977


xwatkins:
- I wonder if we should nest `blast` under `tools` where we can later have `Dashboard` as a component.
- `src/utils` is empty, are there no shared functions across all functionality?
- a lot of the components under `src/uniprotkb/components/entry` are also used to display column content in the results (they all appear in `ColumnConfiguration.tsx`). I'm wondering whether these should just be at the root of `src/uniprotkb/components`, or grouped in a different directory?
- we should be consistent in our naming of utils files, for instance `KeywordsUtil.tsx` and `XrefUtils.tsx` could be renamed to `keywordsUtil.ts` and `xrefUtils.tsx` (although maybe we don't even need the `utils` suffix anymore?). My preference (debatable) is to name files which have no jsx with `ts`, and start lower case because they are not components.

aurel-l:
>     * I wonder if we should nest `blast` under `tools` where we can later have `Dashboard` as a component.
tools would be at the same level than uniprotkb then, right?

>     * `src/utils` is empty, are there no shared functions across all functionality?
this is what is in `src/shared/utils`, no?

dlrice:
# 1.
> I wonder if we should nest blast under tools where we can later have Dashboard as a component.
>> tools would be at the same level than uniprotkb then, right?

Sounds like a good idea. We would have:
```
├── app
├── index.tsx
├── messages
├── shared
├── tools
│   ├── blast
│   └── align
├── types
└── uniprotkb
```

# 2. 
> * `src/utils` is empty, are there no shared functions across all functionality?
>> this is what is in src/shared/utils, no?

* @xwatkins  I'm guessing this was comment made before `git clean -fd`?
* Yeah, that's right @aurel-l 


# 3.
> a lot of the components under src/uniprotkb/components/entry are also used to display column content in the results (they all appear in ColumnConfiguration.tsx). I'm wondering whether these should just be at the root of src/uniprotkb/components, or grouped in a different directory?

That's a very good point. I was scratching my head over this a bit and was thinking what to name a directory for these shared components. I ended up doing it like this because in both cases of the components usage (results and entry-page) it's still just views of the entry data. I find it's easier to understand what's in a directory if it's completely flat or completely structured (that's my view - does anybody feel otherwise?) so I would prefer to put the entry and results shared components in a directory - I'm just not sure what that be. We could:

1. Create a directory to hold those components that create views of entry-data and are shared between the entry page and the results page:
```
├── column-select
├── customise-table
├── download
├── entry
├── shared | entry-fields | entry-data | entry-views | entry-info | shared-entry
├── query-builder
├── results
└── search
```

2. Add page suffixes
```
├── column-select
├── customise-table-page
├── download-page
├── entry-page
├── query-builder-page
├── results-page
├── search-page
└── entry
```

3. Add a suffix just to entry to distinguish from an entry directory
```
├── column-select
├── customise-table
├── download
├── entry
├── entry-page
├── query-builder
├── results
└── search
```

# 4a.
> we should be consistent in our naming of utils files, for instance KeywordsUtil.tsx and XrefUtils.tsx could be renamed to keywordsUtil.ts and xrefUtils.tsx (although maybe we don't even need the utils suffix anymore?). My preference (debatable) is to name files which have no jsx with ts, and start lower case because they are not components.

That's another good point. I didn't focus on file renaming in this task as it seemed outside the scope of restructuring. But I think know would be a good time to do though as we're already touching everything else in the codebase so might as well. It would be good to establish some conventions and put it in the coding guidelines. Here's what I'm thn

| File Extension | Filename Case                                                          |
| -------------- | ------------------------------------------------------------- |
| scss           | PascalCase ?                                                  |
| tsx            | PascaleCase (must have JSX)                                   |
| ts             | camelCase (no JSX)                                            |
| json           | camelCase (? we use PascalCase but package.json is camelCase) |
| svg            | kebab-case                                                    | 
| directories    | kebab-case                                                    | 

# 4b.
To your other point I agree and could do the following:

remove `Utils?` suffix from:
```
uniprotkb/utils
├── KeywordsUtil.tsx
├── XrefUtils.tsx
├── queryUtils.ts
```
remove `Converter` suffix:
```
├── adapters
│   ├── ExpressionConverter.ts
│   ├── ExternalLinksConverter.ts
│   ├── FamilyAndDomainsConverter.ts
│   ├── FunctionConverter.ts
│   ├── InteractionConverter.ts
│   ├── LiteratureConverter.ts
│   ├── NamesAndTaxonomyConverter.ts
│   ├── PathologyAndBiotechConverter.ts
│   ├── ProteinHighlights.ts
│   ├── ProteinNamesConverter.ts
│   ├── ProteinProcessingConverter.ts
│   ├── SectionConverter.ts
│   ├── SequenceConverter.ts
│   ├── StructureConverter.ts
│   ├── SubcellularLocationConverter.ts
│   ├── UniProtkbConverter.ts
```
remove `Page` suffix:
```
│   ├── components
│   │   ├── error-pages
│   │   │   ├── ErrorHandler.tsx
│   │   │   ├── ErrorPage.tsx
│   │   │   ├── JobErrorPage.tsx
│   │   │   ├── NoResultsPage.tsx
│   │   │   ├── ObsoleteEntryPage.tsx
│   │   │   ├── ResourceNotFoundPage.tsx
│   │   │   ├── ServiceUnavailablePage.tsx
```
remove `Data` suffix (and likewise in other mocks dirs):
```
    ├── __mockData__
    │   ├── ResultFieldsData.ts
    │   ├── StructuredResultFieldsData.json
    │   └── entryModelData.json
```
In addition rename `__mockData__` to `__mocks__` as this seems the official docs suggestion https://jestjs.io/docs/en/manual-mocks

xwatkins:
1. Yep
2. 
3. I would go for 1. for the name I don't think it should have `Entry` in it as it could exist even without the entry existing. Maybe something like `data-views`?
4.a. 👍
4.b. let's just keep it as it is for now. sometimes it's nice to see what something is by just looking at the file name and not the path.

dlrice:
3. `entry-data-views`?

4a. I think I made a mistake in the `scss`. Here's my final version:

| File Extension | Filename Case               |
| -------------- | --------------------------- |
| j\|tsx          | PascaleCase (contains JSX) |
| j\|ts           | camelCase   (no JSX)       |
| json           | camelCase                   |
| s?css          | kebab-case                  |
| svg            | kebab-case                  | 
| directories    | kebab-case                  |

If you're both happy with this I'll put it in the coding-guidelines and standardise the filename/directory cases now.

xwatkins:
3. I don't think we should mention "entry" as they would still exist in the results table if we didn't have an entry.

dlrice:
A valid point. I am equating anything to do with a protein as `entry` (including the row results). I am hesitant to use `data-views` only because it doesn't seem specific enough.

How about `protein-data-views`?

xwatkins:
Yeah? or uniprotkb? don't mind

PR #129, opened 2020-05-12
Remove redux fetch pt ii

xwatkins:


dlrice:
I noticed a few unexpected behaviours:

### 1. Facets persisting when new query is started
Steps to reproduce:
1. Search for `a4` (http://localhost:8080/uniprotkb?query=a4)
2. Click the `reviewed` facet (http://localhost:8080/uniprotkb?query=a4&facets=reviewed:true)
3. The search bar query will still have `a4` but I think that's ok (facets aren't supposed to update the query are they?)
4. Change the search bar query to `mouse` and click search (http://localhost:8080/uniprotkb?facets=reviewed%3Atrue&query=mouse)
5. The facets are still there from the previous search. This is different to the gh-pages behaviour which removes all facets on a new search.


### 2. Loading indicator not immediately shown
Scrolling to the bottom of the first batch of results does not show the loading indicator when loading the second batch. It is visible however if I scroll up a small amount and then back down.

src/uniprotkb/components/entry/EntryPublicationsFacets.tsx
```diff
@@ -1,29 +1,40 @@
...
+  const { selectedFacets } = getParamsFromURL(search);
   const url = getUniProtPublicationsQueryUrl(accession, selectedFacets);
   const { loading, data, status, error } = useDataApi(url);
 
   const addFacet = (name: string, value: string) => {
-    setSelectedFacets([...selectedFacets, { name, value }]);
+    const facet: SelectedFacet = { name, value };
+    history.push({
+      pathname: `/uniprotkb/${accession}/publications`,
+      search: `${facetsAsString([...selectedFacets.concat(facet)])}`,
```
dlrice, line 24: Template string not needed here

src/uniprotkb/components/entry/EntryPublicationsFacets.tsx
```diff
@@ -1,29 +1,40 @@
...
   const removeFacet = (name: string, value: string) => {
-    setSelectedFacets(
-      selectedFacets.filter(
-        selectedFacet =>
-          !(selectedFacet.name === name && selectedFacet.value === value)
-      )
-    );
+    history.push({
+      pathname: `/uniprotkb/${accession}/publications`,
+      search: `${facetsAsString(
```
dlrice, line 31: Template string not needed here

src/uniprotkb/components/results/utils.ts
```diff
@@ -0,0 +1,64 @@
+import queryStringModule from 'query-string';
```
dlrice, line 1: UniProtKB utils are now grouped at `./src/uniprotkb/utils/*`

src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
```diff
@@ -821,15 +821,6 @@ exports[`Entry view should render 1`] = `
             <h3>
               GO Annotations
             </h3>
-            <div
-              class="GoRibbon__container"
-            >
-              <div
-                class="GoRibbon__container__message"
-              >
-                Loading...
```
dlrice, line 830: Just wondering where this went to? I can see the `Loading...` message in the browser.

xwatkins: If this is what I think we don't want to test the GO ribbon component itself so I mocked it (returns an empty object)

src/uniprotkb/components/results/utils.ts
xwatkins: Moved. Although it feels weird to have a generic `utils` with utils files which are then specific to other locations...maybe we can improve this later.

xwatkins:
Fixed 1. and addressed other comments.
For 2. I noticed there is a bit of a visual glitch when scrolling and I think it would take time to investigate. As it's minor I'd be more inclined to create a Jira rather than try to address it now.

src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
dlrice: Ah that's much nicer.

xwatkins: We were getting warnings in the results because of it! 😿

xwatkins: I don't know why I put a crying cat there. Still adjusting to the usage of emojis.

PR #130, opened 2020-05-12
Handle 303 redirects

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-23974
Go to: http://localhost:8080/uniprotkb/Q14301
Make sure you see an in-page message, notifying you about the 303 redirect.

### DIFF
--- a/src/messages/types/messagesTypes.ts
+++ b/src/messages/types/messagesTypes.ts
@@ -15,6 +15,7 @@ export enum MessageFormat {
 export enum MessageTag {
   BLAST_JOB = 'BLAST_JOB',
   DOWNTIME = 'DOWNTIME',
+  REDIRECT = 'REDIRECT',
 }
 
 export type MessageType = {
--- a/src/shared/hooks/__tests__/useDataApi.spec.ts
+++ b/src/shared/hooks/__tests__/useDataApi.spec.ts
@@ -150,4 +150,21 @@ describe('useDataApi hook', () => {
       status: 200,
     });
   });
+
+ test('detect redirect', async () => {
+   mock.onGet(url).reply(200, 'some data');
+   mock.onGet(url2).reply(() => axios.get(url));
+
+   const { result, waitForNextUpdate } = renderHook(() => useDataApi(url2));
+
+   expect(result.current).toEqual({ loading: true });
+
+   await waitForNextUpdate();
+   expect(result.current).toEqual({
+    loading: false,
+    data: 'some data',
+    status: 200,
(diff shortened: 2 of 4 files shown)


its-mahdi:
I wasted many hours to figure out Axios mock responses are not exactly the same as browser responses. This bit looks a little bit hacky, but it's the best I could do:

https://github.com/ebi-uniprot/uniprot-website/blob/handle-303-redirects/src/shared/hooks/useDataApi.ts#L44-L56

aurel-l:
> I wasted many hours to figure out Axios mock responses are not exactly the same as browser responses. This bit looks a little bit hacky, but it's the best I could do:
> 
> https://github.com/ebi-uniprot/uniprot-website/blob/handle-303-redirects/src/shared/hooks/useDataApi.ts#L44-L56

After investigating I saw there was an issue with casing of `responseURL`/`responseUrl` in the mocked response (so, part of `axios-mock-adapter`) compared to the axios response in the browser.
Ideally I think that'd need to be raised as an issue in the library, because it's weird to have a condition in the code just to have the tests work, that defeats a bit the purpose of doing tests.
Anyone else having an opinion on what we should do. Drop the corresponding test? Something else?
If we keep the test and the condition for the test to work, maybe I'd write it a more similar way just to make it clear it's the casing itself which is different, and not depend on another part of the axios API.
```

      if (
        action.response &&
        action.response.request.responseURL &&
        action.response.request.responseURL !== action.originalURL
      ) {
        newState.redirectedTo = action.response.request.responseURL;
      } else if (
        // Issue with casing in axios-mock-adapter?
        action.response &&
        action.response.request.responseUrl &&
        action.response.request.responseUrl !== action.originalURL
      ) {
        newState.redirectedTo = action.response.request.responseUrl;
      }
```

its-mahdi:
I didn't realise that's just an alphabetic casing issue. I think your solution is better. Happy to reflect that in the code, if no one else has any better suggestions.

xwatkins:
Agree I think someone should create a bug report and we could use @aurel-l 's code in the meantime with a link to the bug report in the comment:
https://github.com/ctimmerm/axios-mock-adapter/blob/master/src/handle_request.js#L30 is the offending line I think

dlrice:
Why aren't other people struggling with this bug? I don't understand why not?!

its-mahdi:
> Why aren't other people struggling with this bug? I don't understand why not?!

Probably because I am the only lucky one who hits the jackpot all the time!

PR #132, opened 2020-05-15
Blast part i

xwatkins:
This is the old "revived" branch. Some things to change in there following our discussions, but I think it provides a good base to start from.


aurel-l:
Not sure if "approving" here is the right thing to do, as the changes needed are meant to be implemented of the current sprint anyway and this PR has been meant as a starting point

src/blast/data/BlastFormData.ts
```diff
@@ -0,0 +1,125 @@
+export type BlastFormValue = {
```
dlrice, line 1: In the restructure I used a `config` directory for this sort of file.

src/blast/BlastForm.tsx
```diff
@@ -0,0 +1,139 @@
+import React, { FC, Fragment, useState } from 'react';
```
dlrice, line 1: We had previously discussed placing blast under a tools directory:
> ```
> ├── app
> ├── index.tsx
> ├── messages
> ├── shared
> ├── tools
> │   ├── blast
> │   └── align
> ├── types
> └── uniprotkb
> ```

xwatkins:
Yes sorry, this wasn't really supposed to be a real PR but somewhere to hack this back into the goal branch, so maybe let's just make the changes directly on there then when we're happy we can just merge it?

xwatkins:
I've updated the directory structure
I think @aurel-l will rip up the whole state thing so left it alone for now 🙂

PR #133, opened 2020-05-19
Model change 3 trm 24111

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24111
In addition:
* Create common __mocks__/mockApi for uniprotkb
* Add didCancel check in useDataApi hook.
Update dev server to use the following:
* http://wp-np2-be:8095/uniprot/api/configure/uniprotkb/search_terms
* http://wp-np2-be:8095/uniprot/api/configure/uniprotkb/result-fields
* http://wp-np2-be:8095/ for devPrefix for the suggester. 
* http://wp-np2-be:8090/ for the rest of the endpoints
Then reference the jira and investigate the change the column selection and query builder sections.
n/a
Yes


src/shared/hooks/useDataApi.ts
```diff
@@ -86,18 +90,23 @@ const useDataApi = (url?: string): State => {
...
         dispatch({ type: ActionType.ERROR, error });
       }
     );
 
     // handle unmounting of the hook
     // eslint-disable-next-line consistent-return
-    return () => source.cancel();
+    return () => {
+      source.cancel();
+      didCancel = true;
```
aurel-l, line 108: So did you have any concrete issue with that?
Because I remember on the first version of that file I had that same logic implemented, but after a discussion with @its-mahdi we came to the conclusion that `axios.isCancel(error)` was enough to catch a case where the fetch would have been canceled.

src/uniprotkb/components/column-select/__tests__/ColumnSelectContainer.spec.tsx
```diff
@@ -1,21 +1,15 @@
...
 import { ColumnSelectTab } from '../../../types/resultsTypes';
 import { Column } from '../../../types/columnTypes';
-
-const mock = new MockAdapter(axios);
-mock
-  .onGet(mockResultFieldsApi.request)
-  .reply(200, mockResultFieldsApi.response);
+import structuredResultFieldsData from './__mocks__/structuredResultFieldsData.json';
+import resultFields from '../../../__mocks__/resultFields.json';
+import '../../../__mocks__/mockApi';
```
aurel-l, line 12: Multiple things here:
 - In general I'm not a big fan of imports without assigning to a variable / side-effect modules
 - we use `axios-mock-adapter` across multiple tests, and with multiple type of responses. I'm not sure if the logic needed for this specific test should be abstracted in a mock file higher in the directory hierarchy, or shoudl be put in a mock file at all.

aurel-l: Ah, my bad, I see the same file is indeed used in multiple places. As long as you're putting it only as high in the directory as it is needed I'm fine with it.
Also, you'd need to make sure the mock doesn't interact with other places where we're using a mock for axios (as in importing the mock itself, to be able to call its `restore` method to call it in the `afterAll` jest function)

src/shared/hooks/useDataApi.ts
dlrice: Yeah, I only put this in because I was getting some issues thrown up when running jest tests.

aurel-l: All good then 👍

src/uniprotkb/components/column-select/__tests__/ColumnSelectContainer.spec.tsx
dlrice: > Ah, my bad, I see the same file is indeed used in multiple places. As long as you're putting it only as high in the directory as it is needed I'm fine with it.

Moved it a bit lower.

> Also, you'd need to make sure the mock doesn't interact with other places where we're using a mock for axios (as in importing the mock itself, to be able to call its restore method to call it in the afterAll jest function)

I've just tested and we can do `mockApi.reset()` if needed. Out of curiosity, when do you think this would be needed in an `afterAll`? Creating a new mock `const mock = new MockAdapter(axios);` in another test file will create a new instance.

aurel-l: Well it's a bit of me never entirely trusting how mocks affect each other across tests, so if there's a way to clean them up afterwards I would tend to try to use it.
I don't think it is needed now, as it is working now, but I wouldn't want overlooking that at that stage to cause an unexpected error in the future because of the order current, or new, tests are run.

PR #136, opened 2020-06-08
More layout

aurel-l:
- Add space for footer
 - Create and apply a Single-Column layout where needed
 - Fix regression with SideBar layout (side menu now scrolling as expected)


src/uniprotkb/components/query-builder/AdvancedSearchContainer.tsx
```diff
@@ -56,8 +55,10 @@ type State = {
 export class Search extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
-    const { queryString } = props;
-    this.state = { queryString };
+
+    // const { queryString } = props;
```
xwatkins, line 59: Do we want to keep these commented out lines?

src/uniprotkb/components/query-builder/AdvancedSearchContainer.tsx
```diff
@@ -74,12 +75,12 @@ export class Search extends Component<Props, State> {
...
   }
 
-  componentDidUpdate(prevProps: Props) {
-    const { queryString: prevQueryString } = prevProps;
-    const { queryString } = this.props;
-    if (prevQueryString !== queryString) {
-      this.setState({ queryString });
-    }
+  componentDidUpdate() {
+    // const { queryString: prevQueryString } = prevProps;
```
xwatkins, line 79: Doe we need this?

.eslintrc
```diff
@@ -19,41 +19,44 @@
     "sourceType": "module"
   },
```
xwatkins, line 20: I'm guessing this was mostly some re-arranging?

aurel-l: Yes, reordering alphabetically, and change rules from numbers to the corresponding string representation (`off`, `warn`, `error`)

src/uniprotkb/components/query-builder/AdvancedSearchContainer.tsx
aurel-l: I wasn't sure what the component was doing, and it seemed functional after commenting big chunks of that file. I assumed it was leftover code from a while ago, but I mainly commented it to make TypeScript API but keep the code handy in case something was really necessary.

xwatkins: Ok - just merge it and I'll have a look

PR #138, opened 2020-06-09
Job message link

dlrice:
**Additional change in this PR**
When submitting in BlastForm, dispatch createJob after sleep to prevent error pop-up message immediatly showing up when viewing the BlastForm which causes update-on-unmounted error.

**Job message link**
n/a
1. Run a valid blast job
2. Navigate away from the dashboard
3. Wait for the pop-up message
4. Click the link
n/a
n/a


src/tools/blast/components/BlastForm.tsx
```diff
@@ -228,15 +228,18 @@ const BlastForm = () => {
...
-    });
+    sleep(1000)
+      .then(() => {
+        history.push(LocationToPath[Location.Dashboard], { parameters });
+      })
+      .then(() => {
+        // we emit an action containing only the parameters and the type of job
+        // the reducer will be in charge of generating a proper job object for
+        // internal state
+        dispatch(actions.createJob(parameters, 'blast', jobName));
```
aurel-l, line 241: Why has this been moved here? It effectively delays by one second triggering the logic to submit the job to the server but I don't see why.

PR #140, opened 2020-06-09
Refactor state

xwatkins:
Break up form state into smaller state items.

### DIFF
--- a/src/tools/blast/components/BlastForm.tsx
+++ b/src/tools/blast/components/BlastForm.tsx
@@ -37,7 +37,7 @@ import { Tool } from '../../types';
 import * as actions from '../../state/toolsActions';
 
 import { LocationToPath, Location } from '../../../app/config/urls';
-import initialFormValues, {
+import defaultFormValues, {
   BlastFormValues,
   BlastFormValue,
   BlastFields,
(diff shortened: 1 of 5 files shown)


src/tools/blast/components/BlastForm.tsx
```diff
@@ -67,31 +67,32 @@ const getAutoMatrixFor = (sequence: string): FormParameters['matrix'] => {
...
+}> = ({ formValue, updateFormValue }) => {
+  if (!formValue) {
     return null;
   }
+  const label = BlastFields[formValue.fieldName as keyof typeof BlastFields];
   return (
     <section className="blast-form-section__item">
-      <label htmlFor={type}>
-        {type}
+      <label htmlFor={label}>
```
aurel-l, line 79: The label here is wrapping the corresponding select element, so no need to had extra ids and for attributes to link them.

src/tools/blast/components/BlastForm.tsx
```diff
@@ -108,83 +109,98 @@ const BlastForm = () => {
   const dispatch = useDispatch();
   const history = useHistory();
 
+  const getInitialValues = () => {
```
aurel-l, line 112: this probably should be wrapped in a react `useMemo` function as it's quite heavy logic on each render for something used only once on mount

src/tools/blast/config/BlastFormData.ts
```diff
@@ -58,7 +58,7 @@ export default Object.freeze({
     selected: '',
   }),
   [BlastFields.targetDb]: Object.freeze({
-    fieldName: 'database',
+    fieldName: 'targetDb',
```
aurel-l, line 61: This should probably have had been changed the other way around, as 'database' is what is used in the internal job parameters (`FormParameters`) and on the server (`ServerParameters`)

src/tools/blast/components/BlastForm.tsx
xwatkins: Done

src/tools/blast/components/BlastForm.tsx
xwatkins: Done

src/tools/blast/config/BlastFormData.ts
xwatkins: Done

PR #142, opened 2020-06-12
Blast results table

xwatkins:
Caveat: there is a bug in `protvista-track` which is preventing the centering of the feature. Either link it using the nightingale branch `fix-track-centering` or remove the "height" parameter.

https://www.ebi.ac.uk/panda/jira/browse/TRM-24308

Run Blast job, go to results


aurel-l:
Probably not directly part of the code in this PR directly but more related to the used component in nightingale. I've noticed the rectangles were blurry on my external monitor (non-retina) because of the default anti-aliasing.
Setting `shape-rendering="crispEdges"` on the svg solves this issue.
See difference in the following image when applied (top) or not (might need zooming in to see the difference)
Screenshot 2020-06-15 at 09 25 07

xwatkins:
Happy that's the issue, I thought it was my eyesight going. Will add that in Nightingale.

PR #143, opened 2020-06-16
Slide panel

xwatkins:
Adding basic initial side panel component. If the PR is accepted I will then apply the same to the customise columns on results and the Download as part of the Blast pages.


its-mahdi:
Does the **Download** button on the search result page has to act like a toggle button? Or could it be like a toggle button? Because I think being able to close the window just by clicking on the same button again would be quite user-friendly.

its-mahdi:
The columns in the accordion menu might need to be responsive. In a normal to small size screen the 3-column layout makes the text wrap and it all looks confusing and somewhat broken.

its-mahdi:
The **Taxonomic lineage** ID is duplicated, causing React to throw warnings:

    [DOM] Found 2 elements with non-unique id #checkbox-lineage

I don't know how easy it can be fixed -- specially if it is all coming from server, but worth having a look at as I suspect it can cause some bugs in the future if IDs are the same.

its-mahdi:
When trying to customise the column order, `react-beautiful-dnd` throws a warning:

```react-beautiful-dndDroppable: unsupported nested scroll container detected.A Droppable can only have one scroll parent (which can be itself)Nested scroll containers are currently not supported.We hope to support nested scroll containers soon: https://github.com/atlassian/react-beautiful-dnd/issues/131👷‍ This is a development only message. It will be removed in production builds.```

its-mahdi:
When I press on the **Preview** button, the loading and showing the preview happens off-canvas. Even as an experienced web user I did notice the change in the scroll-bar on the second time. I thought the preview button is not functional when I tried it the first time.

So, maybe it can auto-scroll to the end of the column when and while the preview is loading?

PR #145, opened 2020-06-16
Add scroll to view, add sequence to table

xwatkins:
This is a (yet to be created) subtask of https://www.ebi.ac.uk/panda/jira/browse/TRM-24238

There's been requests (on the covid platform) to show the sequence in some form as part of the `protvista-datatable`. This PR offers a simple solution "for now" while we decide how we want to integrate the rest of requirements, and has been discussed with @SangyaP 

Will cherry-pick the PR on covid-19 when approved.

### DIFF
--- a/src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
+++ b/src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
@@ -2030,7 +2030,7 @@ exports[`Entry view should render 1`] = `
             Features
           </h3>
           <protvista-manager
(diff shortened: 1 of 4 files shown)


aurel-l:
Please point me to an example protein and where in the entry page I'm supposed to look at to check what is new/different in this PR

xwatkins:
- Go to `.../uniprot-website/uniprotkb/P05067`
- Click on Protein Processing in the in-page-nav
- Notice the + icon on each row of the protvista-datatable
- Clicking on it triggers the opening of another row, containing the sequence

Note: the styling and layout for this section is a bit off (it doesn't look like it belongs to the same row, doesn't highlight etc...) but this will need to be fixed in Nightingale

PR #146, opened 2020-06-17
Blast download panel

xwatkins:
Wrap the direct BLAST download functionality into the new `SidePanel` container. Use dropdown for format selection.


src/tools/blast/components/results/BlastResultsButtons.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const [displayDownloadPanel, setDisplayDownloadPanel] = useState(false);
+
+  return (
+    <>
+      {displayDownloadPanel && (
+        <Suspense fallback>
+          <SidePanel>
+            <BlastResultDownload
+              id={jobId}
+              onClose={() => setDisplayDownloadPanel(!displayDownloadPanel)}
```
aurel-l, line 21: Why not simply set to `false`?

xwatkins: In the first PR on side panel, @its-mahdi mentioned that the Download button can still be visible and we both thought you'd expect it to toggle the panel if you clicked on it again.

aurel-l: Well, that's definitely OK for the other place where you call `setDisplayDownloadPanel` (the download button) but here the `onClose` is quite clear about the intent of what should be done within the function.

xwatkins: Good point, renamed to `onToggleDisplay`

PR #147, opened 2020-06-17
No blast/job results

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24316

No new tests, but updated other snapshots as it was required.

If you go to the dashboard when you don't have any jobs in the page, you should see an illustration and a warning message underneath.


its-mahdi:
This is failing right now because it requires the latest version of `Message` component from Franklin, however, that's not release yet.

PR #148, opened 2020-06-18
Disabled facets

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24347

This is a very minor feature.

Open a BLAST result page. Then navigate to Text Output or Tool Input tabs. You should see a grey box on the sidebar instead of the facets.

### DIFF
--- a/src/tools/blast/components/results/BlastResult.tsx
+++ b/src/tools/blast/components/results/BlastResult.tsx
@@ -33,6 +33,14 @@ const BlastResultToolInput = lazy(() =>
   )
 );
 
+enum TabLocation {
+  Overview = 'overview',
+  Taxonomy = 'taxonomy',
+  HitDistribution = 'hit-distribution',
+  TextOutput = 'text-output',
+  ToolInput = 'tool-input',
+}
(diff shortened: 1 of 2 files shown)


src/tools/blast/components/results/BlastResult.tsx
```diff
@@ -112,16 +110,38 @@ const BlastResult = () => {
...
+
+  // Deciding what should be displayed on the sidebar
+  const facetsSidebar = <BlastResultSidebar loading={apiLoading} data={data} />;
+  const emptySidebar = (
+    <div className="sidebar-layout__sidebar-content--empty" />
+  );
+  let sidebar;
+
+  switch (match.params.subPage) {
+    case 'text-output':
```
xwatkins, line 129: Could define these as variables/symbols to avoid potential typo related issues later

its-mahdi:
I used an enum now for the tab locations. The sidebar extended at 100% looks a little bit odd to me. It feels like you can scroll it, but I have made 100% now, see if that looks better.

PR #149, opened 2020-06-18
Update button activity selection, use DataTable

xwatkins:
- Button active/disabled on results page and blast page linked to selection in data table
- use DataTable component for Blast results

### DIFF
--- a/src/uniprotkb/components/results/ResultsButtons.tsx
+++ b/src/uniprotkb/components/results/ResultsButtons.tsx
@@ -58,17 +58,15 @@ const ResultsButtons: FC<{
       <div className="button-group">
         <button
           type="button"
-          className={`button tertiary ${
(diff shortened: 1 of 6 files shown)


src/uniprotkb/components/results/ResultsButtons.tsx
```diff
@@ -56,10 +56,20 @@ const ResultsButtons: FC<{
         </Suspense>
       )}
       <div className="button-group">
-        <button type="button" className="button tertiary disabled">
+        <button
+          type="button"
+          className={`button tertiary ${
+            selectedEntries.length === 1 ? '' : 'disabled'
```
aurel-l, line 62: All of these should be using the button's native `disabled` attribute. I'd even go as far as saying this should be used **instead** of the class.

PR #153, opened 2020-06-22
implement resubmit from server parameters

aurel-l:
- [x] What is the link to the Jira that this PR is for? [TRM-24356](https://www.ebi.ac.uk/panda/jira/browse/TRM-24356)
  Go to a finished blast job, click "resubmit", you should be taken to the Blast submission form with the prefilled fields

### DIFF
--- a/src/tools/blast/components/results/BlastResultsButtons.tsx
+++ b/src/tools/blast/components/results/BlastResultsButtons.tsx
@@ -45,7 +45,6 @@ const ResubmitButton: FC<ResubmitButtonProps> = ({ inputParamsData }) => {
           getSuggesterUrl(uniProtKBApiUrls.organismSuggester, idCleaned)
         ).then((response) => {
           const firstSuggestion = response?.data?.suggestions?.[0]?.value;
-          console.log(firstSuggestion, idCleaned);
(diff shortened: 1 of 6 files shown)


src/tools/blast/components/results/BlastResultsButtons.tsx
```diff
@@ -1,11 +1,98 @@
...
+            );
+          }
+        });
+      });
+
+    try {
+      // wait up to 2 seconds to get the information
+      await Promise.race([Promise.all(taxonRequests), sleep(2000)]);
+    } catch (_) {
+      /* */
```
xwatkins, line 61: Should we just use the taxids in that case?

aurel-l: I just thought it would look nicer to display the names in the same way that if you would have entered them through the taxon search box.
Also, I'm making sure that this is optionnal in the code, in the sense that if it fails to retrieve the information, or if takes too long, it is not blocking the flow of things and it would fallback to displaying the taxid as the label.

xwatkins: Ok, hadn't seen that it was optional.

PR #155, opened 2020-06-24
Blast filtering logic

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24302
No. Please make suggestions where and what to test.
Go to a Blast results page. You should see the histograms on the sidebar.


src/tools/blast/components/results/BlastResult.tsx
```diff
@@ -42,6 +42,11 @@ const BlastResultToolInput = lazy(() =>
     /* webpackChunkName: "blast-result-tool-input" */ './BlastResultToolInput'
   )
 );
+const BlastResultHitDistribution = lazy(() =>
+  import(
+    /* webpackChunkName: "blast-result-tool-input" */ './BlastResultHitDistribution'
```
aurel-l, line 47:
```suggestion
    /* webpackChunkName: "blast-result-hit-distribution" */ './BlastResultHitDistribution'
```

src/tools/blast/components/results/BlastResultsParametersFacets.tsx
```diff
@@ -0,0 +1,164 @@
...
+    );
+  };
+
+  const onBlastParamChange = (paramName: string, values: number[]) => {
+    addFacet(paramName, values.join('-'));
+  };
+
+  const binSize = 100;
+
+  const histogramSettings = {
```
xwatkins, line 91: These could go up to parent component and be shared with the tab visualisation

its-mahdi:
@aurel-l Do you have any suggestion if code needs to be moved around? I'm not sure if everything there is in the right files. There are also some similar functionality between the `Facets` component and the one that I have implemented. e.g. `addFacet`, etc. Tried to combine these two together, but the Facets component has built with a different goal in mind I guess, so it proved to be difficult.

Let me know if you think any typings should be added or moved into different files too.

its-mahdi:
Another branch ahead of this PR is already merged to 08.

PR #159, opened 2020-07-01
Align basic form

aurel-l:
- [x] What is the link to the Jira that this PR is for? [TRM-24390](https://www.ebi.ac.uk/panda/jira/browse/TRM-24390) and [TRM-24391](https://www.ebi.ac.uk/panda/jira/browse/TRM-24391)
- [x] Have you written tests and do these have >= 75% coverage of the code that you are contributing? not sure it covers 75%, but there are some tests
   - Go to 'Align', enter sequences, submit job. Will submit the job and present it in the dashboard.
   - Select multiple accession in data tables and click the align button

First part of the Align jiras, includes a basic form (no accession to fasta field yet), selected entries to align feature, and job submission (no result page yet). Also, refactoring to make code more generics for BLAST and Align jobs (and future types of jobs)


aurel-l:
> Wondering if `getCheckJobStatus`, `getSubmitJob` etc.. should be under `state` instead of `utils` as they're only used by the middleware?
Yep, I could definitely move them there.
> What format(s) does it accept to separate sequences? Was trying with line return but keep getting an error. Wondering if checks should be added to `SequenceSubmission`?
What I found was working is to have headers between the sequences. For example, try to select multiple entries from a result page and create an align job to see what FASTA format I made it generate.

src/tools/state/getCheckJobStatus.ts
```diff
@@ -0,0 +1,118 @@
...
+    const { data: status } = await fetchData<Status>(
+      urlConfig.statusUrl(job.remoteID),
+      { Accept: 'text/plain' },
+      undefined,
+      { responseType: 'text' }
+    );
+    // get a new reference to the job
+    let currentStateOfJob = getState().tools[job.internalID];
+    // check that the job is still in the state (it might have been removed)
+    if (!currentStateOfJob) return;
```
its-mahdi, line 30: Same here, removing this one-liners would be nice.

src/tools/state/getCheckJobStatus.ts
```diff
@@ -0,0 +1,118 @@
...
+      const response = await fetchData<BlastResults>(
+        urlConfig.resultUrl(job.remoteID, 'jdp?format=json')
+      );
+
+      const results = response.data;
+
+      // get a new reference to the job
+      currentStateOfJob = getState().tools[job.internalID];
+      // check that the job is still in the state (it might have been removed)
+      if (!currentStateOfJob) return;
```
its-mahdi, line 66: and this one-liner here as well.

src/app/state/rootInitialState.ts
```diff
@@ -9,7 +9,7 @@ import messagesInitialState, {
 } from '../../messages/state/messagesInitialState';
 import toolsInitialState, {
   ToolsState,
-} from '../../tools/state/toolsInitialState';
+} from '../../tools/types/toolsInitialState';
```
dlrice, line 12: A minor quibble: `toolsInitialState` is not strictly state though it is stored in the `types` directory.

src/tools/adapters/parameters.ts
```diff
@@ -0,0 +1,172 @@
...
+          taxids,
+          stype,
+          sequence,
+          database,
+        } = serverParameters as PublicServerParameters[JobTypes.BLAST];
+
+        if (scores !== alignments) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            `mismatch between number of scores (${scores}) and number of alignments (${alignments})`
```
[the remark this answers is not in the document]
aurel-l: This situation could only happen in these cases:
 - There is a regression in our logic to submit jobs
 - Someone submitted a job through the service directly (e.g. command line, or other tool implemented by someone else) and is try to view it through our interface with their job ID

This shouldn't really generate an error on our side but it might create a mismatch between the number of results displayed and the number the user expects to see.

src/tools/align/components/AlignForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+    setSequence(defaultFormValues[AlignFields.sequence]);
+    setJobName(defaultFormValues[AlignFields.name]);
+  };
+
+  // the only thing to do here would be to check the values and prevent
+  // and prevent submission if there is any issue
+  const submitAlignJob = (event: FormEvent | MouseEvent) => {
+    event.preventDefault();
+
+    if (!sequence) return;
```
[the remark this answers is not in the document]
aurel-l: Just force of habit, changing it now

src/app/state/rootInitialState.ts
aurel-l: Yes, looking at this specific file it's quite obvious this is not following the rest of the app's structure. I'll change that now.

PR #165, opened 2020-07-09
Blast hsp panel

xwatkins:

### DIFF
--- a/src/tools/blast/components/results/HSPDetailPanel.tsx
+++ b/src/tools/blast/components/results/HSPDetailPanel.tsx
@@ -1,0 +1,29 @@
+import React, { FC } from 'react';
+import ProtvistaTrack from 'protvista-track';
+import ProtvistaNavigation from 'protvista-navigation';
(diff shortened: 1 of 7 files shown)


src/tools/blast/components/results/HSPDetailPanel.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <section className="hsp-row">
+          <protvista-manager attributes="displaystart displayend">
+            <protvista-navigation length={hsp_align_len} />
+            <protvista-msa ref={setMSAData} length={hsp_align_len} />
+          </protvista-manager>
+        </section>
+        <section className="hsp-row">
+          {/* 
+          TODO listen to "highlight" event from block above and set on these 2 tracks,
+          working as protvista-manager
```
dlrice, line 96: Does this comment mean that we will use a `protvista-manager` around these two tracks or that we need to emulate its event management within the HSPDetailPanel (using eg react state)?

PR #174, opened 2020-07-30
Align basic result page

aurel-l:
  - [TRM-24392](https://www.ebi.ac.uk/panda/jira/browse/TRM-24392)
  - [TRM-24463](https://www.ebi.ac.uk/panda/jira/browse/TRM-24463)
  - [TRM-24393](https://www.ebi.ac.uk/panda/jira/browse/TRM-24393)
  - Submit an Align job, navigate to its result page
- [x] Are there any linting errors that required you to add exceptions? (a temporary one)

Basic Align result page:
 - Added routing logic for align result pages and subpages
 - Moved and refactored BLAST components that could be reused for Align
 - Added some placeholder for the main tabs of Align result tabs
 - "Tool input" and "Text output" are functionnal
 - Download is functional (formats specific for Align)


src/tools/components/ResultDownload.tsx
```diff
@@ -0,0 +1,166 @@
...
+const ResultDownload = memo<ResultDownloadProps>(
+  ({
+    jobType,
+    id,
+    onToggleDisplay,
+    nHits,
+    isTableResultsFiltered,
+    isTableRowSelected,
+  }) => {
+    const [fileFormat, setFileFormat] = useState(options[jobType][0].format);
```
its-mahdi, line 97: This is a little bit of nitpicking really, but I was thinking that if we add a function like `getDefaultOption(jobType)` then we don't have to rely on `options[jobType][0]`. The reason being if we are going to use this in other places, it would be much more reliable to hide the implementation inside a function.

Up to you if you want to keep it the way it is now or not though. It's just a suggestion.

src/tools/styles/ToolsResult.scss
```diff
@@ -0,0 +1,15 @@
+.tools-result {
+  & .tabs__header {
+    z-index: 3;
```
its-mahdi, line 3: I have always set the `z-index` value within a band, that is, 100 for containers, 500 for buttons, 999 for modals. This is so if I need to put something between 100 and 200, I have got an index available. This is again nitpicking, but I thought I will mention it regardless.

src/tools/align/components/results/AlignResult.tsx
```diff
@@ -0,0 +1,271 @@
...
+  // >(
+  //   useMemo(
+  //     () =>
+  //       getAccessionsURL(accessionsFilteredByLocalFacets, {
+  //         selectedFacets: urlParams.selectedFacets,
+  //         facets: [],
+  //       }),
+  //     [accessionsFilteredByLocalFacets, urlParams.selectedFacets]
+  //   )
+  // );
```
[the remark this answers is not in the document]
aurel-l: Since we won't have filtering on Align I guess I can remove it indeed

src/tools/align/components/results/AlignResult.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  // Note: this function is duplicated in ResultsContainer.tsx
+  // const handleSelectedEntries = (rowId: string) => {
+  //   const filtered = selectedEntries.filter((id) => id !== rowId);
+  //   setSelectedEntries(
+  //     filtered.length === selectedEntries.length
+  //       ? [...selectedEntries, rowId]
+  //       : filtered
+  //   );
+  // };
```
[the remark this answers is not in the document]
aurel-l: I guess this one can stay because we will be need to select sequences from the ones aligned to use in the action bar (to blast or align again)

src/tools/styles/ToolsResult.scss
aurel-l: Yeah, I can put it higher. It was just what I needed to get over the nightingale track, but it definitely doesn't hurt ot have a bit of margin to put other values in between if needed

PR #175, opened 2020-07-30
2020 08 package updates

dlrice:
Once approved will merge into 2020-09 sprint branch and covid client.


dlrice:
Note: couldn't update `history` because of basename issue: https://github.com/ReactTraining/history/issues/810

PR #182, opened 2020-08-24
use useSize, and stagger rendering of tracks

aurel-l:
- Use useSize hook to dynamically change the size of chunks in HSP wrapped view
 - Delay rendering of tracks after the 4th chunk to a bit later, to prioritise what is visible

Issue: probably with the MSA viewer track in Nightingale
 - I had to force the re-rendering of the tracks themselves, because they didn't really like getting their data changed on-the-fly -> end up with the right end of the track truncated somehow, and misaligned.


dlrice (approved):
The resize and delayed-loading works quite nicely. I did however notice for the delayed loading of the wrapped rows that if I immediately attempt to repeatedly scroll to the bottom I get to the fourth chunk but can't proceed further (as expected). If I scroll for 10 seconds the additional chunks still don't load -- it seems the scrolling blocks the debounce timer. I am wondering if it would be useful to have a place holder/loader (of approximate size of each row) where the chunk should appear so that the scroll will extend to approximately the right place?

aurel-l:
Yes, the logic to render the rest of the chunks is to wait for 1 second, then to wait for the browser to have "available time" (uses [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) in the background, if available in the browser). I could provide it a timeout, of "wait, at most, 500ms more".
I guess you scrolling non-stop would have your browser delay as much as possible this "idle callback".

Good point for the placeholder, I can definitelly add that easily, I'll modify it today.

aurel-l:
I've now abstracted more the logic to know how many items of a list to render at any given time, put it into a hook, and used it also on the Blast hits themselves, as they were taking a looooooong time to render their Nightingale tracks at it was visibly slowing down the first rendering of the page.

PR #183, opened 2020-08-25
Enrich align with features

aurel-l:
- [x] What is the link to the Jira that this PR is for? [TRM-24476](https://www.ebi.ac.uk/panda/jira/browse/TRM-24476)

 - Now we have available at the top-level of an Align result the `sequenceInfo` object containing a map of accession to enriched data (parsed sequence + features).
   - Only for sequences for which the accession was extractable from the header, and for which the sequence matches what we have in our API.
 - Used this to generate links for the labels in the phylogenetic tree and percentage identity matrix tabs


xwatkins (approved):
Looks good. Was wondering whether there might be other cases in the future where we will have to parse XML and if so whether `inputParamsXMLToObject.ts` could be more generic but can't think of any right other cases right now.

PR #189, opened 2020-09-08
Multiple alignment overview tracks with a certain height

its-mahdi:


src/tools/components/MSAView.tsx
```diff
@@ -123,17 +129,11 @@ const MSAView: FC<MSAViewProps> = ({
...
-        node.data = getFullAlignmentSegments(alignment)
-          .map(({ trackData }) => trackData)
-          .flat();
-      }
-    },
-    [alignment]
-  );
+  useEffect(() => {
+    if (alignment) {
+      setOverviewTracksData(getFullAlignmentSegments(alignment));
```
aurel-l, line 134: This is not async nor resulting from a user action, so that should'nt really be in a useEffect and setting as state directly dependent of a prop.
That should be in the main part of the render function, and if the calculation is expensive, that should be wrapped by a useMemo

PR #191, opened 2020-09-09
Do not fix page headers trm 24703

dlrice:
The approach of hard-coding the height of the top of the blast's table's header cells is hacky and I'm happy to change this - just not sure how to do that yet.

https://www.ebi.ac.uk/panda/jira/browse/TRM-24703
tests should have changed

- UniProtKB results card & table view
- BLAST table results

n/a
n/a


aurel-l (approved):
Looks good to me.

Changed the colour of the tabs' background because having the half-transparent thing didn't make sense anymore I believe.

I'd get someone else's opinion about also adding some space at the top of the sidebar. Because now it looks a bit stuck to the top, especially in entry pages and search result pages.

PR #192, opened 2020-09-15
Uniref types

its-mahdi:
Just some basic uniref entry types

### DIFF
--- a/src/uniref/adapters/uniRefConverter.ts
+++ b/src/uniref/adapters/uniRefConverter.ts
@@ -5,8 +5,8 @@ enum GeneOntologyAspect {
 }
 
 enum MemberIdType {
-  UNIPROTKB_SWISSPROT = 'UniProtKB Reviewed (Swiss-Prot)',
-  UNIPROTKB_TREMBL = 'UniProtKB Unreviewed (TrEMBL)',
+  UNIPROTKB_REVIEWED = 'UniProtKB Reviewed (Swiss-Prot)',
+  UNIPROTKB_UNREVIEWED = 'UniProtKB Unreviewed (TrEMBL)',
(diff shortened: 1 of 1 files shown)


src/uniref/adapters/uniRefConverter.ts
```diff
@@ -0,0 +1,77 @@
...
+  UniRef90 = 'UniRef90',
+  UniRef50 = 'UniRef50',
+}
+
+type GeneOntologyEntry = {
+  ancestors: GeneOntologyEntry[];
+  aspect: GeneOntologyAspect;
+  name: string;
+  id: string;
+} | null;
```
aurel-l, line 23: Not sure about that, because lower in this file you can have an array of those, or for its own ancestors. Which means `[null, null, null, null]` is valid. Shouldn't you just put `goTerms?: GeneOntologyEntry[]`, or `goTerms: GeneOntologyEntry[] | null`?

aurel-l (changes requested):

src/uniref/adapters/uniRefConverter.ts
its-mahdi: So I had a chat with Leo yesterday and for UniRef they will return `ancestors: [null]`, always. But the ancestors is actually a nested array of `GeneOntologyEntry`, at least in theory. So, sticking to the data, we should defined it as `ancestors: null[]` but that just looks odd. That's why I made the whole thing nullable. And although we won't ever get `[null, null,...]` the data that we are going to get is `[null]`, so not ideal really.

aurel-l (approved):

its-mahdi:
Yeah, that makes sense! I will change those and merge it afterwards!

PR #193, opened 2020-09-15
Use nightingale type definition

xwatkins:
Use the new type definition from the nightingale variation adapter component.


src/uniprotkb/components/protein-data-views/FeaturesTableView.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { Fragment, useCallback, useState, FC } from 'react';
 import { TemplateResult } from 'lit-html';
 import ProtvistaDatatable from 'protvista-datatable';
+import { Feature as VariantFeature } from 'protvista-variation-adapter/dist/es/variants';
```
aurel-l, line 4: Not sure if that should be imported from the dist folder. Couldn't there be a way to expose that at the library level directly?

src/uniprotkb/components/protein-data-views/VariationView.tsx
```diff
@@ -7,7 +7,12 @@ import ProtvistaManager from 'protvista-manager';
 import ProtvistaSequence from 'protvista-sequence';
 import ProtvistaNavigation from 'protvista-navigation';
 import ProtvistaVariation from 'protvista-variation';
+import { filterConfig, colorConfig } from 'protvista-uniprot';
 import { transformData } from 'protvista-variation-adapter';
+import {
```
aurel-l, line 12: same thing here

aurel-l (approved):

aurel-l (approved):

PR #195, opened 2020-09-17
Querystring parser

aurel-l:
Added the concept of a querytring parser.

Implemented through TDD. Testing by parsing, then stringifying the resulting object, and comparing the resulting querystring with the original. Might have missed edge cases if they were not covered in the existing test set.


xwatkins (approved):
Looks good, nice list of test cases which I think covers everything.

PR #197, opened 2020-09-22
Uniref entry

aurel-l:
Basic UniRef search result and entry pages.

Not going further with this PR, waiting for further mockups to be finalised before.

Moved some (a lot of) files around to generalise a bit logic that used to be only for UniProtKB entries. Still plenty of things that could be moved more but we need to wait for mockups to see which other components will actually need to be generalised


src/shared/config/apiUrls.ts
```diff
@@ -76,6 +76,11 @@ const apiUrls = {
       '/publications'
     ),
   organismSuggester: '/uniprot/api/suggester?dict=organism&query=?',
+
+  // TODO: move that to UniRef-specific file?
```
xwatkins, line 80: Initially thought `apiUrls` would contain a json object of namespaces (uniprot, uniref etc...), each containing all the urls specific to that namespace. However over time it's turned into something else so maybe having one per namespace is not a bad idea.

src/uniref/components/entry/Entry.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+    disabled: !hasContent(transformedData[section.name]),
+  }));
+
+  const displayMenuData = [
+    {
+      name: 'Entry',
+      icon: <TremblIcon />,
+      itemContent: (
+        <InPageNav sections={sections} rootElement=".sidebar-layout__content" />
```
xwatkins, line 88: Might be a good time to revive https://github.com/ebi-uniprot/franklin-sites/pull/76 ?

src/uniref/components/entry/Entry.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const sections = UniRefEntryConfig.map((section) => ({
+    label: section.name,
+    id: section.name,
+
+    disabled: !hasContent(transformedData[section.name]),
+  }));
+
+  const displayMenuData = [
+    {
+      name: 'Entry',
```
xwatkins, line 85: As we don't have "tabs", only the in-page nav is needed, not the "Entry" header - UX chat this morning also points to this being moved to actual tabs in the future, so up to you if you want to keep it in or remove it now.

src/uniref/components/entry/MembersSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+        };
+      }
+    },
+    [data]
+  );
+
+  return (
+    <div id={EntrySection.Members}>
+      <Card title={EntrySection.Members}>
+        <protvista-datatable ref={setTableData} />
```
xwatkins, line 86: If we stick with `protvista-datatable` for this we should have a chat about improvements to the component and workarounds to the whole issue with passing `lit-html` and issue with things like rendering components like evidence tags.

src/shared/config/apiUrls.ts
aurel-l: It's a bit of a monster right now 😅

src/uniref/components/entry/Entry.tsx
aurel-l: Indeed. I was waiting for the last layout changes by Dan before going to have a look, so now I can look at it again

src/shared/config/apiUrls.ts
xwatkins: In my mind it was going to be this beautifully formatted simple thing! 😆

src/uniref/components/entry/MembersSection.tsx
aurel-l: I have some ideas and workaround about passing slots instead of using lit-html. But that'd definitely need a call.
Also, I'm not sure how to implement the facets, because right now, when looking at [https://www.uniprot.org/uniref/UniRef50_N9S125](https://www.uniprot.org/uniref/UniRef50_N9S125) for example, you see the table and facets look like a search result page. But there's not going to be a lot of facets, so probably we can implement facetting differently and make it work with protvista-datatable

src/uniref/components/entry/Entry.tsx
aurel-l: See #203

src/uniref/components/results/UniRefCard.tsx
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+import React, { FC } from 'react';
+import { Card } from 'franklin-sites';
+import { useHistory } from 'react-router-dom';
+
+// import { UniProtkbAPIModel } from '../../adapters/uniProtkbConverter';
```
its-mahdi, line 6: Should all of the commented code in this file stay for now or it can be added later on, if needed?

its-mahdi:
I also noticed in a few places you have added parenthesis around a single argument of a one-liner lambda e.g. `x => null` to `(x) => null`. I'm not sure if you have done this manually or automatically, but I remember the former is correct form in our style-guides -- unless something has changed and I missed it.

Nothing too important, but for the sake of clarification.

src/tools/align/components/results/AlignResult.tsx
```diff
@@ -109,30 +102,35 @@ const useParamsData = (
...
+    if (!match?.params?.subPage) {
       history.replace(
         history.createHref({
           ...history.location,
           pathname: `${history.location.pathname}/${TabLocation.Overview}`,
         })
       );
     }
-  }, [match.params.subPage, history]);
+  }, [match, history]);
```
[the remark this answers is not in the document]
aurel-l: Well, it would be called more times (every time the subPage changes). But the first thing is an if to not do anything if a subPage is present. Thing is, match might be undefined (if we use the align result component in the wrong place), so I need to check match, as it will cause an error to check match.param.subPage if match is undefined.

src/shared/utils/utils.ts
```diff
@@ -54,3 +54,21 @@ export const getBEMClassName = ({
   }
   return className;
 };
+
+export const hasContent = (obj: Record<string | number | symbol, unknown>) => {
```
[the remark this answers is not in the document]
aurel-l: Yeah, I'm not sure what it's supposed to do either, the shape of object it takes is quite restrictive and I'm not sure who created it and in what context and what issue it tried to solve in a generic way. That's a bit why I didn't touch it, just moved it.

src/uniprotkb/components/entry/Entry.tsx
```diff
@@ -99,26 +89,50 @@ const Entry: React.FC<EntryProps> = ({ addMessage, match }) => {
     [transformedData]
   );
 
+  const listOfIsoformAccessions = useMemo(
```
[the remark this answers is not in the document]
aurel-l: Rewrote a bit this logic in the latest commit. Didn't do reduce because it messes up with TypeScript types and I think it night be less complex to read with just for loops.

src/uniprotkb/components/entry/Entry.tsx
```diff
@@ -99,26 +89,50 @@ const Entry: React.FC<EntryProps> = ({ addMessage, match }) => {
...
-    const inactiveEntryData = (data as unknown) as UniProtkbInactiveEntryModel;
+    if (!match) {
+      return <ErrorHandler />;
+    }
 
     return (
       <ObsoleteEntryPage
-        accession={accession}
-        details={inactiveEntryData.inactiveReason}
+        accession={match.params.accession}
```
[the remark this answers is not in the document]
aurel-l: No, this is fine like this. TypeScript would have caught it if this was a problem. If no match I'm handling this just before with an `<ErrorHandler />`

src/uniprotkb/components/results/ResultsView.tsx
```diff
@@ -141,17 +148,34 @@ const ResultsView: React.FC<ResultsTableProps> = ({
...
         data={allResults}
-        dataRenderer={(dataItem: UniProtkbAPIModel) => (
-          <UniProtKBCard
-            data={dataItem}
-            selected={selectedEntries.includes(dataItem.primaryAccession)}
-            handleEntrySelection={handleEntrySelection}
-          />
-        )}
+        dataRenderer={(dataItem: UniProtkbAPIModel | UniRefAPIModel) => {
+          if (namespace === Namespace.uniref) {
```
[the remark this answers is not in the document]
aurel-l: changed

src/uniref/components/results/UniRefCard.tsx
aurel-l: I'd keep it for now, to clearly show that this file is a fork of the UniProtKB version of it. Whenever it's clearer what needs to go in there it might make it more obvious that there is similar code in UniProtKB because it's here in the comments and we'd need to generalise it. Otherwise we can remove it later when we flesh out this component.
In any case, we are going to have to go over this file in the near future.

aurel-l:
> I also noticed in a few places you have added parenthesis around a single argument of a one-liner lambda e.g. `x => null` to `(x) => null`. I'm not sure if you have done this manually or automatically, but I remember the former is correct form in our style-guides -- unless something has changed and I missed it.
> 
> Nothing too important, but for the sake of clarification.

This is not really my doing, this is added automatically by Prettier since its last major version

its-mahdi (approved):

PR #200, opened 2020-09-25
Msa refactor trm 24745

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24745
Updated existing tests but could use more tests.
Blast Results > Overview > Click Hit > View: Overview | Wrapped
Align Results > Overview > View: Overview | Wrapped
No

Created [jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-24853) to wire in functionality of align checkboxes.


src/tools/components/AlignOverview.tsx
```diff
@@ -0,0 +1,225 @@
...
+          features.filter(({ type }) => type === annotation)
+        );
+        processedFeatures = transformFeaturesPositions(processedFeatures);
+        node.data = processedFeatures;
+      }
+    },
+    [features, annotation]
+  );
+
+  const overviewHeight: string = (alignment && alignment.length > 10
```
aurel-l, line 137: Not sure why this needs to casted into a string, as later it is just parsed to a number again with the component that will use it

src/tools/align/components/results/AlignLabel.tsx
```diff
@@ -50,12 +66,19 @@ const AlignLabel: FC<Props> = ({ accession, info, loading, children }) => {
...
   return (
-    <span className={className}>
+    <span className={className} style={style}>
+      {withCheckbox && <input type="checkbox" />}
       {reviewImg}
       {before}
       {/* inject a link to the entry page */}
       <Link to={`/uniprotkb/${accession}`}>{accession}</Link>
-      {after.join(accession)}
+      {onIdClick ? (
```
aurel-l, line 75: Is it on purpose that this is only applied to the part after the extracted accession?
If no accession is detected we will just get a `before` and that's it.
Also, the whole label should react to the selection, ideally all except the link, but oh well, the link is taking you on a different page anyway, so it's fine, nobody will see the effect of the change of selected id

src/tools/align/components/results/AlignLabel.tsx
```diff
@@ -8,22 +8,38 @@ import { EntryType } from '../../../../uniprotkb/adapters/uniProtkbConverter';
 import { ParsedSequenceAndFeatures } from '../../utils/useSequenceInfo';
 
 import './styles/AlignLabel.scss';
+import { MSAInput } from '../../../components/AlignmentView';
+import { Sequence } from '../../../components/AlignWrapped';
 
 type Props = {
```
xwatkins, line 14: I know it's not part of the PR, but we usually have more descriptive name for Prop types, in this case `AlignLabelProps`

src/tools/components/AlignmentView.tsx
```diff
@@ -155,8 +189,8 @@ const MSAWrapper: React.FC<{
         </fieldset>
       </div>
       <div>
-        {activeView === View.overview ? (
-          <MSAView
+        {activeView === View.overview && tool === Tool.blast && (
```
aurel-l, line 192: If Overview and Wrapped components take the same props, maybe it would be better to do, before the return:
```js
let Component;
if (activeView === View.overview) {
  Component = tool === Tool.blast ? BlastOverview : AlignOverview;
} else {
  Component = tool === Tool.blast ? BlastWrapped : AlignWrapped;
}
```
and in the return just use `<Component ...`

xwatkins (changes requested):
Let's discuss next week...I spent quite a bit of time last sprint refactoring the Overview and Wrapped view so we could use the same code for BLAST and Align and this undoes that by again creating different components.

src/tools/components/AlignOverview.tsx
dlrice: I think this was an accident.

dlrice: Fixed.

src/tools/components/AlignmentView.tsx
dlrice: Yes, I've incorporated your suggestion

src/tools/align/components/results/AlignLabel.tsx
its-mahdi: As per our conversation we are going to leave this for now until UX review is done, so we are clear what needs to be done.

aurel-l:
I realised the annotations in blast results in wrapped view were not appearing for me. Does that happen to you too?

dlrice:
> I realised the annotations in blast results in wrapped view were not appearing for me. Does that happen to you too?

Yeah, fixed in the last commit. Also seeing this issue in align overview. Now looking into that.

aurel-l (approved):
OK for me, pending merge conflict resolution

PR #202, opened 2020-09-28
MSA PR in the PR

aurel-l:
PR inside the PR

handles:
 - connect align label checkboxes to blast and align action buttons
   - add checkboxes and connect to action buttons in phylotree ad pim components
   - in wrapped view, toggle all checkboxes corresponding to the same sequence
 - floor the coordinate number that was appearing as a decimal when first switching to overview
 - make whole align label clickable as way to select one of the alignments
 - optimise use of grid and cell positionning
   - push the cells to the bottom to keep alignment without injecting dynamic style depending on track information
   - merge "rows" into one big section and handle grids as grids (stuff all the cells in one grid, and assign areas to those cells)
   - define size of side columns depending on content and not based on fixed sizing anymore -> better use of space

Notes and rough edges:
 - ~in wrapped view, the right column gets a different size based on the size of number displayed~
   - -> ~fixable by defining a huge grid with repeated grid-area patterns~ -> done (and it even simplifies the code)
 - in overview, the zoom of the nightingale components jumps every time the checkbox is selected

### DIFF
--- a/src/tools/components/Overview.tsx
+++ b/src/tools/components/Overview.tsx
@@ -53,7 +53,8 @@ export type BlastOverviewProps = {
 };
 
 // NOTE: hardcoded for now, might need to change that in the future if need be
-const heightStyle = { height: '20px' };
+const sequenceHeight = 20;
+const heightStyle = { height: `${sequenceHeight}px` };
 
 const AlignOverview: FC<BlastOverviewProps> = ({
   alignment,
@@ -204,7 +205,7 @@ const AlignOverview: FC<BlastOverviewProps> = ({
           <protvista-msa
             ref={setMSAAttributes}
             length={alignmentLength}
-            height={alignment.length * 20}
+            height={alignment.length * sequenceHeight}
             colorscheme={highlightProperty}
             {...conservationOptions}
           />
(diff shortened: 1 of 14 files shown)


src/tools/components/Wrapped.tsx
```diff
@@ -113,68 +113,52 @@ const MSAWrappedRow: FC<MSAWrappedRowProps> = ({
...
+          >
+            {s.name || ''}
+          </AlignLabel>
+        ))}
+      </div>
+      <div className="track">
+        <protvista-msa
+          ref={setMSAAttributes}
+          length={rowLength}
+          height={sequences.length * 20}
```
dlrice, line 143: Could make use of `heightStyle` here

src/tools/components/Overview.tsx
```diff
@@ -150,88 +153,77 @@ const AlignOverview: FC<BlastOverviewProps> = ({
...
+      <div className="track">
+        <protvista-manager
+          ref={managerRef}
+          attributes="displaystart displayend"
+        >
+          <protvista-navigation length={alignmentLength} />
+          <protvista-msa
+            ref={setMSAAttributes}
+            length={alignmentLength}
+            height={alignment.length * 20}
```
dlrice, line 207: Could make use of `heightStyle` here

dlrice (commented):
Is it possible to prevent the checkbox selection on an align sequence from changing which data populates the annotation track?

aurel-l:
> Is it possible to prevent the checkbox selection on an align sequence from changing which data populates the annotation track?

Handled this and your inline comments in the latest commit

dlrice (approved):

PR #204, opened 2020-10-01
2020 11 uniref  better msa highlights

aurel-l:
   - [TRM-24862](https://www.ebi.ac.uk/panda/jira/browse/TRM-24862)
   - Change highlight option in any MSA view

Depends on [https://github.com/ebi-uniprot/franklin-sites/pull/106](https://github.com/ebi-uniprot/franklin-sites/pull/106) (hence failing tests)

 - Removed the 2 highlight options that were only for nucleotides
 - Categorised other options and use a TreeSelect to make the selection cleaner (category names pending review from a curator though)


src/tools/components/MSAWrapper.tsx
```diff
@@ -83,33 +80,24 @@ const MSAWrapper: React.FC<{
...
-            </ul>
-          </div>
-        </DropdownButton>
+        <TreeSelect
+          data={colorSchemeTree}
+          onSelect={handleHighlightSelect}
+          autocomplete
+          autocompleteFilter
+          autocompletePlaceholder="Enter a color scheme"
+          label="Highlight properties"
```
dlrice, line 97: How it is now is perfectly fine but I am wondering if it would be useful to change the label to the selected colour scheme?

dlrice (approved):

PR #206, opened 2020-10-06
Handling 400 errors

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24855

Run locally and go to: http://localhost:8080/uniprotkb/A9H9K

You should see at least one pop-up, indicating an error.


aurel-l:
Changed a bit the test to test the correct call of the redux dispatch with the correct object describing a pop-up message

aurel-l (approved):

PR #207, opened 2020-10-08
Fix issue with variant highlight

xwatkins:
Noticed an issue where clicking on a variant (dot) in the viewer wouldn't highlight it in the table. This is because both components need to use "transformedData" which has `protvistaFeatureId` set.
Fixed and updated tests.

### DIFF
--- a/__mocks__/protvista-variation-adapter.js
+++ b/__mocks__/protvista-variation-adapter.js
@@ -1,5 +1,13 @@
 module.exports = {
   transformData: () => ({
     sequence: 'ABCD',
+    variants: [
+      {
+        accession: 'accessionData',
+        variant: 'A',
+        start: 10,
+        end: 10,
+      },
+    ],
   }),
 };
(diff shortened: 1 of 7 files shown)


aurel-l:
I get a bunch of "undefined>undefined" in the "Change" column of the variants though

aurel-l:
I get quite a lot of ">undefined" in the Change column. See for example variant at position 16 in P11166

xwatkins:
> I get quite a lot of ">undefined" in the Change column. See for example variant at position 16 in P11166

Did you run `yarn`? This could be because `ProtVista-variation-adapter` wasn’t updated

aurel-l:
> > I get quite a lot of ">undefined" in the Change column. See for example variant at position 16 in P11166
> 
> Did you run `yarn`? This could be because `ProtVista-variation-adapter` wasn’t updated

I did. First I hadn't and I got just plenty of "undefined>undefined", but now that I have updated not all of them have undefined in there, but quite a lot do

aurel-l (approved):

PR #208, opened 2020-10-08
2020 11 uniref query builder endpoints

xwatkins:


aurel-l:
Just to keep track of the 2 issues I'm having as of right now I'll mention them here:
 - Prefilling form with an `autoCompleteQueryTerm`:
   - e.g. `taxonomy_id:9606` => need to link it back to `taxonomy_name` field
 - Preflling form with a cross reference (remove prefix and use it to find correct crossref term)

aurel-l:
Finished covering all edge-cases I could think of.

Last things to check:
 - autofill of terms with "autoCompleteQueryTerm" -> try taxonomy for example, when you go back to the field it autofills, but if you resubmit from there the search will be different. Not sure how to handle that.
 - big warning in the AdvancedSearch component test (see `// FIXME:` comment).

And we probably need more tests, but please someone review/try the new code before.

aurel-l:
Also, should we rename the URL to match what is actually being displayed in the link? advanced-search -> query-builder.
If so, maybe rename the components and variable names to match that too

src/query-builder/utils/fieldInitializer.ts
```diff
@@ -40,13 +40,10 @@ const initializer = (
...
     const { regex } = field;
-    let re: RegExp;
+
     const cleanRegex = regex.replace('(?i)', '');
-    if (regex.length !== cleanRegex.length) {
-      re = new RegExp(cleanRegex, 'i');
-    } else {
-      re = new RegExp(regex);
-    }
+    const re = new RegExp(cleanRegex, regex === cleanRegex ? undefined : 'i');
```
xwatkins, line 45: 👍 I knew at the time there was a cleaner way of doing this, but had brain fog!

PR #209, opened 2020-10-10
2020 12 alignment entry  trm 24560

aurel-l:
- [x] What is the link to the Jira that this PR is for? [TRM-24560](https://www.ebi.ac.uk/panda/jira/browse/TRM-24560) (I think, Jira is down at the time I'm opening this PR 😅)
- [x] Is it possible to visually evaluate this PR in the browser? If so, what are the steps to do this (eg specific URLs or sections)? BLAST result, click on differents scores

Depends on [https://github.com/ebi-uniprot/franklin-sites/pull/109](https://github.com/ebi-uniprot/franklin-sites/pull/109)

Now we are colouring the overview of the match in the BLAST result page, and the scoring chips, differently according to which score is visualised. Click on chip to change visualised score.

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,6 +178,7 @@
     "lodash": "4.17.20",
     "mini-css-extract-plugin": "0.11.3",
     "node-sass": "4.14.1",
+    "node-sass-json-importer": "4.3.0",
     "npm-run-all": "4.1.5",
     "prettier": "2.1.2",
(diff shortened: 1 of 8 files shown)


dlrice (approved):
Looks good to me but maybe a UI/X person would like a peek too (if they haven't already).

PR #211, opened 2020-10-15
2020 12 alignment entry  computationaly mapped

xwatkins:
Adding a component to display computationally mapped isoforms (by using the gene centric service), and corresponding tests.


src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
```diff
@@ -0,0 +1,168 @@
...
+      `${LocationToPath[Location.UniProtKBResults]}?query=(${queryString})`
+    );
+  }, [history, filteredData]);
+
+  if (error) {
+    if (status === 404) {
+      // Fail silently, this just means there's no data
+      return null;
+    }
+    dispatch(
```
aurel-l, line 119: Not sure if this should pop-up a message here. This is just one section of an entry, and the user might not even need it or see it whenever they see the pop-up. I would have that as a localised error message inlined in the card

xwatkins: Good point, I'd forgotten this was the approach we decided to take. Will update accordingly

aurel-l (approved):

PR #213, opened 2020-10-19
2020 12 alignment gaps

dlrice:
[TRM-24890](https://www.ebi.ac.uk/panda/jira/browse/TRM-24890)

The inclusion of gap insertions in the overview and wrapped views works by partitioning the sequence string into substrings which consist of either `-`s (insertion) or anything that is not `-` (sequence). Each substring is this then added as a feature fragment where the insertions positions are counted and are 1-based (we refer to this as string coordinates). Eg `AB--C` is from 1 to 5 (while in the protein alignment this would be 1 to 3). When rendering, the protvista tracks use the 1-based string coordinates to line up with the MSA. For example, imagine a sequence has 30 insertions then the following would be the start and end coordinates:

```
MSA: 1-50
  coordinate: 1-20 (insertions aren't counted)
protvista-track: 1-50
```

There is currently an issue that needs resolving where the start of the track is offset by a small ~1/2 amino acid width. To reproduce this run the following alignment:

```
>sp|O00311|CDC7_HUMAN Cell division cycle 7-related protein kinase OS=Homo sapiens OX=9606 PE=1 SV=1
MEASLGIQMDEPMAFSPQRDRFQAEGSLKKNEQNFKLAGVKKDIEKLYEAVPQLSNVFKI
EDKIGEGTFSSVYLATAQLQVGPEEKIALKHLIPTSHPIRIAAELQCLTVAGGQDNVMGV
KYCFRKNDHVVIAMPYLEHESFLDILNSLSFQEVREYMLNLFKALKRIHQFGIVHRDVKP
SNFLYNRRLKKYALVDFGLAQGTHDTKIELLKFVQSEAQQERCSQNKSHIITGNKIPLSG
PVPKELDQQSTTKASVKRPYTNAQIQIKQGKDGKEGSVGLSVQRSVFGERNFNIHSSISH
ESPAVKLMKQSKTVDVLSRKLATKKKAISTKVMNSAVMRKTASSCPASLTCDCYATDKVC
SICLSRRQQVAPRAGTPGFRAPEVLTKCPNQTTAIDMWSAGVIFLSLLSGRYPFYKASDD
LTALAQIMTIRGSRETIQAAKTFGKSILCSKEVPAQDLRKLCERLRGMDSSTPKLTSDIQ
GHASHQPAISEKTDHKASCLVQTPPGQYSGNSFKKGDSNSCEHCFDEYNTNLEGWNEVPD
EAYDLLDKLLDLNPASRITAEEALLHPFFKDMSL
>sp|P05067|A4_HUMAN Amyloid-beta precursor protein OS=Homo sapiens OX=9606 PE=1 SV=3
MLPGLALLLLAAWTARALEVPTDGNAGLLAEPQIAMFCGRLNMHMNVQNGKWDSDPSGTK
TCIDTKEGILQYCQEVYPELQITNVVEANQPVTIQNWCKRGRKQCKTHPHFVIPYRCLVG
EFVSDALLVPDKCKFLHQERMDVCETHLHWHTVAKETCSEKSTNLHDYGMLLPCGIDKFR
GVEFVCCPLAEESDNVDSADAEEDDSDVWWGGADTDYADGSEDKVVEVAEEEEVAEVEEE
EADDDEDDEDGDEVEEEAEEPYEEATERTTSIATTTTTTTESVEEVVREVCSEQAETGPC
RAMISRWYFDVTEGKCAPFFYGGCGGNRNNFDTEEYCMAVCGSAMSQSLLKTTQEPLARD
PVKLPTTAASTPDAVDKYLETPGDENEHAHFQKAKERLEAKHRERMSQVMREWEEAERQA
KNLPKADKKAVIQHFQEKVESLEQEAANERQQLVETHMARVEAMLNDRRRLALENYITAL
QAVPPRPRHVFNMLKKYVRAEQKDRQHTLKHFEHVRMVDPKKAAQIRSQVMTHLRVIYER
MNQSLSLLYNVPAVAEEIQDEVDELLQKEQNYSDDVLANMISEPRISYGNDALMPSLTET
KTTVELLPVNGEFSLDDLQPWHSFGADSVPANTENEVEPVDARPAADRGLTTRPGSGLTN
IKTEEISEVKMDAEFRHDSGYEVHHQKLVFFAEDVGSNKGAIIGLMVGGVVIATVIVITL
VMLKKKQYTSIHHGVVEVDAAVTPEERHLSKMQQNGYENPTYKFFEQMQN
```

When looking at the `chain` feature `O00311` you will see this in the first row (I've played with the window width and it is persistent).


dlrice:
The `content-visibility` property is nice but my only reservation is it compatibility as it's only currently functioning in chrome. At least we will have graceful failure (at the cost of performance) in this case as everything will just render at once. I have tried BLASTing Titin (as you've suggested before) and it does cause the page to freeze so we will probably want to investigate some ways to dealing with large numbers of rows.

dlrice:
Aside from the occasional and small mismatches seen in the msa and nightingale locations, this is ready for review.

src/tools/utils/sequences.ts
```diff
@@ -133,3 +125,51 @@ export const removeFeaturesWithUnknownModifier = (features?: FeatureData) =>
...
   );
+
+export const createGappedFeature = (
+  feature: ProcessedFeature,
+  sequence: string
+) => {
+  /*
+  input: feature and sequence are both 1-based
+  */
+  const BLOCK = /(?<insertion>[-]+)|(?<protein>[^-]+)/g;
```
aurel-l, line 136: I remembered we had issues before with named capture groups and Firefox support. I would say that now since my latest changes in the Babel config this is transpiled/polyfilled correctly and that shouldn't be a problem. But I would recommend double-checking on Firefox 77 or less (FF78 is the first supporting this)

aurel-l:
About the off-by-0.5 issue when a feature starts on the first aminoacid, I would say it would be good to reproduce out of the uniprot website context and if it can be reproduced open an issue on the nightingale repo. I think this might be out of the scope of the related Jira and would need its own Jira later on.

aurel-l (approved):
Apart from the offset issue that shouldn't prevent this PR to be closed, and after the check on the Firefox support, I'd say you can go ahead and merge this

src/tools/utils/sequences.ts
dlrice: Checked on FF72 and all good.

PR #215, opened 2020-10-22
2020 12 a few alignment fixes

dlrice:
A few bug fixes for align:

1. When a user provides a sequence that isn't in UniProtKB we now set `from`, `to`, `length`. Without these we were getting some negative value errors in protvista's SVGs.
2. Don't set `initialDisplayEnd` if `displayEnd` is already defined as this was causing the navigation to alternate between `initialDisplayEnd` and `displayEnd` during scrolling.

### DIFF
--- a/src/tools/align/components/results/AlignResultOverview.tsx
+++ b/src/tools/align/components/results/AlignResultOverview.tsx
@@ -70,11 +70,11 @@ const enrichParsed = (
       features: features ?? [],
       ...sequence,
       // not sure yet if that is needed or not
-      ...getFromToLength(sequence.sequence),
+      // ...getFromToLength(sequence.sequence),
       // or if those values are enough
-      // from: 1,
-      // to: sequence.sequence?.length || 0,
-      // length: sequence.sequence?.length || 0,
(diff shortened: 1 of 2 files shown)


src/tools/align/components/results/AlignResultOverview.tsx
```diff
@@ -47,23 +47,35 @@ const enrichParsed = (
...
+    let features;
+    if (sequence.name && nameToSequenceInfo[sequence.name]) {
+      info = nameToSequenceInfo[sequence.name];
+      features = removeFeaturesWithUnknownModifier(info?.features);
     }
+    sequences[index] = {
+      ...info,
+      features: features ?? [],
+      ...sequence,
+      // not sure yet if that is needed or not
```
aurel-l, line 72: Has this been clarified later on then? I from reading the PR comment I understand it has (yes, needed). If that is correct maybe remove or rewrite those comments as to not see them later and lose time on this.

aurel-l (approved):

aurel-l (approved):

PR #217, opened 2020-10-23
Add caution comment

xwatkins:
During Preethi’s presentation, I noticed we were missing a comment type in the new website: Caution.
See https://www.uniprot.org/uniprot/P14210#miscellaneous for an example. As it’s highlighted in the old website, I take it it has to be highlighted for the user too so thought of using a message component of status “warning”.


aurel-l (approved):
Accepted, but realised it needed some changes in franklin-sites to avoid message styles bleeding into the content of that message

PR #218, opened 2020-11-02
Use tabs for entry page

xwatkins:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24989

Updated snapshots

Navigate to entry view. Tabs are now horizontal, sticky tabs and the side panel only contains the InPageNav or a grey rectangle. I'm also sharing code to display the data for the protein cards in results and the one at the top of the entry page.


aurel-l (approved):
Added 2 small commits (happy to revert or modify if disagree).
Approving otherwise.

xwatkins:
Thanks! 
`  // if URL doesn't finish with "entry" redirect to /entry by default`
Was wondering about that, and the fact we have direct links to subsections of the main entry page e.g. https://ebi-uniprot.github.io/uniprot-website/uniprotkb/Q54DK3#Function and whether changing the url in that case was a good idea or not... (I still don't have a clear preference)

aurel-l:
> ` // if URL doesn't finish with "entry" redirect to /entry by default`
> Was wondering about that, and the fact we have direct links to subsections of the main entry page e.g. https://ebi-uniprot.github.io/uniprot-website/uniprotkb/Q54DK3#Function and whether changing the url in that case was a good idea or not... (I still don't have a clear preference)

Well, the way it's handles now it still works, links to `/uniprotkb/P11166#ptm-processing`, for example, are working and redirecting automatically to `/uniprotkb/P11166/entry#ptm-processing` without any issue

PR #220, opened 2020-11-10
2020 13 uniref results

xwatkins:
- Made `ResultsContainer` and `ResultsView` components more generic
- Created `ColumnConfiguration` for UniRef
- Re-enabled tests for `ResultsContainer`
- Created more issues under https://www.ebi.ac.uk/panda/jira/browse/TRM-24743 (Action buttons, download functionality)


aurel-l:
One a first quick look, I think the logic for the "Types" column is flawed somewhere, because all of the entries appear to only have UniProtKB unreviewed members right now.

src/uniref/adapters/uniRefConverter.ts
```diff
@@ -70,7 +64,7 @@ export type UniRefLiteAPIModel = {
   organismCount: number;
   representativeId: string;
   seedId: string;
-  memberIdTypes: MemberIdType;
+  memberIdTypes?: string[];
```
aurel-l, line 67: Not sure why the change to remove the specificity and make it a string. Aren't we supposed to only have 3 possible options here?

src/uniref/config/__tests__/__snapshots__/UniRefColumnConfiguration.spec.tsx.snap
```diff
@@ -0,0 +1,67 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`UniRefColumnConfiguration component should render all columns: Cluster ID 1`] = `
+<DocumentFragment>
+  <a
+    href="/uniref/undefined"
```
aurel-l, line 6: Something's off here too 🤔

xwatkins: Was passing the wrong object...

src/uniref/adapters/uniRefConverter.ts
xwatkins: Not sure either. Updated to `EntryType`

xwatkins:
> One a first quick look, I think the logic for the "Types" column is flawed somewhere, because all of the entries appear to only have UniProtKB unreviewed members right now.

I'm able to see some UniProtKB members by searching for P05067 in UniRef. Have merged `2020_13` in the branch and added UniParc icon.

src/shared/components/entry/EntryTypeIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+        title={title || 'This marks an unreviewed entry'}
+      >
+        <TremblIcon />
+      </span>
+    );
+  }
+  if (entryTypeToCheck === EntryType.UNIPARC) {
+    return (
+      <span
+        className="entry-title__status icon--unreviewed"
```
aurel-l, line 43: `icon--unreviewed` is wrong here. It either requires a `icon--uniparc` class added in Franklin, or colouring to be applied from the uniprot-website codebase

src/shared/components/results/ResultsView.tsx
```diff
@@ -1,46 +1,88 @@
...
 };
 
-const ResultsView: React.FC<ResultsTableProps> = ({
+const convertRow = (
+  row: UniProtkbAPIModel | UniRefLiteAPIModel,
+  namespace: Namespace
+) => {
+  switch (namespace) {
+    case Namespace.uniprotkb:
+      return uniProtKbConverter(row as UniProtkbAPIModel) as UniProtkbUIModel;
```
aurel-l, line 59: The 2nd `as` is not needed here

src/shared/components/results/ResultsView.tsx
```diff
@@ -1,46 +1,88 @@
...
+        primaryAccession;
+    case Namespace.uniref:
+      return ({ id }: { id: string }) => id;
+    default:
+      return null;
+  }
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const ColumnConfigurations: Partial<Record<Namespace, Map<any, any>>> = {
```
aurel-l, line 80: Maybe not needed for now, but maybe add a `TODO:` comment to remove the `any, any` types eventually.
We need a Column type covering UniProtKBColumns, UniRefColumns, etc for the first any, and a Column renderer type with `label: string` and a render definition for the second any.

src/shared/components/entry/EntryTypeIcon.tsx
xwatkins: I'll add an `icon--uniparc`, I think it's more consistent.

src/shared/components/results/ResultsView.tsx
xwatkins: Removed

src/shared/components/results/ResultsView.tsx
xwatkins: Added comment

src/shared/components/entry/EntryTypeIcon.tsx
xwatkins: Added to franklin, will be available next time we publish

aurel-l (approved):

PR #223, opened 2020-11-17
2020 13 uniprot entry changes

its-mahdi:
https://www.ebi.ac.uk/panda/jira/browse/TRM-24990

1) Go to: http://localhost:8080/uniprotkb/P05067/entry#sequence
You should see "Sequence & Isoforms" on the sidebar, as well as the section title in the page.

2) Go to: http://localhost:8080/uniprotkb/P28028/entry#sequence
You should see "Sequence & Isoform" on the sidebar, as well as the section title in the page.

3) Go to: http://localhost:8080/uniprotkb/P02879/entry#sequence
You should see "Sequence" on the sidebar, as well as the section title in the page.
You should also see "Phenotypes" instead of "Disease & Drugs"

4) Go to: http://localhost:8080/uniprotkb/P35613/entry#function
You should see "Biotechnology" under the Function section.

Notes:

1) There is a known bug on the sidebar. If you go to http://localhost:8080/uniprotkb/P35613/entry and click on "Sequence & Isoforms" you will see an indicative border next to the item you just clicked. However, if you go to http://localhost:8080/uniprotkb/P05067/entry and repeat the same process, the border will appear next to "Family & Domains".

Please let me know if this is the case for you too and if you have any suggestions.

2) Probably the most important part of this PR is the way that the switching between "Sequence", "Sequence & Isoform" and "Sequence & Isoforms" is happening. This is because the original value is part of an Enum. I tried to find a way to avoid concatenating strings as it wasn't ideal -- because this isn't just concatenating a "s" to another string, so had to extend the Enum and create a utility function to handle this.


src/uniprotkb/adapters/uniProtkbConverter.ts
```diff
@@ -123,11 +126,14 @@ const uniProtKbConverter = (data: UniProtkbAPIModel): UniProtkbUIModel => {
     [EntrySection.NamesAndTaxonomy]: convertNamesAndTaxonomy(dataCopy),
     [EntrySection.SubCellularLocation]: convertSubcellularLocation(dataCopy),
     [EntrySection.DiseaseAndDrugs]: convertDiseaseAndDrugs(dataCopy),
+    [EntrySection.Phenotypes]: convertDiseaseAndDrugs(dataCopy),
```
aurel-l, line 129: The changes to this file are slightly disconcerting to me. It generates duplicated logic, and make some logic run multiple times with the same output (3 times `convertDiseaseAndDrugs`).

src/uniprotkb/types/entrySection.ts
```diff
@@ -5,8 +5,11 @@ enum EntrySection {
   Interaction = 'Interaction',
   NamesAndTaxonomy = 'Names & Taxonomy',
   DiseaseAndDrugs = 'Disease & Drugs',
+  Phenotypes = 'Phenotypes',
   ProteinProcessing = 'PTM/Processing',
-  Sequence = 'Sequence & Isoforms',
+  Sequence = 'Sequence',
```
aurel-l, line 10: Regarding the previous comment, I would change that enum to be functions returning a title

src/uniprotkb/components/entry/DiseaseAndDrugsSection.tsx
```diff
@@ -22,13 +22,18 @@ const DiseaseAndDrugsSection: FC<{
...
     return null;
   }
+
+  const taxonomyBasedSection =
+    taxId === 9606 ? EntrySection.DiseaseAndDrugs : EntrySection.Phenotypes;
+
   return (
-    <div id={EntrySectionIDs[EntrySection.DiseaseAndDrugs]} data-entry-section>
-      <Card title={EntrySection.DiseaseAndDrugs}>
+    <div id={EntrySectionIDs[taxonomyBasedSection]} data-entry-section>
```
aurel-l, line 35: Following previous comment.
So that here you would call the function with the tax id here for example and decide what title to generate

src/uniprotkb/components/entry/Entry.tsx
```diff
@@ -91,19 +93,31 @@ const Entry: FC = () => {
...
-        id: section.id,
-        disabled:
-          section.name === EntrySection.ExternalLinks
-            ? !hasExternalLinks(transformedData)
-            : !hasContent(transformedData[section.name]),
-      })),
-    [transformedData]
-  );
+      UniProtKBEntryConfig.map((section) => {
+        const taxId =
```
aurel-l, line 100: I would just move the definition of taxId and numberOfIsoforms outside of the map (but inside the useMemo), as this will be the same value at every loop of the map

src/uniprotkb/components/results/__tests__/__snapshots__/UniProtKBCard.spec.tsx.snap
```diff
@@ -108,7 +108,7 @@ exports[`UniProtKBCard component should render 1`] = `
         class="card__link"
       >
         <a
-          href="/uniprotkb/P21802#sequence"
+          href="/uniprotkb/P21802#Sequence"
```
aurel-l, line 111: I think this is incorrect, it should be the lowercase id used and it looks like it's using the name

src/uniprotkb/adapters/proteinHighlights.ts
```diff
@@ -33,31 +33,31 @@ const highlightToEntrySection: {
   };
 } = {
   [highlightSection.domains]: {
-    link: `#${EntrySectionIDs[EntrySection.Function]}`,
+    link: `#${getEntrySectionNameAndId(EntrySection.Function).name}`,
```
aurel-l, line 36: I guess related to the comment I left just a few seconds before. You might want to use id to use in a hash link

its-mahdi: Okay, makes sense. I used the name because that's what we used previously, but it won't work anymore!

aurel-l (approved):

PR #225, opened 2020-11-25
2020 13 uniref entry members

aurel-l:
- [x] What is the link to the Jira that this PR is for? All of the currently "in progress" subtasks of [TRM-25002](https://www.ebi.ac.uk/panda/jira/browse/TRM-25002)
- [x] Is it possible to visually evaluate this PR in the browser? If so, what are the steps to do this (eg specific URLs or sections)? Browse to a UniRef entry, check the "members" section. Also, try the action buttons.
- [x] Are there any linting errors that required you to add exceptions? `consistent-return` (x2), `default-case`

Handle the tasks related to the UniRef entry pages, specifically the "Members" section, and the action buttons.

Notes:
 - RelatedClusters component added, but commented out for now as it requires changes on the backend to handle this specific query.
 - Added a "direct" parameter in a result page query string to handle the case where we expect to get only one result (e.g. accession search, or previous point about related cluster -> directly jump to the related lower identity cluster) and want to redirect for a search result page to an entry page.
 - Added a usePrefetch hook to be used on pages with infinite scroll to preload page `n+1` just after loading page `n`.
 - Added types for UniParc as it was required for this PR to have basic types to be able to load sequence data from UniParc endpoint -> will need to be double-checked when working on UniParc pages.


src/shared/hooks/useDataApi.ts
```diff
@@ -15,7 +15,7 @@ export type UseDataAPIState<T> = {
   data?: T;
   status?: AxiosResponse['status'];
   statusText?: AxiosResponse['statusText'];
-  headers?: AxiosResponse['headers'];
+  headers?: Record<string, string>;
```
xwatkins, line 18: Did you have to change this because of custom headers? Why is it an issue now but wasn't before?

src/shared/hooks/usePrefetch.ts
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useMemo } from 'react';
```
xwatkins, line 1: Nice!

src/shared/hooks/__tests__/usePrefetch.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    const { rerender } = renderHook((props) => usePrefetch(props.url), {
+      initialProps: { url },
+    });
+
+    const link = queryByTestId(document, 'prefetch');
+    expect(link).not.toBeNull();
+    expect(link.getAttribute('rel')).toBe('prefetch');
+    expect(link.getAttribute('href')).toBe(url);
+
+    rerender({ url: url2 });
```
xwatkins, line 39: Just checking- there can be multiple prefetch `link`s at the same time right? not just one `link` with a changeable url?

src/uniprotkb/utils/resultsUtils.ts
```diff
@@ -15,16 +15,19 @@ const facetsAsArray = (facetString: string): SelectedFacet[] => {
...
   });
 };
+
 export type URLResultParams = {
   query: string;
   selectedFacets: SelectedFacet[];
   sortColumn: SortableColumn;
   sortDirection: SortDirection;
   activeFacet?: string;
+  direct?: boolean;
```
xwatkins, line 25: We need to think about a way to redirect to an entry directly if there's only 1 result - IDK if the current website uses `direct` to redirect to it or not?

src/shared/components/results/ResultsView.tsx
```diff
@@ -113,6 +126,21 @@ const ResultsView: FC<ResultsTableProps> = ({
     prevViewMode.current = viewMode;
   });
 
+  // redirect to entry directly when only 1 result and query marked as "direct"
```
xwatkins, line 129: Aaaaannnnd that makes my other comment irrelevant 🤣

xwatkins (commented):
There was _a lot_ to unpack there...! It would be great in the future if there could be separate smaller PRs for the main changes/new features. Maybe by branching off your work in progress branch and submitting a PR back to it.

It would be good if we could have a team chat to make everyone aware of the prefetch and where/how to use it.

Most of my comments are more actual comments rather than change requests, so approved in principle - but we need to have that chat about the related clusters issue.

src/shared/hooks/usePrefetch.ts
xwatkins: (that's a comment on the functionality, not the `import` BTW 😄

src/shared/components/action-buttons/ToolsButton.tsx
```diff
@@ -49,19 +50,26 @@ const ToolsButton: FC<ToolsButtonProps> = ({
...
+        })
       );
 
       if (entries !== entriesRef.current) {
         // it means that by the time we get here, the selection has changed
         return;
       }
       history.push(LocationToPath[location], {
-        parameters: { sequence: sequences.join('\n\n') },
+        parameters: { sequence: sequences.filter(Boolean).join('\n\n') },
```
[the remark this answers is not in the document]
aurel-l: It's cool but it doesn't help with TypeScript to ensure something is not undefined. I still have issues with that, you need to use a function that returns the filtered type with the `is` keyword or something like this, but I never find it straightforward to use.
In this case TypeScript is happy joining together undefined values so it's fine, it's just that I have to know as the developer that this won't be the case.

src/shared/components/entry/EntryTypeIcon.tsx
```diff
@@ -10,18 +10,25 @@ import './styles/entry-type-icon.scss';
...
+  entryId?: string;
   title?: string;
-}> = ({ entryType, title }) => {
+}> = ({ entryType, entryAccession, entryId, title }) => {
   let entryTypeToCheck = entryType;
   if (typeof entryType === 'string') {
     entryTypeToCheck = getEntryTypeFromString(entryType);
   }
+  if (!entryTypeToCheck && entryAccession && entryId) {
+    entryTypeToCheck = entryId.startsWith(`${entryAccession}_`)
```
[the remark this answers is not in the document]
aurel-l: Leo actually told me to do the assumption I'm doing here. If the accession is the same than the beginning of the ID, it's unreviewed.
If this doesn't feel right to do, that is one of the things we need to ask backend to provide for us on the UniRef members data

src/shared/hooks/useDataApi.ts
aurel-l: `AxiosResponse['headers']` is typed as `any` in the library definition package, which is why it didn't block us before, we could quite literally do `any`thing with this value.
I just made it more restrictive here after seeing that I had made an error and TS hadn't caught it.

src/shared/hooks/__tests__/usePrefetch.spec.ts
aurel-l: Yes, you can have multiple prefetch links.
Here in this test I can be sure there is only one because I'm using the hook once.
And I made sure to have the hook reuse the same link element it's using all along its lifecycle and to cleanup after itself on unmount for performance reasons

src/uniprotkb/utils/resultsUtils.ts
aurel-l: For a normal search on the current website I don't know, because there's no frontend redirect, it's probably handled by the server directly.
We need to be cautious here and we need to have a UX discussion of this behaviour if we are to apply it by default, to reasons on the top of my head:
 - What do you do if you only have 1 result but it's just because it's using a really restrictive query or facets?
   - We need to redirect only if the query was an accession? or an ID? or part of an ID (uniparc search for "123" => autoredirect to UPI00...000123)?
   - We need to be clearer when a search result page is a search result page and when we have an entry page (thinking UniRef here) because if I get lost sometimes and don't realise on which type of page I am, I can't expect a casual user to understand that they've been redirected "for their good".

Also, not a fan of having different behaviour (redirect or not) on 2 similar buttons really next to each other ("expand" and "list" clusters). But I'm just matching current behaviour in this PR for now.

src/shared/components/results/ResultsView.tsx
aurel-l: 😅 just matching current behaviour here, but we need to discuss that with UX I think, for the reasons raised in my reply to your other comment

aurel-l:
> There was _a lot_ to unpack there...! It would be great in the future if there could be separate smaller PRs for the main changes/new features. Maybe by branching off your work in progress branch and submitting a PR back to it.

Yeah, I could have separated the action buttons from the members table content renderer. There was a also a lot of sprint branch catching up that got caught up in the middle so I had to do stuff unrelated to the PR scope to keep up (e.g. changes to the UniRef sections to match the new UniProtKB logic)

> It would be good if we could have a team chat to make everyone aware of the prefetch and where/how to use it.

Yep, that's one thing I wanted just one person to have a look at and be on board before just showing it to everyone. Also one of the things I wanted to discuss in the follow up chat after whoever reviewed had a look at the code and the related tests.

> Most of my comments are more actual comments rather than change requests, so approved in principle - but we need to have that chat about the related clusters issue.

src/app/config/urls.ts
```diff
@@ -41,6 +42,7 @@ export const LocationToPath = {
   [Location.UniRefResults]: '/uniref',
   [Location.UniParcEntry]: '/uniparc/:accession',
   [Location.UniParcResults]: '/uniparc',
+  [Location.TaxonomyEntry]: '/taxonomy/:accession',
```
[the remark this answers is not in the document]
aurel-l: I think I'll create a jira for that to tackle all the remaining links that are generated "manually" to replace them with their corresponding `generatePath` + `Location.<>`. I don't want to make this PR even bigger 😅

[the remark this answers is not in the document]
xwatkins: Good plan

PR #226, opened 2020-11-26
2020 14 uniref customise columns

dlrice:
[TRM-25036](https://www.ebi.ac.uk/panda/jira/browse/TRM-25036)
The column select and customise table code test coverage is sufficient but I had to skip the download tests as download is now broken due its reliance on column select so the global coverage is below our ratcheted threshold. Download is a separate task [TRM-25118](https://www.ebi.ac.uk/panda/jira/browse/TRM-25118) and didn't want to do both at once to keep PRs as short as possible. So that other devs don't also have to contend with the lower global from this branch, I won't immediately merge it in to the sprint branch but instead build on top of it for the download task (`2020_14-uniref-download`) and finally submit a PR against it ie:
```
2020_14-uniref-customise-columns <-- 2020_14-uniref-download
```
Then once approved merge it all in to the sprint branch.
In uniref and uniprotkb results.
No
Yes

# Changes
* Moved customise table and column select to shared directory and are now functions of namespace.
* Customise table is now a sliding panel and positioned on the left to avoid scrollbar issues observed when on the right.
* Created `sticky-bottom-right` class and used this in the blast panel and the customise table.
* Replaced `mapDispatchToProps`, `mapStateToProps` with `useDispatch`, `useSelector`. Why? Because that resolved a type error I couldn't get rid of with mapDispatchToProps, mapStateToProps.

# Unrelated changes
* Fixed CatalyticActivity tests so console errors are caught as they were annoying me.
* Don't include `__test-helpers__` in test global coverage count

# Notes
* There is a warning in the console about having two #checkbox-lineage which is due to the results fields json file having lineage repeated twice (`curl -s https://wwwdev.ebi.ac.uk/uniprot/api/configure/uniprotkb/result-fields | jq '.' | grep -F -B 3 -A 3 '"name": "lineage"'`). This warning existed before this PR and I'm looking into this with the backend.


src/shared/components/customise-table/CustomiseTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <form
+      onSubmit={handleSubmit}
+      className="customise-table"
+      data-testid="customise-table-form"
+    >
+      <ColumnSelect
+        onChange={handleChange}
+        // TODO temporary casting to UniProtKBColumn to make TS happy
```
xwatkins, line 43: think you can remove that now?

src/shared/components/layouts/SlidingPanel.tsx
```diff
@@ -4,22 +4,31 @@ import cn from 'classnames';
 
 import './styles/sliding-panel.scss';
 
-type Position = 'top' | 'bottom' | 'left' | 'right';
+export enum Position {
```
xwatkins, line 7: As they are used as css properties, I wonder if it might be better to have them all lowercase?

src/shared/components/column-select/utils/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+type PreparedField = {
+  id: Column;
+  label: string;
+  key: string;
+};
+
+export const prepareFields = (fields: ReceivedField[], exclude: Column[]) => {
+  const prepared: PreparedField[] = [];
+  fields.forEach(({ label, name, id }) => {
```
xwatkins, line 23: chained `filter()` and `map()` could also work here

xwatkins (changes requested):
Something I've noticed - which I guess was broken before - is that when you sort by a column and then remove it the sort order and column are still in the url: they should be removed upon column selection if that column has been removed.

dlrice:
> Something I've noticed - which I guess was broken before - is that when you sort by a column and then remove it the sort order and column are still in the url: they should be removed upon column selection if that column has been removed.

👍 Good catch. Fixed.

src/shared/components/column-select/ColumnSelect.tsx
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  return (
+    <div className="column-select">
+      <ColumnSelectDragDrop
+        columns={fieldDataForSelectedColumns}
+        onDragDrop={handleDragDrop}
+        onRemove={handleSelect}
+      />
+      <button
```
aurel-l, line 122: Just a note that new usages of buttons from Franklin should probably use the `Button` component now

dlrice: Replaced in the files that I have touched with the exception of drag and drop which is covered by https://www.ebi.ac.uk/panda/jira/browse/TRM-24303

aurel-l (approved):

xwatkins (approved):

aurel-l:
I'll just merge that because I need those changes, and also mine from my other PR, to move forward with a task. I'll slightly lower the test coverage for the rest of the sprint. Make sure that at some point you do un-skip the tests you mentioned in the PR description

aurel-l:
Also, I don't want to mess up you local setup for now if you have ongoing code changes, so I merged without squashing and didn't delete the branch, just in case. Please delete it eventually.

PR #228, opened 2020-12-04
2020 14 uniref download

dlrice:
https://www.ebi.ac.uk/panda/jira/browse/TRM-25197
UniProtKB and UniRef results --> Download
No
N/A


src/shared/config/resultsDownload.ts
```diff
@@ -0,0 +1,42 @@
...
+  [FileFormat.rdfXml]: 'rdf',
+  [FileFormat.text]: 'txt',
+  [FileFormat.gff]: 'gff',
+  [FileFormat.list]: 'list',
+  [FileFormat.json]: 'json',
+};
+
+export const fileFormatsWithColumns = [FileFormat.tsv, FileFormat.excel];
+
+export const nsToFileFormatsResultsDownload: Partial<Record<
```
aurel-l, line 36: I guess the Partial here is supposed to be temporary, otherwise TypeScript would complain about exhaustivity of the Record.
We'll need to make sure to update them when we do add more namespace.
In other places we have put all the namespace, but the value either null, or empty string when not implemented yet. But it's also not a solution to have us remember about it later.
I'd say add a `// TODO:` comment before this and other constructs like these for now, not sure if there's a better way for us to remember to change the code and remove `Partial` eventually.

dlrice: Good idea - added some comments for our future selves.

src/shared/components/download/Download.tsx
```diff
@@ -0,0 +1,258 @@
...
+        <label htmlFor="data-selection-true">
+          <input
+            id="data-selection-true"
+            type="radio"
+            name="data-selection"
+            value="true"
+            checked={downloadAll}
+            onChange={handleDownloadAllChange}
+          />
+          Download all ({totalNumberResults})
```
aurel-l, line 186: It probably needs franklin's `LongNumber` here

src/shared/components/download/Download.tsx
```diff
@@ -0,0 +1,258 @@
...
+          </Fragment>
+        )}
+        <section className="button-group sliding-panel__button-row sticky-bottom-right">
+          <Button variant="secondary" type="button" onClick={() => onClose()}>
+            Cancel
+          </Button>
+          <Button
+            variant="secondary"
+            type="button"
+            onClick={() => handlePreview()}
```
aurel-l, line 246:
```suggestion
            onClick={handlePreview}
```

src/tools/components/ResultDownload.tsx
```diff
@@ -109,13 +110,7 @@ const ResultDownload = memo<ResultDownloadProps>(
...
     const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
       e.preventDefault();
-      const link = document.createElement('a');
-      link.href = toolsURLs(jobType).resultUrl(id, fileFormat);
-      link.target = '_blank';
-      link.setAttribute('download', '');
-      document.body.appendChild(link);
-      link.click();
-      document.body.removeChild(link);
+      downloadFileInNewTab(toolsURLs(jobType).resultUrl(id, fileFormat));
```
aurel-l, line 113: Not sure why this needs to be done instead of just having the submit button be a link to the output of `toolsURLs(jobType).resultUrl(id, fileFormat)` with all the attributes as set inside that `downloadFileInNewTab` function

aurel-l:
As a general comment, I think the `ErrorBoundary` component needs to be used in the 2 places where the download logic is used.
To reproduce, just make it throw an error before the render of one of the 2 download forms you have and place the error boundary so that the least amount of what was already in the page gets replaced by the error message (as in "it was already there, don't remove it if there's an issue in the download logic").

I would thinking the error needs to be displayed within the sliding panel.
Maybe wrapping the `children` bit of the return of the sliding panel (SlidingPanel.tsx l.32) would be a good place to put that boundary.

src/tools/components/ResultDownload.tsx
dlrice: When the submit button is clicked we download in a new tab as well as call `onToggleDisplay` - are you thinking have the button as a link as well as having an onClick handler to deal with `onToggleDisplay`?

aurel-l: Yep. As long as you don't block the normal click event you should still be able to do both things

dlrice: Updated

aurel-l (approved):

PR #229, opened 2020-12-08
2020 14 tiles update

xwatkins:
## Purpose
Update the home page layout and tiles to match the new mockup: https://www.figma.com/file/z9fShsjTqUxbsERlKKVDlZ/Homepage-design?node-id=0%3A1

## Approach
- Use the new grid to implement the layout
- Update the tile component, as well as text
- Add and optimise SVGs as Tile background

## Testing
- Updated snapshots

## Checklist


aurel-l:
Since those SVGs are just added to this project once and not processed or optimised through the build logic. Have you optimised them already?
I've run all the franklin icons through this tool https://jakearchibald.github.io/svgomg/ and made sure I've used a compression level where there would be no visual difference between the initial image and the optimised one.

src/app/components/HomePageNonCritical.tsx
```diff
@@ -1,72 +1,181 @@
 import { Tile, HeroContainer } from 'franklin-sites';
 
-import Block from './Block';
+import colors from '../../../node_modules/franklin-sites/src/styles/colours.json';
+
+// Namespaces
+import UniProtKBIllustration from '../../svg/uniprotkb_illustration.svg';
```
aurel-l, line 6: All of those I would probably prefer to load them as actual files. Here we are using them as images so we don't have any advantage bundling them inside react code.
Loading them as files would:
 - improve caching (at the cost of more different requests)
 - reduce javascript payload (quicker to download and to process)
 - make the image rendering non-blocking for the page (can appear later, it's fine)
Either make it specific per-import by doing something like `import uniProtKBIllustrationURL from '!file-loader!../../svg/uniprotkb_illustration.svg';` or change the webpack config to have all the SVGs inside that `svg` folder use `file-loader`.
In any case, usage would be as `backgroundImage={}`

xwatkins: TRM-25295

aurel-l (approved):

PR #232, opened 2020-12-10
2020 14 column renderers changes

aurel-l:
partial PR

 - Review and change column renderers for UniRef and UniProtKB. Highlighting:
   - Use more ExpandableList
   - Change output of the render method to `ReactNode`
   - Use more optional chaining
   - Remove more uses of `v1`
   - New UniParc ID column renderer for UniProtKB
   - More spacing 😅

Please *ignore* the columns renderers for UniParc as they are work in progress.


src/uniprotkb/adapters/uniProtkbConverter.ts
```diff
@@ -72,6 +72,8 @@ export type UniProtkbAPIModel = {
   entryAudit?: EntryAudit;
   references?: Reference[];
   lineages?: LineageData[];
+  // How is that defined? What goes in this?
+  extraAttributes?: Record<string, unknown>;
```
xwatkins, line 76: This is actually something we should be using for UniProtKB cards - we are currently getting pretty much all (or maybe all?) the domain to be able to count the number of items to display at the bottom of cards. This was added later by backend to prevent us from doing so and we should update it. I will create a Jira.

xwatkins (approved):

PR #235, opened 2020-12-15
2020 14 uniparc results

aurel-l:
## Purpose
Colum renderers for UniParc [TRM-25177](https://www.ebi.ac.uk/panda/jira/browse/TRM-25177)

## Approach
It defines _some_ of the columns needed for UniParc

## Testing
Snapshots of the columns

## Checklist
   - not really, but actually the most of it has already been reviewed, the content of this PR only covers [this commit](https://github.com/ebi-uniprot/uniprot-website/commit/9fb0588b58366c3ce96d8d521a504c7cf566b657).
   - missing proper renderers for multiple columns, and added UniParc to generalisation code that was not covered before too

Need to have a chat with backend because there are a lot of issues with different columns. Test only the ones that are clearly defined. The ones returning `null` need to be iterated upon after backend fixes/changes.

I guess it's better to get this through now because other people are starting to work on UniParc and that would get some of the generalisation code out of the way and accessible to others


aurel-l:
By the way, it also solves [TRM-25179](https://www.ebi.ac.uk/panda/jira/browse/TRM-25179) for free

its-mahdi (approved):

PR #236, opened 2020-12-15
2020 15 uniparc entry sequence

xwatkins:
## Purpose
Create the basic scaffold of the UniParc entry page, and add the section to display the sequence.
https://www.ebi.ac.uk/panda/jira/browse/TRM-25289

## Approach
Took the same approach as in UniRef, and extracted the `SequenceSection` component to the shared directory.

## Testing
The tests check that some of the expected text is rendered when the request is successful. They also check that the loading component is displayed and the error component when something goes wrong.

## Checklist


src/shared/components/entry/SequenceSection.tsx
```diff
@@ -1,17 +1,22 @@
...
-import { formatLargeNumber, hasContent } from '../../../shared/utils/utils';
+import { formatLargeNumber, hasContent } from '../../utils/utils';
 
 import EntrySection, {
   getEntrySectionNameAndId,
-} from '../../types/entrySection';
+} from '../../../uniref/types/entrySection';
 
-import { SequenceUIModel } from '../../../uniprotkb/adapters/sequenceConverter';
+export type Sequence = {
```
aurel-l, line 10: Since it's used in other places maybe this should live in `shared/types` rather than within `shared/components`

aurel-l (approved):
LGTM
Didn't realise before that the `primaryAccession` in the sequence section was actually never used, good catch

PR #238, opened 2020-12-22
2020 14 uniparc results

aurel-l:
## Purpose
Search results page for UniParc [TRM-25177](https://www.ebi.ac.uk/panda/jira/browse/TRM-25177)

## Approach
 - Have a renderer for every column in the UniParc search results page
 - Added facets

## Testing
Snapshot for every column renderer + 2 edge cases tests

## Checklist
  - 😬 I also fixed an issue when switching namespaces when in search results page (before it would use stale data with new renderers)


src/uniparc/config/UniParcColumnConfiguration.tsx
```diff
@@ -92,20 +172,54 @@ UniParcColumnConfiguration.set(UniParcColumn.organismID, {
 
 UniParcColumnConfiguration.set(UniParcColumn.organism, {
   label: 'Organisms',
-  // TODO:
-  render: () => null,
+  /* TODO:
```
aurel-l, line 175: This data is missing from the payload

src/uniparc/config/UniParcColumnConfiguration.tsx
```diff
@@ -58,6 +65,71 @@ export const UniParcColumnConfiguration = new Map<
   }
 >();
 
+/* NOTE:
```
xwatkins, line 68: Seems like a good idea - would move to shared utils though?

src/shared/components/results/ResultsContainer.tsx
```diff
@@ -84,10 +84,11 @@ const Results: FC = () => {
   const initialApiUrl = getAPIQueryUrl({
     namespace,
     query,
-    columns,
+    columns: namespace === Namespace.uniparc ? undefined : columns,
```
xwatkins, line 87: Not too keen on that but it's linked to the hack L81 (instead of using `[]` we should explicitly define the columns we want to retrieve for cards and for table for each namespace, and I guess "undefined" would be accepted). Maybe move this to L81 so all the hack is in the same place and add a comment here https://www.ebi.ac.uk/panda/jira/browse/TRM-25288

src/shared/components/results/ResultsView.tsx
```diff
@@ -104,8 +109,10 @@ const ResultsView: FC<ResultsTableProps> = ({
   const initialApiUrl = getAPIQueryUrl({
     namespace,
     query,
-    columns,
+    columns: namespace === Namespace.uniparc ? undefined : columns,
```
xwatkins, line 112: Same comment as above

xwatkins (approved):

src/uniparc/config/UniParcColumnConfiguration.tsx
aurel-l: will do

src/shared/components/results/ResultsContainer.tsx
aurel-l: I can't really do that in l.81 as the value as it is in l.81 is required in l.180 and passed down to the result view component in order to render the correct columns. l.87 is the only place where I need to override it

xwatkins: Ok just add to the Jira 👍

PR #241, opened 2020-12-23
first eslint custom rules

aurel-l:
## Purpose
Allow adding new custom rules specific for our coding guidelines
For now, added 2 custom rules:
 - a return statement has to be on its own line
 - link generation for the `<Link />` component, and for direct use through `history` methods, need to use the UniProt specific Location object ([see Jira here](https://www.ebi.ac.uk/panda/jira/browse/TRM-25235))

## Testing
Tests on the linter rules using eslint test helper. *But*, they are not connected to the main codebase testing, and there is no linting on the linting rules themselves. -> How does that need to be handle? Does that actually need to be done?

## Checklist
  - Does 2 things, introduces new linting rules, and fixes them in the codebase


its-mahdi (approved):
All looks good to me. Just one point: using a generic rule that forces block statements instead of one-liners would have solved the "return on it's own line" issue too.

https://eslint.org/docs/rules/curly

I think this is fine for now, but up to you if you like to replace this with `curly` rule for example.

aurel-l:
Yeah, I should probably replace that with the existing "curly" rule. As you can guess I did this one because it was an easy one to implement and to test how custom rules worked before doing the other more complex rule.

its-mahdi:
Yes, the custom rules look good! Up to you if you want to merge now or add curly first.

its-mahdi:
Looks great!

PR #242, opened 2020-12-23
Add logic for examples and text

xwatkins:
## Purpose
Add search examples, upload lists link and update list of examples and search input header sentence when the namespace selector is used on the homepage.

## Approach
Hoisted the namespace state hook to the parent component so that the header sentence can be updated after selection. Added buttons to add examples to the search box but the behaviour of this is TBC.

## Testing
- Update snapshots
- Add a test for namespace selection
- Add a test for example selection

## Checklist

### DIFF
--- a/src/app/components/HomePage.tsx
+++ b/src/app/components/HomePage.tsx
@@ -22,11 +22,15 @@ const UniProtFooter = lazy(
 const mission =
   'UniProt is the world’s leading high-quality, comprehensive and freely accessible resource of protein sequence and functional information.';
 
+// TODO remove "Partial" once all namespaces are added
 const namespaceFindYour: Partial<Record<Namespace, string>> = {
   [Namespace.uniprotkb]: 'protein',
   [Namespace.uniref]: 'protein cluster',
   [Namespace.uniparc]: 'protein sequence',
   [Namespace.proteomes]: 'proteome',
+  [Namespace.taxonomy]: 'taxon',
+  [Namespace.publications]: 'publication',
+  [Namespace.keywords]: 'keyword',
 };
 
 const HomePage = () => {
--- a/src/app/components/HomePage.tsx
+++ b/src/app/components/HomePage.tsx
@@ -22,8 +22,7 @@ const UniProtFooter = lazy(
 const mission =
   'UniProt is the world’s leading high-quality, comprehensive and freely accessible resource of protein sequence and functional information.';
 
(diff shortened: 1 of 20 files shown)


src/app/config/urls.ts
```diff
@@ -80,7 +80,9 @@ export const jobTypeToPath = (type: JobTypes, result?: boolean) => {
     case JobTypes.BLAST:
       return LocationToPath[result ? Location.BlastResult : Location.Blast];
     case JobTypes.IDMAP:
```
aurel-l, line 82: If we are removing the name of id mapping, do change the job type too to avoid having places with IDmapping or related terms, and others saying upload list

xwatkins: Good point, missed that.

src/tools/dashboard/components/Dashboard.tsx
```diff
@@ -39,7 +39,10 @@ const Dashboard = () => {
       <div>
         Try using <Link to={LocationToPath[Location.Blast]}>BLAST</Link>,{' '}
         <Link to={LocationToPath[Location.Align]}>Align</Link>,{' '}
-        <Link to={LocationToPath[Location.IDMap]}>ID Mapping/Retrieve</Link> or{' '}
+        <Link to={LocationToPath[Location.UploadList]}>
+          ID Mapping/Retrieve
```
aurel-l, line 43: Is that kept on purpose?

xwatkins: Yes - need to talk with UX about it as we need to call this functionality in a consistent way. Testing showed that mentioning "list of ids" worked well on the home page but existing users were keener on having "ID mapping" as a header. So it needs more research.

src/app/components/HomePage.tsx
```diff
@@ -20,29 +22,58 @@ const UniProtFooter = lazy(
...
-      </ErrorBoundary>
-    </main>
-    <Suspense fallback={null}>
-      <ErrorBoundary>
-        <UniProtFooter />
-      </ErrorBoundary>
-    </Suspense>
-  </>
-);
+// TODO remove "Partial" once all namespaces are added
```
aurel-l, line 25: I think the comment doesn't apply as it's covering all the currently defined Namespaces, and the `Partial` should be removed.

aurel-l (approved):

PR #250, opened 2021-01-08
basic lazy component preload proposal

aurel-l:
## Purpose
Proposal for a wrapper around react's `lazy` method to load component to add a preload capability

## Approach
 - Added a `preload` method to react's lazy component to use on mouse over, on focus, or on any occasion that we might see fit.
 - As an example, applied it to the query builder component, to start loading the component when the user is about to need it.

## Testing
No unit testing (quite tricky to unit-test). Check the network tab when using the preload feature and when actually needing the component, the corresponding webpack chunks should only load once. React's normal lazy feature should be conserved.

## Checklist


xwatkins (approved):

xwatkins:
LGTM - so we explicitly call `preLoad` where we think it would be good to do it. That might be the tricky bit!

aurel-l:
There's only so much magic we can do 😅 I think it's already a good improvement to be able to do that in a single `.preload()` call, but we need to think exactly where we want to do this and under which conditions. Maybe later, with actual usage we can realise if there is a way to automatise this if we are using common heuristics in different places.

PR #251, opened 2021-01-08
2020 15 side panel click outside

dlrice:
## Purpose
[Close sliding panel when user clicks outside.](https://www.ebi.ac.uk/panda/jira/browse/TRM-25265)

## Approach
* Inside of SlidingPanel added a listener to the body for mouse down events and calls the prop onClickOutside if the target is outside.
* Updated instances of SlidingPanel in the codebase to use this behaviour.

## Testing
* Click outside of the component
* Click inside of the component

## Checklist

### DIFF
--- a/src/shared/components/layouts/SlidingPanel.tsx
+++ b/src/shared/components/layouts/SlidingPanel.tsx
@@ -38,7 +38,7 @@ const SlidingPanel: FC<{
   useEffect(() => {
     if (onClickOutside) {
       const handleClickOutside = (e: MouseEvent) => {
-        if (!node?.current?.contains(e.target as Node)) {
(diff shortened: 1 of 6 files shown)


aurel-l:
Looks good, but I have some doubts about what should be the behaviour:
 - Should we cancel the normal event if the click is outside, as to prevent accidentaly clicking a button or a link in the main area when what the user wanted was just to close the side panel?
 - Should we just call the `onClickOutside` `onClose` and handle over events that might want to mean "close the side panel" for a normal user? On the top of my head I'm thinking about the `esc` key being pressed while focus is inside, but there might be some more accessibility consideration we need to take into account.
 - Should than handler be made mandatory?

Tagging @its-mahdi and @xwatkins to keep them in the loop.

PR #256, opened 2021-01-15
Update model

xwatkins:
## Purpose
Model change in backend regarding the gene centric service

## Approach
Updated the type definition and fields accordingly

## Testing
Updated test cases to match new model

## Checklist


src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
```diff
@@ -15,15 +15,20 @@ import apiUrls from '../../../shared/config/apiUrls';
...
+  id: string;
+  sequence: Sequence;
   entryType: string;
-  sequenceLength: number;
+  uniProtkbId: string;
+  proteinName: string;
+  organism: OrganismData;
   geneName: string;
-  geneNameType: string;
+  proteinExistence: string;
```
aurel-l, line 29: This feels like it could be a union of possible values instead of the generic `string`

src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
```diff
@@ -15,15 +15,20 @@ import apiUrls from '../../../shared/config/apiUrls';
...
+  flagType: string;
```
aurel-l, line 30: Same thing here (also, what is this? Could you add a comment above to explain)

src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
```diff
@@ -40,9 +45,9 @@ const ComputationalyMappedSequences: FC<{ primaryAccession: string }> = ({
   const columns = useMemo(
```
aurel-l, line 45:
```suggestion
  const columns = useMemo<Array<
    label: string;
    name: string;
    render: (data: ProteinEntryLight) => ReactNode;
  >>(
```
Could probably allow you to remove all the `...: ProteinEntryLight) => ...` below

src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
xwatkins: Same as https://github.com/ebi-uniprot/uniprot-website/blob/master/src/tools/blast/types/apiSequenceData.ts#L23 - it is free text (e.g. "1: Evidence at protein level")

aurel-l: Looks like there's only 5 possible strings here https://www.uniprot.org/help/protein_existence

src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
xwatkins: Just checked with Jie, `flagType` will be removed as the value should only ever be "Fragment". Currently we also have "Precursor" but that is a bug and should be removed at the next release. Removing it from the type def.

src/uniprotkb/components/entry/ComputationallyMappedSequences.tsx
xwatkins: You're correct, it's not free text - but I remember why this was added as a string, it's because the format doesn't match what is displayed on the current website (no numbers) and so it was planned to strip the number at some point. Added an enum for now and will make a minor jira to remember to update the string when displaying it

aurel-l (approved):

PR #257, opened 2021-01-19
2021 01 similar proteins

xwatkins:
## Purpose
Adding the "Similar Proteins" section to the UniProtKB entry.
This is the first part of the work required, retrieving the UniRef data, organising it in tabs and displaying the basics. Next step will be to retrieve additional data about the "members" and deciding on the best way to display it based on input from UX.

## Approach
- Add a data "converter" to extract the list of isoforms and add it to the UI model object
- Create a new component which uses the list of isoforms and canonical accessions to query the UniRef endpoint
- Filter out unwanted data (canonical and isoform accessions from results) and group both by Identity level and by UniProtKB accession
- Display the results in tabs based on groups defined above.

## Testing
Add a test for the data converter as well as the new `SimilarProteins` component, mocking data retrieval and checking that tabs work properly.

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/SimilarProteins.tsx
+++ b/src/uniprotkb/components/entry/SimilarProteins.tsx
@@ -1,10 +1,14 @@
 import { Loader, Message, Tabs, Tab, Card, Button } from 'franklin-sites';
 import { FC, useMemo } from 'react';
-import { groupBy } from 'lodash-es';
+import { groupBy, sortBy } from 'lodash-es';
 import { generatePath, Link } from 'react-router-dom';
 import { getClustersForProteins } from '../../../shared/config/apiUrls';
 import useDataApi from '../../../shared/hooks/useDataApi';
-import { UniRefLiteAPIModel } from '../../../uniref/adapters/uniRefConverter';
+import {
+  UniRefEntryType,
+  UniRefEntryTypeToPercent,
+  UniRefLiteAPIModel,
+} from '../../../uniref/adapters/uniRefConverter';
 import EntrySection, {
   getEntrySectionNameAndId,
 } from '../../types/entrySection';
@@ -27,8 +31,9 @@ const SimilarProteins: FC<{
       const { results } = data;
       // Remove all items with only 1 member as it will be canonical/isoform
       const filtered = results.filter((item) => item.members.length > 1);
+      const ordered = sortBy(filtered, 'entryType');
 
-      const clusterTypeGroups = groupBy(filtered, 'entryType');
+      const clusterTypeGroups = groupBy(ordered, 'entryType');
       // Note: thought we could use Dictionary<T> here but lodash doesn't
       // seem to export it
       const allClusterTypesGroups: {
@@ -64,7 +69,13 @@ const SimilarProteins: FC<{
       <Card title={nameAndId.name}>
         <Tabs>
           {Object.keys(clusterData).map((clusterType) => (
-            <Tab id={clusterType} title={clusterType} key={clusterType}>
+            <Tab
+              id={clusterType}
+              title={`${
+                UniRefEntryTypeToPercent[clusterType as UniRefEntryType]
+              } identity`}
+              key={clusterType}
+            >
               {Object.keys(clusterData[clusterType]).map((representativeId) => (
                 <section key={representativeId} className="text-block">
                   <h4>{representativeId}</h4>
--- a/src/uniprotkb/components/entry/SimilarProteins.tsx
+++ b/src/uniprotkb/components/entry/SimilarProteins.tsx
@@ -1,6 +1,6 @@
 import { Loader, Message, Tabs, Tab, Card, Button } from 'franklin-sites';
 import { FC, useMemo } from 'react';
-import { groupBy, sortBy } from 'lodash-es';
+import { groupBy } from 'lodash-es';
 import { generatePath, Link } from 'react-router-dom';
 import { getClustersForProteins } from '../../../shared/config/apiUrls';
 import useDataApi from '../../../shared/hooks/useDataApi';
@@ -31,9 +31,8 @@ const SimilarProteins: FC<{
       const { results } = data;
       // Remove all items with only 1 member as it will be canonical/isoform
       const filtered = results.filter((item) => item.members.length > 1);
-      const ordered = sortBy(filtered, 'entryType');
 
-      const clusterTypeGroups = groupBy(ordered, 'entryType');
+      const clusterTypeGroups = groupBy(filtered, 'entryType');
       // Note: thought we could use Dictionary<T> here but lodash doesn't
       // seem to export it
       const allClusterTypesGroups: {
@@ -68,7 +67,7 @@ const SimilarProteins: FC<{
     <div id={EntrySection.SimilarProteins}>
       <Card title={nameAndId.name}>
         <Tabs>
-          {Object.keys(clusterData).map((clusterType) => (
+          {Object.values(UniRefEntryType).map((clusterType) => (
             <Tab
               id={clusterType}
               title={`${
--- a/src/uniprotkb/components/entry/SimilarProteins.tsx
+++ b/src/uniprotkb/components/entry/SimilarProteins.tsx
@@ -1,0 +1,20 @@
+import { FC } from 'react';
+// import { Tabs } from 'franklin-sites';
+import { getClustersForProteins } from '../../../shared/config/apiUrls';
+import useDataApi from '../../../shared/hooks/useDataApi';
+
+const SimilarProteins: FC<{
+  isoforms: { isoforms: string[] };
+  primaryAccession: string;
+}> = ({ isoforms, primaryAccession }) => {
+  const searchUrl = getClustersForProteins([
+    primaryAccession,
+    ...isoforms.isoforms,
+  ]);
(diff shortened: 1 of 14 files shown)


src/uniprotkb/components/entry/SimilarProteins.tsx
```diff
@@ -0,0 +1,123 @@
...
+                                  }
+                                )}
+                              >
+                                {member}
+                              </Link>
+                            </li>
+                          ))}
+                        {row.members &&
+                          row.memberCount > row.members?.length && (
+                            <li>{row.memberCount} more</li>
```
aurel-l, line 105: This should probably be a link to the corresponding UniRef entry

src/uniprotkb/components/entry/SimilarProteins.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const nameAndId = getEntrySectionNameAndId(EntrySection.SimilarProteins);
+
+  // TODO get extra columns about the cluster members
+
+  return (
+    <div id={EntrySection.SimilarProteins}>
+      <Card title={nameAndId.name}>
+        <Tabs>
+          {Object.keys(clusterData).map((clusterType) => (
```
aurel-l, line 66: This needs to be sorted per identity somehow. Tabs in the current website are "100% Identity", "90% Identity", "50% Identity"

src/uniprotkb/components/entry/SimilarProteins.tsx
```diff
@@ -0,0 +1,123 @@
...
+                            {
+                              accession: row.id,
+                            }
+                          )}
+                        >
+                          {row.id}
+                        </Link>
+                      </h5>
+                      <ul className="no-bullet">
+                        {/* Note: move following to its own component  */}
```
aurel-l, line 86: I'm guessing this is pending, right?

src/uniprotkb/components/entry/SimilarProteins.tsx
```diff
@@ -0,0 +1,123 @@
...
+  primaryAccession: string;
+}> = ({ isoforms, primaryAccession }) => {
+  const searchUrl = getClustersForProteins([
+    primaryAccession,
+    ...isoforms.isoforms,
+  ]);
+  // Get the clusters in which the canonical and isoforms are found
+  const { loading, data, error } = useDataApi<{
+    results: UniRefLiteAPIModel[];
+  }>(searchUrl);
```
aurel-l, line 24: I'm guessing the default query has a result limit of 25 by default.
It would be good to have an idea of the maximum number of clusters one entry can belong to to know if we need to ask for more results.
Also, to mitigate this issue, maybe we should have 3 different calls for each tab/identity level. That would reduce network payload and also spread that 25 results limit across the 3 different identity levels

src/uniprotkb/components/entry/SimilarProteins.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <div id={EntrySection.SimilarProteins}>
+      <Card title={nameAndId.name}>
+        <Tabs>
+          {Object.keys(clusterData).map((clusterType) => (
+            <Tab id={clusterType} title={clusterType} key={clusterType}>
+              {Object.keys(clusterData[clusterType]).map((representativeId) => (
+                <section key={representativeId} className="text-block">
+                  <h4>{representativeId}</h4>
+                  {clusterData[clusterType][representativeId].map((row) => (
```
aurel-l, line 71: Not sure how I feel about this area as it is right now. Something tells me by looking at the structure of the code that you expected to have more information for each member, but you don't have it.
Is it pending a change in the API?
If not, we should maybe consider reducing the quantity of members displayed and focus on information from the UniRef entries themselves (reuse the UniRef result card here for example?).
Right now, visually, I'd say it looks like a temporary dump of data awaiting to be styled.

aurel-l:
I understand this is a "first part". So I commented things that might actually be things you planned on tackling on the second part, feel free to disregard them for now then.
But yeah, as said in the description, it definitely needs a further discussion with UX and also talk about technical considerations regarding how we get the data we need from the backend.

src/uniprotkb/components/entry/SimilarProteins.tsx
xwatkins: Good point about the limit, I'll check.

I wanted to wait until part II to decide on whether to do 1 call per tab or not - I will be using the accession service to get extra information about all the members which are displayed....and I'm hoping the list of members as strings won't go away once the new model is released!

src/uniprotkb/components/entry/SimilarProteins.tsx
xwatkins: Yes you are right, it's just a dump of data waiting to be styled at the moment.

src/uniprotkb/components/entry/SimilarProteins.tsx
aurel-l: I can't remember for sure what was decided, but I think the list of members as a string might go away, in favour of just using the `/members` endpoint, but that endpoint should give you enough information about the members themselves to not need to use the accession endpoint

src/uniprotkb/components/entry/SimilarProteins.tsx
xwatkins: Yes, depending on what's happening in part II with the retrieval of members info, I think this might end up being its own component

src/uniprotkb/components/entry/SimilarProteins.tsx
xwatkins: Issue then is that I would have to make a request per cluster and per isoform so not as good as one request for all clusters and 1 request for all members. I'm hoping it doesn't go away as part of the `uniref/search` service as this would be a problem for the results table too?

aurel-l: Well, the current way we are using the members column of the uniref configuration is working is not correct, as it's alway ever displaying up to 10 members only, so in the search results it always appears as that there is only ever up to 10 members in an entry...

aurel-l (changes requested):
OK, so the discussions points have been raised.
I guess most of them will be tackled in part 2.

The only required change that can go in this one is the topic of the ordering of the tabs, and maybe the tab naming (UniRefXX -> XX% Identity?)

aurel-l:
Not sure why, but this entry has "similar proteins", but the menu on the side appears inactive:
http://localhost:8080/uniprotkb/A0A0B7GQ86/entry

src/uniprotkb/components/entry/SimilarProteins.tsx
aurel-l: Part of this comment addressed now, still pending ordering

xwatkins:
> Not sure why, but this entry has "similar proteins", but the menu on the side appears inactive:
> http://localhost:8080/uniprotkb/A0A0B7GQ86/entry

Good catch... the logic to decide whether this is enabled or not is flawed, it currently relies on knowing whether there's isoforms or not but it should be on whether anything is found in the `SimilarProteins` component... Will have to have a think

xwatkins:
> > Not sure why, but this entry has "similar proteins", but the menu on the side appears inactive:
> > http://localhost:8080/uniprotkb/A0A0B7GQ86/entry
> 
> Good catch... the logic to decide whether this is enabled or not is flawed, it currently relies on knowing whether there's isoforms or not but it should be on whether anything is found in the `SimilarProteins` component... Will have to have a think

In the end it turns out every single (active) UniProtKB is a cluster member, so `Similar Proteins` should always be an active link.

aurel-l:
> > > Not sure why, but this entry has "similar proteins", but the menu on the side appears inactive:
> > > http://localhost:8080/uniprotkb/A0A0B7GQ86/entry
> > 
> > 
> > Good catch... the logic to decide whether this is enabled or not is flawed, it currently relies on knowing whether there's isoforms or not but it should be on whether anything is found in the `SimilarProteins` component... Will have to have a think
> 
> In the end it turns out every single (active) UniProtKB is a cluster member, so `Similar Proteins` should always be an active link.

Ah, it does help a lot.
I guess it could be the case the a UniRef100 entry only contains this entry, in which case the 100% identity tab would be empty, but I'm wondering if there could be a case where even at 90 and 50% identity an entry would still be alone in its cluster 🤔

aurel-l (approved):

xwatkins:
> > > > Not sure why, but this entry has "similar proteins", but the menu on the side appears inactive:
> > > > http://localhost:8080/uniprotkb/A0A0B7GQ86/entry
> > > 
> > > 
> > > Good catch... the logic to decide whether this is enabled or not is flawed, it currently relies on knowing whether there's isoforms or not but it should be on whether anything is found in the `SimilarProteins` component... Will have to have a think
> > 
> > 
> > In the end it turns out every single (active) UniProtKB is a cluster member, so `Similar Proteins` should always be an active link.
> 
> Ah, it does help a lot.
> I guess it could be the case the a UniRef100 entry only contains this entry, in which case the 100% identity tab would be empty, but I'm wondering if there could be a case where even at 90 and 50% identity an entry would still be alone in its cluster 🤔

It might be possible. I'm going to create an issue to look into refactoring the way "active/inactive" works because at the moment the process is redundant but I think it warrants its own task outside of this sprint.

PR #258, opened 2021-01-22
2021 01 taxonomy search facets

aurel-l:
## Purpose
Taxonomy filtering within facets [TRM-24701](https://www.ebi.ac.uk/panda/jira/browse/TRM-24701).

## Approach
Update franklin to have finer control over how facets are rendered, create custon facet for taxonomy.

## Testing
Updated existing snapshots.

## Checklist
  - improved test coverage in other areas of the codebase

This should work in any of the namespace searches, and this will not be rendered in the blast results (pending other jira later as logic is different).
Note that even though it is within the facets UI area, it is affecting the search filters, not the search facets.


src/shared/utils/__tests__/getNextURLFromHeaders.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    const headers = new Headers({
+      link: '<https://www.example.com/>; rel="next"',
+    });
+    const parsedHeaders = Object.fromEntries(headers.entries());
+    expect(getNextURLFromHeaders(parsedHeaders)).toBe(
+      'https://www.example.com/'
+    );
+  });
+
+  it("shouldn't return anything if no next URL if present", () => {
```
dlrice, line 14:
```suggestion
  it("shouldn't return anything if relationship is not next", () => {
```

src/shared/utils/__tests__/getNextURLFromHeaders.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    const headers = new Headers({
+      link: '<https://www.example.com/>; rel="canonical"',
+    });
+    const parsedHeaders = Object.fromEntries(headers.entries());
+    expect(getNextURLFromHeaders(parsedHeaders)).toBeUndefined();
+  });
+
+  it("shouldn't return anything if no next URL if present", () => {
```
dlrice, line 22:
```suggestion
  it("shouldn't return anything if no next URL is present", () => {
```

src/shared/components/results/ResultsFacets.tsx
```diff
@@ -1,50 +1,36 @@
...
+const ResultsFacets: FC<{ facets: FacetObject[]; isStale?: boolean }> = ({
   facets,
   isStale,
 }) => {
-  const match = useRouteMatch<{ subPage?: string }>(
-    LocationToPath[Location.BlastResult]
+  const isSearchPage = Boolean(useNS());
+
+  const splitIndex = facets.findIndex(
+    (facet) => facet.name === 'model_organism'
```
dlrice, line 19: Would be good to have facet names as an enum

src/uniparc/components/entry/Entry.tsx
```diff
@@ -52,7 +52,7 @@ const Entry: FC = () => {
     redirectedTo,
     headers,
   } = useDataApi<UniParcAPIModel>(baseURL);
-  // const facetData = useDataApi<Facet[]>(`${baseURL}/facets`);
+  // const facetData = useDataApi<FacetObject[]>(`${baseURL}/facets`);
```
dlrice, line 55: Are these two comments (also L34) needed? If so, could you comment with a `TODO:` to flag for future plans.

dlrice (commented):
Both of the following:
```
a4 AND (taxonomy_id:9606) OR (taxonomy_id:9605) => 194,337 results
a4 AND (taxonomy_id:9606) AND (taxonomy_id:9605) => 210 results
```
are represented the same in the Taxonomy filter:
```
Taxonomy
9606
9605
```
I am not sure if we want to be explict here by preceding by the logical conjunction in the facet. In another example:
```
a4 OR (taxonomy_id:9606) NOT (taxonomy_id:9606)
```
has the facets as:
```
Taxonomy
9606
9606
```

dlrice:
Came across this behaviour while reviewing: `a4 AND (taxonomy_id:9606) AND (taxonomy_id:9605) ` (note dangling space) causes the final term to be parsed as `All: (taxonomy_id:9605)` as opposed to the expected `Taxonomy [OC]:9605`

src/shared/components/results/ResultsFacets.tsx
aurel-l: I think that deserves a TODO.
Because we have a limited set of possible facets available on the server, depending on the endpoint/namespace. But we also have a subset of those facets that are the ones we actually query the server explicitely.
I think these need to be tied somehow (see the defaultFacets variable in the apiUrls.ts file).
So a change taking into account everything would be quite a major refactor don't you think?

src/uniparc/components/entry/Entry.tsx
aurel-l: This can actually go, I assumed I was asking backend for these facets in UniParc, but they need to be generated on the client

aurel-l:
> Came across this behaviour while reviewing: `a4 AND (taxonomy_id:9606) AND (taxonomy_id:9605) ` (note dangling space) causes the final term to be parsed as `All: (taxonomy_id:9605)` as opposed to the expected `Taxonomy [OC]:9605`

Good catch, I'll tackle that on Monday in a different commit directly on the sprint branch or on another PR depending on the size of the fix

PR #260, opened 2021-01-26
copy and inject styles

aurel-l:
Ugly trick in order to inject the styles we are interested in.

I made it flexible enough so that you can add more selectors as you realise they are needed inside the shadow root.
I just put 3 selectors at the moment, but we might actually need more.

### DIFF
--- a/src/uniprotkb/components/protein-data-views/SubcellularLocationView.tsx
+++ b/src/uniprotkb/components/protein-data-views/SubcellularLocationView.tsx
@@ -25,25 +25,38 @@ const getSubcelluarLocationId = (id: string) => {
   return null;
 };
 
+// 😷, get style rules from the document context in order to
+// copy and inject them in the shadow dom later through the template
+const getStylesContaining = (...selectors: string[]) =>
+  Array.from(document.styleSheets)
+    .flatMap((stylesheet) =>
+      Array.from(stylesheet.rules).map(
+        (rule) =>
+          selectors.some((selector) => rule.cssText.includes(selector)) &&
+          rule.cssText
+      )
+    )
+    .filter(Boolean) as string[];
+
 const SubcellularLocationView: FC<{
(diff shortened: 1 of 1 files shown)


src/uniprotkb/components/protein-data-views/SubcellularLocationView.tsx
```diff
@@ -25,25 +25,38 @@ const getSubcelluarLocationId = (id: string) => {
...
-      subcellularLocations
-        ?.flatMap(({ location }) => getSubcelluarLocationId(location.id))
-        .filter(Boolean)
-        .join(',')
-    );
     if (lineage && taxonId && comments && !isVirus(lineage)) {
-      ref.current.insertAdjacentHTML(
+      const sls = comments?.flatMap(({ subcellularLocations }) =>
+        subcellularLocations
+          ?.flatMap(({ location }) => getSubcelluarLocationId(location.id))
```
aurel-l, line 55: By the way, there is probably a typo here in the name of the function.
And also, here `.map` would work just fine as it's returning a single item.

dlrice: Good catch on the typo. I removed the inner `flatMap` but kept the outer as otherwise it's an array of array of strings that is being joined. Could work with just the `map` as joining arrays of arrays yields the required comma separated string as `['a','b'].toString() => 'a,b'` but it's not immediately clear to me that this is why it works out so prefer the explicit flatMap.

dlrice (approved):

src/uniprotkb/components/protein-data-views/SubcellularLocationView.tsx
aurel-l: Yeah, I was just mentionning the inner one. I didn't realise skipping the outer one still made it work with some magic joining logic, but I'm with you on keeping the outer one.

PR #263, opened 2021-01-29
2021 01 proteomes results api UI models

dlrice:
## Purpose
Proteomes namespace API and UI models needed.

## Approach
* Created API and UI models types
* Table view functional with currently supported column render functions (aside from BUSCO which is to follow in a separate PR)
* Moved `Xref`, `Citation` and `Lineage` types to shared apiModel file
* Added proteomes mock data provided by back end
* Temporarily renamed `proteomes` namespace to `proteome` until backend update as agreed in previous web meeting.

## Testing
Tested column render functions as we do for other namespaces.

## Checklist
  mostly but some things needed changing in the process


src/app/config/urls.ts
```diff
@@ -43,7 +43,7 @@ export const LocationToPath = {
   [Location.UniParcEntry]: '/uniparc/:accession',
   [Location.UniParcResults]: '/uniparc',
   [Location.ProteomesEntry]: '/proteomes/:accession',
-  [Location.ProteomesResults]: '/proteomes',
+  [Location.ProteomesResults]: `/${Namespace.proteomes}`,
```
dlrice, line 46: I am think tying everything to the `Namespace` enum would be a good thing to do more generally in our urls.

aurel-l: Good idea. In that case, do it for everything in this file of URLs.

src/proteomes/adapters/proteomesConverter.ts
```diff
@@ -0,0 +1,98 @@
...
+  proteomeCrossReferences: Xref[];
+  proteinCount: number; // used in the entry for each component TODO: eventually will be supported by backend in 2021_02 - 2021_03
+};
+
+export type CpdReport = {
+  proteomeCount: number;
+  stdCdss: number;
+  averageCdss: number;
+  confidence: number;
+  status: string;
```
aurel-l, line 29: would be nice to have a union of possible values here

src/proteomes/adapters/proteomesConverter.ts
```diff
@@ -0,0 +1,98 @@
...
+};
+
+export type BuscoReport = {
+  complete: number;
+  completeSingle: number;
+  completeDuplicated: number;
+  fragmented: number;
+  missing: number;
+  total: number;
+  lineageDb: string;
```
aurel-l, line 39: here too, I understand there is a limited set of possible values for this field too

src/proteomes/adapters/proteomesConverter.ts
```diff
@@ -0,0 +1,98 @@
...
+  score: number;
+};
+
+export type ProteomeCompletenessReport = {
+  cpdReport: CpdReport;
+  buscoReport: BuscoReport;
```
aurel-l, line 45: Are those (both cpd and busco) present on all proteome entries? I though some of them didn't have this as there is not always a busco visualisation on the current website

src/proteomes/config/ProteomesColumnConfiguration.tsx
```diff
@@ -0,0 +1,129 @@
...
+      })}
+    >
+      {id}
+    </Link>
+  ),
+});
+
+ProteomesColumnConfiguration.set(ProteomesColumn.organismID, {
+  label: 'Organism ID',
+  render: ({ taxonomy }) => taxonomy.taxonId,
```
aurel-l, line 62: Needs to be a link to the taxonomy page

src/proteomes/adapters/proteomesConverter.ts
dlrice: Had a chat with Katie Moulang regarding this. BUSCO release new lineageDb's over time so I will keep this as a string type to avoid having to monitor and update.

src/proteomes/adapters/proteomesConverter.ts
dlrice: * CPD always present (at least in current site data)
* BUSCO optional

dlrice:
Addressed all comments

aurel-l (approved):

PR #264, opened 2021-01-29
2021 01 similar proteins ii

xwatkins:
## Purpose
Second part of the "similar proteins" functionality. Display extra information about the cluster members.

## Approach
Created a new component which uses the accessions API to fetch the data for each list of Cluster members. Tried to re-use existing code to display the columns content in the data-table but this failed because of an issue with circular dependencies.

Still missing the query to "Show all" but waiting for backend to tell me what the best query is for that.

## Testing
Mocking the request to the UniRef API and to the accessions API, check presence of data and simulate click to change tab.

## Checklist


src/uniprotkb/components/entry/__tests__/SimilarProteins.spec.tsx
```diff
@@ -1,37 +1,39 @@
...
-  renderWithRouter(
-    <SimilarProteins
-      primaryAccession="P05067"
-      isoforms={{ isoforms: ['P05067-4'] }}
-    />
-  );
-describe('SimilarProteins tests', () => {
-  beforeEach(() => {
-    useDataApi.mockImplementation(() => dataMock);
+const getRendered = async () => {
```
aurel-l, line 15: This should probably be moved inside a `beforeEach` and have it rendered to a variable in scope of all the tests, in order to have the same pattern than most of the other tests have in the codebase.

src/uniprotkb/components/entry/__tests__/SimilarProteins.spec.tsx
```diff
@@ -1,37 +1,39 @@
...
+  return rendered;
+};
 
+describe('SimilarProteins tests', () => {
   it('should call useDataApi and render', async () => {
-    const { findByText } = getRendered();
-    expect(useDataApi).toHaveBeenCalled();
-    expect(await findByText(/0FGN2/)).toBeTruthy();
+    const { findAllByText } = await getRendered();
+    expect(await findAllByText(/0FGN2/)).toBeTruthy();
```
aurel-l, line 31:
```suggestion
    expect(await screen.findAllByText(/0FGN2/)).toBeTruthy();
```
See https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

src/uniprotkb/components/entry/__tests__/SimilarProteins.spec.tsx
```diff
@@ -1,37 +1,39 @@
...
   });
 
   it('should change tabs', async () => {
-    const { getByText, findByText } = getRendered();
-    fireEvent.click(getByText('50% identity'));
+    const { findByText } = await getRendered();
+    fireEvent.click(await findByText(/50% identity/));
```
aurel-l, line 36:
```suggestion
    fireEvent.click(await screen.findByRole('button', { name: /50% identity/ }));
```
See https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen and https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time (Especially if we're going to click it, it would be good to assert that this is a button by using this query)

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -83,56 +85,46 @@ const SimilarProteins: FC<{
...
                         )}
                         <hr />
                       </section>
                     )
                   )}
-                  {/* TODO: check with backend what query should be generated here */}
-                  <Button>View all</Button>
+                  {/* TODO: This query doesn't seem to work currently */}
+                  <Button
+                    onClick={() => {
```
aurel-l, line 116: This needs to be a link.
```
                  <Button
                    component={Link}
                    to={{
                        pathname: LocationToPath[Location.UniProtKBResults],
                        search: `query=(${data?.results
                          .filter(({ entryType }) => entryType === clusterType)
                          .map(({ id }) => `uniref_cluster_100:${id}`)
                          .join(' OR ')})`,
                    }}
                  >
```

src/uniprotkb/components/entry/__tests__/SimilarProteins.spec.tsx
```diff
@@ -1,37 +1,45 @@
...
 
   it('should change tabs', async () => {
-    const { getByText, findByText } = getRendered();
-    fireEvent.click(getByText('50% identity'));
-    expect(await findByText(/P12023/)).toBeTruthy();
+    fireEvent.click(await screen.findByRole('tab', { name: /50% identity/ }));
+    expect(await screen.findByText(/P12023/)).toBeTruthy();
+  });
+
+  it('should generate the correct link to view all', async () => {
```
aurel-l, line 39:
```suggestion
  it('should navigate to correct search page when clicking "view all"', async () => {
```
As you are actually clicking the button and not just checking that it's a link with the correct `href`

src/uniprotkb/components/entry/__tests__/SimilarProteins.spec.tsx
```diff
@@ -1,37 +1,45 @@
...
+    fireEvent.click(await screen.findByRole('button', { name: /View all/ }));
+    expect(historyMock.location.search).toEqual(
```
aurel-l, line 41: Maybe check `historyMock.location.pathname` too as we are changing the whole path and not just the `search` bit

aurel-l:
To comply with what we currently have on the website, the similar protein list shouldn't contain the one we are currently looking at.
If A and B are in the same cluster UniRef100_A, we should only display B in the list of members
Similarly, we shouldn't display a cluster at all if it contains only one member (it will be the one we are look at already)

aurel-l (changes requested):
To comply with what we currently have on the website, the similar protein list shouldn't contain the one we are currently looking at.
If A and B are in the same cluster UniRef100_A, we should only display B in the list of members
Similarly, we shouldn't display a cluster at all if it contains only one member (it will be the one we are look at already)

src/uniprotkb/components/entry/similar-proteins/SimilarProteinsTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+    return null;
+  }
+
+  const { results } = data;
+
+  return (
+    <DataTable
+      data={results}
+      columns={columnConfig}
+      getIdKey={(row: UniProtkbAPIModel) => row.primaryAccession}
```
aurel-l, line 88: If you merge the latest `2021_01` in, you'll be able to remove this type annotation, and the following `onSelect`

src/uniprotkb/components/entry/similar-proteins/SimilarProteinsTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+];
+
+const SimilarProteinsTable: FC<{ members: string[] }> = ({ members }) => {
+  const memberList = useMemo(() => {
+    // Remove UniParc entries
+    return members.filter((member) => !member.startsWith('UPI'));
+    return [];
+  }, [members]);
+
+  const columnConfig = [
```
aurel-l, line 28: This can (and should) go outside of the component.

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -112,19 +110,24 @@ const SimilarProteins: FC<{
...
-                  <Button
-                    onClick={() => {
-                      history.push({
-                        pathname: LocationToPath[Location.UniProtKBResults],
-                        search: `query=(${data?.results
-                          .filter(({ entryType }) => entryType === clusterType)
-                          .map(({ id }) => `uniref_cluster_100:${id}`)
-                          .join(' OR ')})`,
-                      });
+                  <Link
```
aurel-l, line 113: I meant, keep the `Button` component from Franklin, but make it be a react-router-dom `Link` by setting `element={Link}`

xwatkins: Hadn't realised this was an option.

xwatkins:
Think I've addressed all comments

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
aurel-l: 👆🏽

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
aurel-l: Does that not work using `Link`?

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
xwatkins: Ok I'm a little confused and am going to have another look at `<Button>`

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
aurel-l: https://github.com/ebi-uniprot/uniprot-website/pull/264#discussion_r567015967

xwatkins: Done now - don't think I'd fully grasped the way to use Button with the new implementation

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -0,0 +1,151 @@
...
+                            );
+                            return (
+                              <section key={row.id}>
+                                <h5>
+                                  <Link to={unirefEntryUrl}>{row.id}</Link>
+                                </h5>
+                                <SimilarProteinsTable members={row.members} />
+                                {row.memberCount - row.members.length > 0 && (
+                                  <Link to={unirefEntryUrl}>
+                                    {row.memberCount - row.members.length} More
```
aurel-l, line 112:
```suggestion
                                    {row.memberCount - row.members.length - 1} more
```

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -0,0 +1,151 @@
...
+                                accession: row.id,
+                              }
+                            );
+                            return (
+                              <section key={row.id}>
+                                <h5>
+                                  <Link to={unirefEntryUrl}>{row.id}</Link>
+                                </h5>
+                                <SimilarProteinsTable members={row.members} />
+                                {row.memberCount - row.members.length > 0 && (
```
aurel-l, line 110:
```suggestion
                                {row.memberCount - row.members.length - 1 > 0 && (
```

aurel-l (changes requested):
We need to remove the protein from the count here too.
Compare http://localhost:8080/uniprotkb/O43826/entry#similar-proteins with https://www.uniprot.org/uniprot/O43826#similar_proteins for example

xwatkins:
> We need to remove the protein from the count here too.
> Compare http://localhost:8080/uniprotkb/O43826/entry#similar-proteins with https://www.uniprot.org/uniprot/O43826#similar_proteins for example

Actually w

aurel-l (approved):

PR #265, opened 2021-02-01
2021 01 uniref truncated renderers

aurel-l:
## Purpose
Change the UniRef column renderers for the columns with a truncated list of values

## Approach
Replace ExpandableList with a list + a button linking to the entry itself

## Testing
Updated snapshots

## Checklist

Also...
 - Changed the custom eslint rule to handle a edge-case I had forgotten before
 - Created `getEntryPath` and `getEntryPathFor` functions to create URLs more easily for entry pages and applied it in all the column renderers for all namespaces
 - Replaced external links for UniParc XRefs to use Franklin's `ExternalLink` component


src/uniref/config/UniRefColumnConfiguration.tsx
```diff
@@ -69,42 +65,53 @@ UniRefColumnConfiguration.set(UniRefColumn.commonTaxon, {
...
+        <li key={organism}>{organism}</li>
       ))}
-    </ExpandableList>
+      {organisms.length > CUT_OFF && (
+        <Button
+          element={Link}
+          variant="tertiary"
+          to={getEntryPath(Namespace.uniref, id)}
+        >
+          more organisms
```
xwatkins, line 111: For consistancy, the "M" of more should be capitalised

src/uniref/config/UniRefColumnConfiguration.tsx
```diff
@@ -137,19 +144,33 @@ UniRefColumnConfiguration.set(UniRefColumn.types, {
...
+        </li>
       ))}
-    </ExpandableList>
+      {members.length > CUT_OFF && (
+        <Button
+          element={Link}
+          variant="tertiary"
+          to={getEntryPath(Namespace.uniref, id)}
+        >
+          {memberCount - CUT_OFF} more member
```
xwatkins, line 169: For consistancy, the "M" of more should be capitalised

xwatkins (changes requested):
One suggestion, would it be possible to link the "More" links to the "Members" section of the entry directly?

src/uniref/config/UniRefColumnConfiguration.tsx
aurel-l: Not here since there is a number in front of it

aurel-l:
> One suggestion, would it be possible to link the "More" links to the "Members" section of the entry directly?

I tried, but since the Sequence section is small, the Members section is always in view, and since it's in view, it won't scroll to it. It only scrolls to it if not in view at all.

xwatkins (approved):

PR #267, opened 2021-02-02
2021 01 uniparc entry xrefs

aurel-l:
## Purpose
Add more content to the UniParc entry page (for Jira [TRM-25180](https://www.ebi.ac.uk/panda/jira/browse/TRM-25180)).
Partial PR, does not tackled everything (see below)

## Approach
Add cross-reference table, and facet filtering

## Testing
Added new tests

## Checklist

What it should be doing:
 - Rendering list of UniParc cross references in the UniParc entry page
 - Render links to *some* of the cross-reference pages (either internal, for UniProtKB entries, or external URLs). Missing some, pending new endpoint
 - Render 3 facets. Clicking on these facets should update the URL correctly
 - Enrich the "Organisms" facet with taxonomy information (common or scientific name), the values (used in the URL) remain the taxon ID

What it doesn't do yet:
 - Customise columns
 - Render all external links
 - Apply facet filtering
 - ProtVista section (probably need an other Jira for that TBH)
I've paused this because of upcoming changes to the endpoint schemas, and new databases endpoint coming.


src/uniparc/components/entry/XRefsFacets.tsx
```diff
@@ -0,0 +1,169 @@
...
+            name: 'databases',
+            values: Object.entries(countBy(xrefs, (xref) => xref.database))
+              .map(([value, count]) => {
+                let label: ReactNode = value;
+                const entryType = getEntryTypeFromString(value);
+                if (entryType === EntryType.REVIEWED) {
+                  label = (
+                    <>
+                      <EntryTypeIcon entryType={entryType} />
+                      UniProtKB reviewed
```
xwatkins, line 142: This is also used in XrefSection (and maybe other parts of the website), maybe make it a string enum? Same with unreviewed.

src/uniparc/components/entry/XRefsSection.tsx
```diff
@@ -0,0 +1,226 @@
...
+import { Column } from '../../../shared/config/columns';
+import { UniParcXRef } from '../../adapters/uniParcConverter';
+import EntrySection, {
+  getEntrySectionNameAndId,
+} from '../../types/entrySection';
+import { EntryType } from '../../../uniprotkb/adapters/uniProtkbConverter';
+
+import './styles/XRefsSection.scss';
+
+const getColumns = (
```
xwatkins, line 27: I guess this will moved to a column config map after the API model changes?

xwatkins (changes requested):
A few minor comments. Could you please create an issue for adding ProtVista?

src/uniparc/components/entry/styles/XRefsSection.scss
```diff
@@ -0,0 +1,3 @@
+.xref-inactive {
```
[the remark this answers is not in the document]
aurel-l: Where else would it be used?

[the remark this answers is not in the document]
xwatkins: Anywhere where something is inactive, for instance for jobs that are obsolete etc... (I guess they already have similar styling?)

src/uniparc/components/entry/XRefsSection.tsx
aurel-l: Where this piece of logic is located doesn't really depend on the API model change. If we want that in another file we can do it.
I just followed the same approach that we have in other data tables within sections which is collocate the column configuration with the `<DataTable />` component in the same file (as it's not going to be used anywhere else).
See UniProtKB's SimilarProteinsTable and ComputationallyMappedSequences and UniRef's MembersSections

src/uniparc/components/entry/XRefsFacets.tsx
aurel-l: Well, there's actually a bug here, where `TREMBLNEW` is detected as `UniProtKB unreviewed` ('TREMBLNEW' is distinct from `TrEMBL`). I'll refactor the logic in the XRefsSections in order to be reused here.

src/uniparc/components/entry/styles/XRefsSection.scss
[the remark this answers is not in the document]
aurel-l: Ah, I see. I'll have a look.
I need to make sure that we don't disable clicks though, as we can have valid links in an inactive cross-reference, when I'm afraid we might deactivate all possible interactions with the jobs

src/uniparc/components/entry/XRefsSection.tsx
xwatkins: Difference being that this one is customisable whereas the others are not

aurel-l: So this one should stay and the others should move to distinct files?
Also, the one in ComputationallyMappedSequences should be outside of the render function anyway.

aurel-l:
> A few minor comments. Could you please create an issue for adding ProtVista?

Added Jira [TRM-25405](https://www.ebi.ac.uk/panda/jira/browse/TRM-25405)

src/uniparc/components/entry/styles/XRefsSection.scss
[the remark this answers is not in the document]
aurel-l: Yep, that's the case, I would hold off on that for now until a later style refactoring.

src/uniparc/components/entry/XRefsFacets.tsx
aurel-l: done

src/uniparc/components/entry/XRefsSection.tsx
xwatkins: Yep you're right about `ComputationallyMappedSequences` will update that.

What I meant was that because it's selectable columns, it might be better to use the same definition model used in `UniRefColumnConfiguration` and other namespaces. `UniRefEntryColumnConfiguration`  for instance.

aurel-l: Ah I see.
Will do that after the schema update then. Not sure yet if I will get a configure endpoint specific to UniParc entries (different than for UniParc results) or if I'll have to do that on the client somehow. The current column selection component assumes that only result page configuration endpoints exists and automatically loads it so I'd have to override that.
That needs to be done at a later point

xwatkins: Looks like this will be next sprint so unlikely we'll forget 🤞

xwatkins (approved):

PR #271, opened 2021-02-03
added/modified more proteome column renderers

aurel-l:
## Purpose
Added/modified some column configurations for Proteomes

## Approach
Render more columns

## Testing
Updated snapshots

## Checklist

Still pending changes from API

### DIFF
--- a/src/proteomes/adapters/proteomesConverter.ts
+++ b/src/proteomes/adapters/proteomesConverter.ts
@@ -78,12 +78,12 @@ export type ProteomesAPIModel = {
   taxonomy: Taxonomy;
   modified: string;
   proteomeType: ProteomeType;
-  components: Component[];
+  components?: Component[];
   citations: Citation[];
   annotationScore: number;
   superkingdom: string;
   proteomeCompletenessReport: ProteomeCompletenessReport;
-  genomeAssembly: GenomeAssembly;
+  genomeAssembly?: GenomeAssembly;
   geneCount: number;
   genomeAnnotation: GenomeAnnotation;
   taxonLineage: Lineage[];
--- a/src/proteomes/adapters/proteomesConverter.ts
+++ b/src/proteomes/adapters/proteomesConverter.ts
@@ -78,12 +78,12 @@ export type ProteomesAPIModel = {
   taxonomy: Taxonomy;
   modified: string;
   proteomeType: ProteomeType;
-  components?: Component[];
+  components: Component[];
   citations: Citation[];
   annotationScore: number;
   superkingdom: string;
   proteomeCompletenessReport: ProteomeCompletenessReport;
-  genomeAssembly?: GenomeAssembly;
+  genomeAssembly: GenomeAssembly;
   geneCount: number;
   genomeAnnotation: GenomeAnnotation;
   taxonLineage: Lineage[];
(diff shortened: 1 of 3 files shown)


src/proteomes/config/ProteomesColumnConfiguration.tsx
```diff
@@ -109,29 +116,49 @@ ProteomesColumnConfiguration.set(ProteomesColumn.lineage, {
...
-  label: 'CPD',
+  label: <abbr title="Complete Proteome Detector">CPD</abbr>,
   render: ({ proteomeCompletenessReport }) =>
     proteomeCompletenessReport.cpdReport.status,
 });
 
 ProteomesColumnConfiguration.set(ProteomesColumn.genomeAssembly, {
   label: 'Genome assembly ID',
-  render: ({ genomeAssembly }) => genomeAssembly?.assemblyId,
+  render({ genomeAssembly }) {
```
dlrice, line 126:
```suggestion
  render: ({ genomeAssembly }) => {
```

src/proteomes/adapters/proteomesConverter.ts
```diff
@@ -78,12 +78,12 @@ export type ProteomesAPIModel = {
   taxonomy: Taxonomy;
   modified: string;
   proteomeType: ProteomeType;
-  components: Component[];
+  components?: Component[];
```
dlrice, line 81: Do you know these are going to be optional from the API?

aurel-l: All of the fields that correspond to an API `field` value should be optional TBH, as to not rely on them in other renderers.
Or maybe the ResultPage should operate on `Partial<>` versions of the payload to be on the safe side?

src/proteomes/config/ProteomesColumnConfiguration.tsx
aurel-l: That's me that always tend to use an object value with an arrow function when the return value is directly returned, as in, it doesn't require curly braces. And always switch to the method syntax when it requires more logic.
But sure

src/proteomes/adapters/proteomesConverter.ts
dlrice: So far in the code base we are assuming the API will provide objects with some guaranteed attributes. I think assuming nothing is guaranteed is a bigger question that we would want to discuss with the team.

src/proteomes/config/ProteomesColumnConfiguration.tsx
dlrice: suggested for consistency with the others

src/proteomes/adapters/proteomesConverter.ts
aurel-l: Not for UniProtKB, but indeed we do for other namespaces. I reverted that for now and we should take a more general approach.
I had started added these `?:` a bit randomly when I got an error popping on my screen after toggling some columns.
I have narrowed that down to when I was switching the `lineage` column so I changed the `.map` to `?.map` instead. It's a race condition when the column is rendered but the new data with the `lineage` field isn't there yet.

A bit worried that might happen a bit randomly depending on which columns get added or removed.

dlrice: I haven't seen that race condition before now but now I do. Seems the problem would lie in column customisation & results view?

dlrice (approved):

src/proteomes/adapters/proteomesConverter.ts
aurel-l: Yes, I guess so. There's some dark magic in the ResultsView component to make this kind of thing work, and it's hard to test automatically, but I guess we're suppose to tackle this race condition there, in a similar way we're doing for the race condition when switching from table to card.

PR #272, opened 2021-02-04
use getEntryPath() everywhere possible

aurel-l:
## Purpose
Use helper function to create paths to entry pages

## Approach
Replace everywhere we were using the generatePath function with the different location objects with this helper function

## Testing
All the tests still pass (generated URLs should stay the same)

## Checklist

### DIFF
--- a/src/shared/components/entry/TaxonomyView.tsx
+++ b/src/shared/components/entry/TaxonomyView.tsx
@@ -41,7 +41,7 @@ const TaxonomyView: FC<TaxonomyDataProps> = ({ data }) => {
 
   const termValue = `${data.scientificName}${
     data.commonName ? ` (${data.commonName})` : ''
-  } ${data.synonyms && data.synonyms.length > 0 ? ` (${data.synonyms})` : ''}`;
+  } ${data.synonyms?.length ? ` (${data.synonyms.join(', ')})` : ''}`;
 
   return (
     <SimpleView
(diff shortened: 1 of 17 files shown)


src/shared/components/entry/TaxonomyView.tsx
```diff
@@ -25,17 +27,15 @@ export const TaxonomyId: FC<{ taxonId?: number }> = ({ taxonId }) => {
...
+        to={getEntryPath(Namespace.taxonomy, taxonId)}
       >{`${taxonId} `}</Link>
       <ExternalLink url={externalUrls.NCBI(taxonId)}>NCBI</ExternalLink>
     </>
   );
 };
 
 const TaxonomyView: FC<TaxonomyDataProps> = ({ data }) => {
-  if (!data) {
+  if (!data?.taxonId) {
```
its-mahdi, line 38: By checking only for `taxonId` here, we are making an assumption that the rest of the data being accessed below is safe, as long as the `taxonId` is present. Is that something we are confident will hold true always? I think the previous form which was checking for `!data` is more future proof.

aurel-l: This check is more restrictive than before as it's both checking for no `data` (as before), and if there is, for no `data.taxonId`. Not sure how you want me to change it to be even more restrictive.

its-mahdi (approved):

PR #273, opened 2021-02-04
hotfix for query builder when opened through the taxonomy fake facet

aurel-l:
## Purpose
Fix issue when injecting a new clause in the query builder via the fake taxonomy facet

## Checklist

### DIFF
--- a/src/query-builder/utils/queryStringProcessor.ts
+++ b/src/query-builder/utils/queryStringProcessor.ts
@@ -82,11 +82,11 @@ const getEmptyClause = (id: number): Clause => ({
   logicOperator: Operator.AND,
 });
 
-export const parse = (queryString = ''): Clause[] => {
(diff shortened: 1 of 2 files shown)


src/query-builder/utils/queryStringProcessor.ts
```diff
@@ -82,11 +82,11 @@ const getEmptyClause = (id: number): Clause => ({
   logicOperator: Operator.AND,
 });
 
-export const parse = (queryString = ''): Clause[] => {
+export const parse = (queryString = '', startId = 0): Clause[] => {
```
xwatkins, line 85: Could you add a comment about what the `startId` represents within the context of parsing the query string?

aurel-l: Done

aurel-l: Closing now so that it goes in the release

PR #276, opened 2021-02-04
Add reference proteome icon.

dlrice:
## Purpose
Add reference proteome icon from franklin. Part of https://www.ebi.ac.uk/panda/jira/browse/TRM-25360.

## Approach
Adds icon

## Testing
no change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "classnames": "2.2.6",
     "core-js": "3.8.2",
     "d3": "5.16.0",
-    "franklin-sites": "0.0.106",
+    "franklin-sites": "0.0.107",
     "history": "4.10.1",
     "idb": "6.0.0",
     "interaction-viewer": "3.1.2",
--- a/src/proteomes/components/AccessionView.tsx
+++ b/src/proteomes/components/AccessionView.tsx
@@ -1,0 +1,37 @@
+import { FC } from 'react';
+import { Link } from 'react-router-dom';
+
+import EntryTypeIcon from '../../shared/components/entry/EntryTypeIcon';
+
+import { getEntryPath } from '../../app/config/urls';
+
+import { Namespace } from '../../shared/types/namespaces';
+import {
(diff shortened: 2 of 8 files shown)


aurel-l:
I know the mockup puts this in the "Organism" column, but the organism is related to the taxon name. The icon being associated with the Proteome entry I would maybe add this at the beginning of the Entry column instead.

aurel-l:
> > I know the mockup puts this icon in the "Organism" column, but the organism is related to the taxon name. The icon being associated with the Proteome entry I would maybe add this at the beginning of the Entry column instead.
> 
> It's in the Organism column on the [current site](https://www.uniprot.org/proteomes/?query=mammalia&sort=score) too. Happy to break with tradition if you both are?
> Screenshot 2021-02-05 at 10 38 35

Oh, I see.
I'd ask for input from @xwatkins or @its-mahdi about that then.
Because, IMHO, in the image you are sharing it's even more obvious that something is off if the logo for "Reference *proteome*" is 7 columns away from the *proteome* entry.

xwatkins:
> > > I know the mockup puts this icon in the "Organism" column, but the organism is related to the taxon name. The icon being associated with the Proteome entry I would maybe add this at the beginning of the Entry column instead.
> > 
> > 
> > It's in the Organism column on the [current site](https://www.uniprot.org/proteomes/?query=mammalia&sort=score) too. Happy to break with tradition if you both are?
> > Screenshot 2021-02-05 at 10 38 35
> 
> Oh, I see.
> I'd ask for input from @xwatkins or @its-mahdi about that then.
> Because, IMHO, in the image you are sharing it's even more obvious that something is off if the logo for "Reference _proteome_" is 7 columns away from the _proteome_ ID column.

I would be happy to brake from tradition to make it more consistent with how we use logos for reviewed/unreviewed/uniparc

aurel-l (approved):

PR #277, opened 2021-02-05
Home page animation

aurel-l:
## Purpose
Implements a blinking cursor and progressive writing of text on the home page when switching namespace in the main search selector

## Approach
Disabled completely if user prefers reduced motion. Otherwise, cursor animation on hover of the hero banner, or when focus is within. Add letter by letter when switching namespace

## Testing
Rewrote existing test. Had to mock the `useReducedMotion` hook to simulate user preference of no motion. Testing only for this.

## Checklist


src/app/components/__tests__/__snapshots__/HomePage.spec.tsx.snap
```diff
@@ -4,19 +4,27 @@ exports[`HomePage component should render 1`] = `
...
-      class="hero-header"
+      class="hero-header home-page__header"
     >
       <div
         class="hero-header__content"
       >
         <h1>
-          Find your protein
+          Find your 
+          <button
```
xwatkins, line 14: Not sure this should be a button? I wouldn't expect it to be clickable and the action triggered is far from it.

src/app/components/HomePage.tsx
```diff
@@ -38,30 +49,95 @@ const namespaceFindYour: Record<Namespace, string> = {
...
+                  ) : (
+                    <>
+                      {displayed.split('').map((letter, index) => (
+                        // eslint-disable-next-line react/no-array-index-key
+                        <span key={index} className="letter">
+                          {letter}
+                        </span>
+                      ))}
+
+                      <span className="cursor">|</span>
```
xwatkins, line 119: Wondering if the cursor should disappear after the word has been written?

xwatkins (changes requested):
A couple of minor comments. Generally I think it works nicely as an additional way to grab the attention to the search box and make it look more modern.

I'm wondering whether the examples could be wired in to that PR - no worries if you feel that should be another PR https://www.ebi.ac.uk/seqdb/confluence/display/UniProt/Namespace+search+term+examples

src/app/components/__tests__/__snapshots__/HomePage.spec.tsx.snap
xwatkins: I understand this is subjective though, so if you feel strongly about it we could run it past the team "what would you expect to happen if you clicked on the text?"

aurel-l: Not sure either. I mainly made it a button because otherwise eslint would complain that this breaks 2 accessibility linting rules. We can either leave it as a button, or make it just plain inaccessible on purpose and override the rules here because the action it's covering is already handled by the actual dropdown.

aurel-l:
Just note that since it was a while ago that I did this PR, now that I looked at it again I found a new way to do this with less JavaScript and more CSS (should be smoother). But I did have to change how text and cursor was moving (should be less jumping around, so I believe it's better)

src/app/components/HomePage.tsx
aurel-l: Not sure if you've noticed but it's only flashing if the user's focus is within the hero header, or if their mouse is. The logic being to grab the attention on alternative actions (because clicking the text would expose the dropdown) is case the user is stuck looking at the header and wondering what they can do. If they're doing anything with the buttons at the top or with the rest of the page, I'm assuming they found something to do, so I don't grab the attention.
What I could do:
 - Change this to flashing only on hover
 - Just flash a couple of times then disappear

src/app/components/__tests__/__snapshots__/HomePage.spec.tsx.snap
xwatkins: I would go for option 2 - as you say it's already covered

src/app/components/HomePage.tsx
xwatkins: Wondering if flashing on hover could mean "You can edit me" so I would favour the second option?

aurel-l: Well, that the point. "Oh can I edit that? Let's try." clicks on it, it opens the dropdown "Oh, so _that_'s how I can 'edit' this"

xwatkins: That's not editing the text in place though (which I would think is the expected behaviour when you see a blinking cursor).

xwatkins (approved):

src/app/components/HomePage.tsx
aurel-l: Well, that can be removed altogether anyway. This PR started as an experiment, I'm not attached to the cursor.

PR #281, opened 2021-02-15
2021 02 optimise BLAST HSPs

aurel-l:
## Purpose
Prevent crashing the browser when calculating and displaying too many HSPs.

## Approach
 - Display them 10 by 10 instead of just all or nothing as before
 - Optimise calculating max scores by simplifying code to use for loops (would cause stack overflow issues otherwise in cases with too manys HSPs)

## Testing
Run BLAST on a Titin sequence (e.g Q9I7U4), max sure to ask for only 50 hits, but it will still take a while to run
Play with the "+n more" button on a BLAST result

## Checklist

### DIFF
--- a/src/tools/blast/components/results/BlastResultSidebar.tsx
+++ b/src/tools/blast/components/results/BlastResultSidebar.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react';
+import { memo } from 'react';
 import { useLocation } from 'react-router-dom';
 import { Loader, Facets, Facet } from 'franklin-sites';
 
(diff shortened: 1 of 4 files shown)


dlrice:
Looks good. In addition I think we should think about disabling/modifying the wrapped detail view as my machine can't cope with it.

dlrice (approved):

aurel-l:
> Looks good. In addition I think we should think about disabling/modifying the wrapped detail view as my machine can't cope with it.

Good point, I hadn't tried it and mine can't cope either.
I think this is part of the thresholds we need to define eventually to decide on how to handle extreme cases, so will need discussion.

PR #283, opened 2021-02-18
2021 01 subcell loc viz

dlrice:
## Purpose
[Integrate @swissprot/swissbiopics-visualizer for subcellular location visualisation](https://www.ebi.ac.uk/panda/jira/browse/TRM-25203)

## Approach
* Had a look at the source code at http://sp.sib.swiss/scripts/uniprot_entry-bundle.js and used this with some modifications.
* Uses tippy lib

An example [on SIB's dev site](http://sp.sib.swiss/uniprot/P05067) (needs VPN)

The following will be completed in subsequent PRs
* TODO: add additional GO template which will also require GO term data
* TODO: handle this case as seen in the source code 'svg .membranes .membrane.subcell_present' eg A1L3X0 doesn't work
* ~TODO: clicking a UniProtKBEvidenceTag doesn't open the content~
* ~TODO: Need to condition on the custom element being loaded rather than sleeping~

## Testing
None

## Checklist


src/uniprotkb/components/protein-data-views/SubcellularLocationWithVizView.tsx
```diff
@@ -0,0 +1,240 @@
...
+            top: ${pictureTop};
+          }
+          .subcell_name {
+            display: none;
+          }
+          .subcell_description {
+            display: none;
+          }
+          .inpicture:before{
+            content: '🔎';
```
aurel-l, line 111:
```suggestion
            content: '🔎 ';
```
Just to have a bit of breathing room after the icon

package.json
```diff
@@ -99,6 +99,7 @@
     "@geneontology/ribbon": "1.11.2",
     "@sentry/browser": "5.30.0",
     "@swissprot/rhea-reaction-visualizer": "0.0.18",
+    "@swissprot/swissbiopics-visualizer": "^0.0.22",
```
aurel-l, line 102:
```suggestion
    "@swissprot/swissbiopics-visualizer": "0.0.22",
```

package.json
```diff
@@ -136,6 +137,7 @@
     "redux": "4.0.5",
     "redux-thunk": "2.3.0",
     "timing-functions": "2.0.1",
+    "tippy.js": "^6.2.7",
```
aurel-l, line 140:
```suggestion
    "tippy.js": "6.2.7",
```

aurel-l (changes requested):
Would push to have those with exact versions, as the rest of our dependencies.

Also, it would be good when everyone's around to talk with the rest of the team about the choice of tippy.js, as a pop-up library might be useful throughout the website, so just to see if this one covers all of our potential use-cases and how it needs to be used since you went throught the work of integrating it for this PR.

aurel-l:
Added a commit with a webpack change to handle tippy.js's CSS

src/uniprotkb/components/protein-data-views/styles/tippy.css
```diff
@@ -0,0 +1,73 @@
+.tippy-box[data-animation='fade'][data-state='hidden'] {
```
aurel-l, line 1: whole file can be removed now I guess?

src/uniprotkb/components/protein-data-views/SubcellularLocationWithVizView.tsx
dlrice: I had a right margin set for the icon but is a space character preferable?

aurel-l: Ah, this wasn't visible somehow when I tested it, that's why I added a space. Will have another look on Monday

src/uniprotkb/components/protein-data-views/SubCellViz.tsx
```diff
@@ -57,9 +68,8 @@ const attachTooltips = (
     `#${locationGroup.id}term`,
   ].join(',');
   const locationTextElements = Array.from(
-    shadowRoot.querySelectorAll(locationTextSelector)
+    document.querySelectorAll(locationTextSelector)
```
aurel-l, line 71: Maybe query that on the instance of the element rather than the whole document, just in case we have multiple times this visualisation on the same page

dlrice:
> These 2 last will not do what they are supposed to as the addEventListener is on an arrow function, and not the instance methid directly.
> I think you can just not remove them for now, hopefully this will not cause too big of a memory leak.

I was afraid that was the case. I could possibly save all of the arrow functions in an array and the iterate through these to remove?

aurel-l:
> > These 2 last will not do what they are supposed to as the addEventListener is on an arrow function, and not the instance methid directly.
> > I think you can just not remove them for now, hopefully this will not cause too big of a memory leak.
> 
> I was afraid that was the case. I could possibly save all of the arrow functions in an array and the iterate through these to remove?

Nah, don't worry about it for now.
Maybe it's a non-issue if we manage to eventually have this library updated.

If you wanted to solve this, you'd need to add only 2 event listener on the node containing all the lines, and do event delegation to detect if the actual text element you are hovering has `.inpicture` and if so call highlight or removeHighlight.
Then you can easily keep a reference to the event listener functions are remove both of them in the return function of the useEffect

aurel-l (approved):
Approving. But please add the remaining TODO as Jiras in the backlog

dlrice:
Been trying to fix types. Can get all except for:
```
src/uniprotkb/components/protein-data-views/SubCellViz.tsx:142:17 - error TS2339: Property 'highLight' does not exist on type 'HTMLElement'.

142           super.highLight(event, target, selector);
                    ~~~~~~~~~

src/uniprotkb/components/protein-data-views/SubCellViz.tsx:152:17 - error TS2339: Property 'removeHiglight' does not exist on type 'HTMLElement'.

152           super.removeHiglight(event, target, selector);
                    ~~~~~~~~~~~~~~

src/uniprotkb/components/protein-data-views/SubCellViz.tsx:212:23 - error TS2339: Property 'highLight' does not exist on type 'Element'.

212             instance?.highLight(
                          ~~~~~~~~~

src/uniprotkb/components/protein-data-views/SubCellViz.tsx:219:23 - error TS2339: Property 'removeHiglight' does not exist on type 'Element'.

219             instance?.removeHiglight(
                          ~~~~~~~~~~~~~~
```
if you've got any ideas @aurel-l ?

aurel-l:
> Been trying to fix types. Can get all except for:
> 
> ```
> src/uniprotkb/components/protein-data-views/SubCellViz.tsx:142:17 - error TS2339: Property 'highLight' does not exist on type 'HTMLElement'.
> 
> 142           super.highLight(event, target, selector);
>                     ~~~~~~~~~
> 
> src/uniprotkb/components/protein-data-views/SubCellViz.tsx:152:17 - error TS2339: Property 'removeHiglight' does not exist on type 'HTMLElement'.
> 
> 152           super.removeHiglight(event, target, selector);
>                     ~~~~~~~~~~~~~~
> 
> src/uniprotkb/components/protein-data-views/SubCellViz.tsx:212:23 - error TS2339: Property 'highLight' does not exist on type 'Element'.
> 
> 212             instance?.highLight(
>                           ~~~~~~~~~
> 
> src/uniprotkb/components/protein-data-views/SubCellViz.tsx:219:23 - error TS2339: Property 'removeHiglight' does not exist on type 'Element'.
> 
> 219             instance?.removeHiglight(
>                           ~~~~~~~~~~~~~~
> ```
> 
> if you've got any ideas @aurel-l ?

I had a look at how Stencil.js was typing their components and try to reproduce that to type this custom elements. (We should do the same in the future when we need to type our own custom elements in the future, thinking about Nightingale).
If you're happy with this last commit, go ahead and merge. Otherwise happy to talk more about it and explore other approaches or for you to revert some of these changes.

A note about the variable names for the 2 public methods we're reusing here. I think I used `event`, `target`, and `selector` but the code I was reverse engineering the minified code was just listing `e`, `t`, and `s` so I came up with these names, but I believe `event` was not correct, I reversed to using `e` in the last commit because I couldn't come up with a better name.

PR #285, opened 2021-02-22
alternative approach

aurel-l:
Alternative approach:
 - tackles issue with style and click handlers within the list of subcellular locations
 - override or works around a lot of the logic implemented but the custom element in the library
 - still missing handling of hover on text and corresponding highlight in visualisation (should be doable though)


webpack.config.js
```diff
@@ -287,6 +287,11 @@ module.exports = (env, argv) => {
             name: 'geneontology',
             chunks: 'all',
           },
+          geneontology: {
```
dlrice, line 290:
```suggestion
          sibSubcell: {
```

dlrice (approved):

webpack.config.js
aurel-l: Good catch

PR #286, opened 2021-02-22
2021 02 proteomes model changes

dlrice:
## Purpose
Reflect proteomes API model changes in the front end codebase

## Approach

- Namespace: `proteome` → `proteomes`
- As directed by backend several proteomes attributes in the response are optional:
  - components
  - genomeAssembly
  - genomeAssemblyUrl
- Added level to GenomeAssembly type (which represents Genome representation)
- Rank enum added to taxonLineage and displayed in column renderer (more to be done for this and is covered in https://www.ebi.ac.uk/panda/jira/browse/TRM-25206)
- Converted mock data to TS (had to use a globally unique name for the file https://github.com/facebook/jest/issues/2070)
- Added top level `redundantTo` and `isolate` fields 
- `exclusionReasons` excluded from mock data for the reason that excluded proteomes are not indexed. Leo pretty sure of this but will check with Katie to confirm and I will update if it turns this is not the case.
- Label change: `Genome representation (RefSeq)` → `Genome representation`

## Testing
Update existing snapshots

## Checklist


src/proteomes/__mocks__/proteomesEntryModelData.ts
```diff
@@ -0,0 +1,125 @@
...
+  description: 'about some proteome',
+  taxonomy: {
+    scientificName: 'Homo sapiens',
+    commonName: 'Human',
+    synonyms: ['synonym'],
+    taxonId: 9606,
+    mnemonic: 'HUMAN',
+  },
+  modified: '2015-11-05',
+  proteomeType: ProteomeType.REDUNDANT,
```
aurel-l, line 19: In other places, in order to be able to not have to use these, especially for big payloads where they would be complicated to notice, I switched to using unions of strings instead of an enum.

aurel-l: To be fait, I found that using unions of strings didn't prevent TS to be able to help me find the possible values through tooling. If I'd start typing `proteomeType: '` VS code would come up with the list of possible values.

dlrice: Updated

aurel-l (changes requested):
Just some linting changes (but eslint is not being run on files not yet covered by TS)

aurel-l (approved):

PR #287, opened 2021-02-22
2021 02 publication model changes

xwatkins:
## Purpose
Adapt to model changes to the literature citations

## Approach
Update types, mocks and adapt code accordingly. Move Publication component to this repo.

## Testing
Moved tests from Franklin. Tests don't currently pass because it depends on an update to franklin first.

## Checklist


src/uniprotkb/components/entry/__tests__/__mocks__/entryPublicationsData.json
```diff
@@ -1,157 +1,1824 @@
...
-            "Masters C.L.",
-            "Grzeschik K.-H.",
-            "Multhaup G.",
-            "Beyreuther K.",
-            "Mueller-Hill B."
-          ],
-          "citationCrossReferences": [{
-              "database": "PubMed",
-              "id": "2881207"
+  "facets": [
```
aurel-l, line 2: This needs to be transformed to a .ts file, please comment the URL that was used to get the data in order to be able to re-fetch it later in case of another schema change

src/uniprotkb/adapters/uniProtkbConverter.ts
```diff
@@ -56,6 +57,9 @@ export const getEntryTypeFromString = (entryTypeString?: string) => {
   if (entryTypeString.match(/Swiss-Prot|reviewed|^sp\|$|^sp$/gi)) {
     return EntryType.REVIEWED;
   }
+  if (entryTypeString.match(/ORCID$/gi)) {
```
aurel-l, line 60: Not sure how I feel about this, this doesn't feel like it should be within the uniProtKB converter. And the EntryTypeIcon is starting to be a bit more than really an "Entry" type icon

src/uniprotkb/components/entry/__tests__/EntryPublications.spec.tsx
```diff
@@ -18,6 +18,6 @@ describe('EntryPublications tests', () => {
     useDataApi.mockImplementation(() => dataMock);
     const { findByText } = getRendered();
     expect(useDataApi).toHaveBeenCalled();
-    expect(await findByText(/ISOFORM APP751/)).toBeTruthy();
+    expect(await findByText(/S-adenosylhomocysteine/)).toBeTruthy();
```
aurel-l, line 21:
```suggestion
    expect(await screen.findByText(/S-adenosylhomocysteine/)).toBeInTheDocument()();
```

src/uniprotkb/adapters/uniProtkbConverter.ts
xwatkins: Initially thought it wasn't a good idea and then thought actually it could be that in the future we have "COMMUNITY_ANNOTATION" entries alongside SwissProt, TrEMBL etc..
However I think what is misleading is `EntryType.ORCID` so happy to change it to EntryType.COMMUNITY_ANNOTATION

aurel-l: I guess it can be better, yes. Still not a fan (for me an annotation is an annotation, about a sequence, and we might have community annotations on reviewed or unreviewed entries, that's why I'm feeling weird about that being here).

xwatkins: I think we can leave it for now - ideally we'd have something like an "icon resolver" and we can pass entry types for different namespaces, etc...

aurel-l (approved):

PR #290, opened 2021-02-25
2021 02 uniparc xref filter

aurel-l:
## Purpose
UniParc entry page xrefs filtering [TRM-25539](https://www.ebi.ac.uk/panda/jira/browse/TRM-25539)

## Approach
Download separately the sequence in one request, and the content of the xrefs table and facets in another.
Ask backend to filter the xrefs

## Testing
Updated tests, visual check on UniParc entries with selection of facets

## Checklist
- [x] I have checked that tests pass and coverage has at least improved, and if not explained the reasons why.

Note that this *does not* include the column selection


src/uniparc/adapters/uniParcConverter.ts
```diff
@@ -4,15 +4,14 @@ import { EntryType } from '../../uniprotkb/adapters/uniProtkbConverter';
...
-}
+export type XRefsInternalDatabases =
+  | 'UniProtKB/Swiss-Prot'
+  | 'UniProtKB/TrEMBL';
 
-export const databaseToEntryType: Record<string, EntryType | undefined> = {
-  [XRefsInternalDatabases.Reviewed]: EntryType.REVIEWED,
-  [XRefsInternalDatabases.Unreviewed]: EntryType.UNREVIEWED,
-};
+export const databaseToEntryType = new Map<string, EntryType>([
```
dlrice, line 11:
```suggestion
export const databaseToEntryType = new Map< XRefsInternalDatabases, EntryType>([
```

src/uniparc/components/entry/XRefsFacets.tsx
```diff
@@ -1,123 +1,155 @@
...
+  FacetObject,
+  FacetValue,
+} from '../../../uniprotkb/types/responseTypes';
 import { OrganismData } from '../../../uniprotkb/adapters/namesAndTaxonomyConverter';
+import { UseDataAPIWithStaleState } from '../../../shared/hooks/useDataApiWithStale';
+
+import '../../../shared/components/results/styles/results-view.scss';
 
-const sortByCount = (a: FacetValue, b: FacetValue) => b.count - a.count;
+const reviewed: XRefsInternalDatabases = 'UniProtKB/Swiss-Prot';
```
dlrice, line 24: Having to rewrite a string seems a downside of not using enums. I know this is way is preferable as it makes it easier to use mock values. Could there be a way to have both, for example:

```
export enum XRefsInternalDatabasesEnum {
  REVIEWED = 'UniProtKB/Swiss-Prot',
  UNREVIEWED = 'UniProtKB/TrEMBL',
}
export type XRefsInternalDatabases = `${XRefsInternalDatabasesEnum}`;
```
will allow us to only write out the string once in the enum definition and we can also export the union type for the API mocks. We could then for example do:

```
const sortByCount = (a: FacetValue, b: FacetValue) => {
  if (
    a.value in XRefsInternalDatabasesEnum &&
    b.value in XRefsInternalDatabasesEnum
  ) {
    return a.value === XRefsInternalDatabasesEnum.REVIEWED ? -1 : 1;
  }
  if (a.value in XRefsInternalDatabasesEnum) {
    return -1;
  }
  if (b.value in XRefsInternalDatabasesEnum) {
    return 1;
  }
  return b.count - a.count;
};
```

src/uniparc/components/entry/XRefsFacets.tsx
```diff
@@ -1,123 +1,155 @@
...
+const unreviewed: XRefsInternalDatabases = 'UniProtKB/TrEMBL';
+const exceptions = new Set<string>([reviewed, unreviewed]);
```
dlrice, line 26: Could be `new Set< XRefsInternalDatabases>` as this is what the set strictly consists of and then using `a.value as XRefsInternalDatabases` when checking membership

src/uniparc/components/entry/XRefsFacets.tsx
```diff
@@ -1,123 +1,155 @@
...
+    organisms,
+    (organism) => organism.taxonId
+  );
+  const taxonCounts = countBy<OrganismData>(
+    organisms,
+    (organism) => organism.taxonId
+  );
+  const taxonomyFacetValues = taxonObjects
+    .map((taxon): OrganismTuple => [taxon, taxonCounts[taxon.taxonId]])
+    .sort((tupleA, tupleB) => tupleB[1] - tupleA[1]);
```
dlrice, line 62: Stylistic suggestion:
```suggestion
    .sort(([, countA], [, countB]) => countB - countA);
```

src/uniparc/adapters/uniParcConverter.ts
aurel-l: Actually reverting and putting an union. It's because I'm not necessary doing `.get()` with one of these two options, but rather with any database name, so I need the map to accept any string, but then I have to take into account that the result of `get()` will be either an `EntryType`, or undefined (for any database which is not an `XRefsInternalDatabases`)

src/uniparc/components/entry/XRefsFacets.tsx
aurel-l: Same comment than for the map. I need this to accept generic strings, not just those 2. I changed with a union just to be clear what it contains

src/uniparc/components/entry/EntryMain.tsx
```diff
@@ -1,31 +1,28 @@
...
 
-function arePropsEqual(prevProps: EntryMainProps, nextProps: EntryMainProps) {
-  // Do NOT re-render the page, as long as the 'id' value is the same.
-  return (
-    prevProps.transformedData.uniParcId === nextProps.transformedData.uniParcId
-  );
-}
-
-const EntryMain: FC<EntryMainProps> = ({ transformedData, metadata }) => (
+const EntryMain = memo<EntryMainProps>(({ transformedData, xrefs }) => (
```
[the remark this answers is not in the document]
aurel-l: changed

src/uniparc/components/entry/XRefsFacets.tsx
aurel-l: I like this approach, I changed it partially like you suggested with the enum and the stringified version of it for the type. Not really sure about the `in` test with enums though so I did keep the Set

dlrice (approved):

PR #292, opened 2021-03-01
Add zoom tool

xwatkins:
## Purpose
Users were confused by the lack of zoom controls for ProtVista components. 

## Approach
This PR addresses this issue by providing zoom in, zoom out and zoom to sequence buttons.
Note: the styling of these buttons is not finalised, but this PR adds the required functionality.

## Testing
The actual functionality is part of Nightingale so will be tested there. Updated snapshots.

## Checklist


src/uniprotkb/components/protein-data-views/NightingaleZoomTool.tsx
```diff
@@ -0,0 +1,38 @@
...
+        /* webpackChunkName: "protvista-zoom-tool" */ 'protvista-zoom-tool'
+      ),
+    'protvista-zoom-tool'
+  );
+
+  return (
+    <div className="button-group">
+      <protvista-zoom-tool length={length}>
+        <span slot="zoom-in" className="nightingale-button-content">
+          <SearchIcon width={iconSize} />+
```
aurel-l, line 24: Not a fan of relying on overlapping text and SVG. This will change depending on user fonts, browsers, or even screens (mine show the "-" not centred, towards the left, and same thing for the "+" but towards the right)
The "+", "-", and "ATG" (or whatever is chosen for this) need to be SVG shapes within the same SVG.

src/uniprotkb/components/protein-data-views/FeaturesView.tsx
```diff
@@ -11,6 +11,7 @@ import FeatureType from '../../types/featureType';
 import { UniProtProtvistaEvidenceTag } from './UniProtKBEvidenceTag';
 import FeaturesTableView, { FeaturesTableCallback } from './FeaturesTableView';
 import { Xref } from '../../../shared/types/apiModel';
+import NightingaleZoomTool from './NightingaleZoomTool';
```
aurel-l, line 14: Is it possible to order imports in some kind of way? Not attached to how, but at least group things that make sense together.

I usually see:
 - external imports
 - components
 - hooks
 - utils
 - types
 - non-standard imports (.scss)
 
 But anything as long as when I open a file I can grasp what it depends on easily.
 See QueryBuilder.tsx or ResultsView.tsx for extreme examples.

aurel-l (changes requested):
Function-wise I think it's fine.
But it needs to have proper icons before being able to go in the main codebase.

src/uniprotkb/components/protein-data-views/FeaturesView.tsx
xwatkins: We should decide on a guideline for this together with @dlrice and document it here: https://github.com/ebi-uniprot/coding-guidelines

I'm happy with the one you're suggesting.

aurel-l: If @dlrice doesn't see any objection I'll add it to the guidelines.
This should be a loose rule though, because it depends a lot on how crowded the imports are, and sometimes you do have to import different types of imports from the same files so the groups get mixed up (mainly thinking of TS types that might be within any kind of import).

xwatkins: yep, loose rules (guidelines) are documented, while hard rules should be enforced by linting

xwatkins:
Updates done.

Also removed red border around error message component as specified here https://github.com/ebi-uniprot/franklin-sites/issues/172

aurel-l (approved):

PR #296, opened 2021-03-02
update the proteomes column renderer for UniParc

aurel-l:
## Purpose
Add missing component in the proteomes renderer of the UniParc results page (table view). Matches behaviour of the current website.

## Testing
Updated snapshot

## Checklist

### DIFF
--- a/src/uniparc/config/UniParcColumnConfiguration.tsx
+++ b/src/uniparc/config/UniParcColumnConfiguration.tsx
@@ -1,4 +1,6 @@
+import { Fragment } from 'react';
 import { Link } from 'react-router-dom';
+import { uniqBy } from 'lodash-es';
 import {
   ExpandableList,
   ExternalLink,
@@ -152,14 +154,24 @@ UniParcColumnConfiguration.set(UniParcColumn.protein, {
   ),
 });
 
+type XRefWithProteomeId = UniParcXRef &
+  Required<Pick<UniParcXRef, 'proteomeId'>>;
 UniParcColumnConfiguration.set(UniParcColumn.proteome, {
   label: 'Proteomes',
-  render: (data) => (
+  render: ({ uniParcCrossReferences }) => (
     <ExpandableList descriptionString="proteomes" displayNumberOfHiddenItems>
-      {xrefGetter(data, 'proteomeId')?.map((accession) => (
-        <Link key={accession} to={getEntryPath(Namespace.proteomes, accession)}>
-          {accession}
-        </Link>
+      {uniqBy(
+        uniParcCrossReferences?.filter((xref): xref is XRefWithProteomeId =>
+          Boolean(xref.proteomeId)
+        ) || [],
+        (xref) => `${xref.proteomeId}-${xref.component}`
+      ).map((xref) => (
+        <Fragment key={`${xref.proteomeId}-${xref.component}`}>
+          <Link to={getEntryPath(Namespace.proteomes, xref.proteomeId)}>
+            {xref.proteomeId}
+          </Link>
+          {xref.component ? `:${xref.component}` : undefined}
+        </Fragment>
       ))}
     </ExpandableList>
   ),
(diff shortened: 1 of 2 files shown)


src/uniparc/config/UniParcColumnConfiguration.tsx
```diff
@@ -152,14 +154,24 @@ UniParcColumnConfiguration.set(UniParcColumn.protein, {
...
+        uniParcCrossReferences?.filter((xref): xref is XRefWithProteomeId =>
+          Boolean(xref.proteomeId)
+        ) || [],
+        (xref) => `${xref.proteomeId}-${xref.component}`
+      ).map((xref) => (
+        <Fragment key={`${xref.proteomeId}-${xref.component}`}>
+          <Link to={getEntryPath(Namespace.proteomes, xref.proteomeId)}>
+            {xref.proteomeId}
+          </Link>
+          {xref.component ? `:${xref.component}` : undefined}
```
xwatkins, line 173: Formatting: would add a space after ':'

aurel-l: I'm matching the current behaviour of the website, that's why I didn't put a space.
How should a user read the information of this cell with the component information? I'm not sure what I'm seeing here, that's why I had followed the current website's implementation

src/uniparc/config/UniParcColumnConfiguration.tsx
```diff
@@ -182,13 +194,14 @@ UniParcColumnConfiguration.set(UniParcColumn.sequence, {
...
+  Required<Pick<UniParcXRef, 'database' | 'id'>>;
 UniParcColumnConfiguration.set(UniParcColumn.accession, {
   label: 'UniProtKB',
   render: ({ uniParcCrossReferences }) => (
     <ExpandableList descriptionString="entries" displayNumberOfHiddenItems>
       {uniParcCrossReferences
-        ?.filter((xref): xref is UniParcXRef &
-          Required<Pick<UniParcXRef, 'database' | 'id'>> =>
+        ?.filter((xref): xref is XRefWithDatabase =>
           Boolean(xref.database?.startsWith('UniProtKB') && xref.id)
```
xwatkins, line 205: Q about this syntax, as I don't think we've been using it so far - would it be good practice to add this to guidelines too? especially when dealing with `null`, `undefined` in conjunction with other statement tests

xwatkins (approved):

src/uniparc/config/UniParcColumnConfiguration.tsx
aurel-l: Which syntax specifically? TypeScript's ` is ` statement with `.filter()` in order to do some type guards?
We already use it in about 9 instances throughout the codebase, so even if not really common, it's not too new.
Not sure how to specify its use in the guidelines though, especially as I believe I'm not too used to using it and always doubt if I'm using it right, also, I'm always using it in conjunction with a `filter()`, so I'm not sure where else we can take advantage of it.

xwatkins: No, I meant using  `Boolean()` instead of directly testing `xref.database?.startsWith('UniProtKB') && xref.id`

aurel-l: Ah, it's just that xref.id is not a boolean, and TS wants the filter function to return a boolean (and not just truthy or falsy values)

src/uniparc/config/UniParcColumnConfiguration.tsx
xwatkins: For proteomes, "component" means either whole genome (bacterial genome), a chromosome (chromosome 21) or mitochondria (I think that's it). I think this is the list of components and associated proteomes, and they just used a semicolon as a separator and forgot to add a space.

src/uniparc/config/UniParcColumnConfiguration.tsx
xwatkins: Ah ok, it's enforced by TS then.

src/uniparc/config/UniParcColumnConfiguration.tsx
aurel-l: There's not much information on the website about those https://www.uniprot.org/help/proteome_component
I'm wondering in the first place why it was represented like this and not as `<Proteome ID> (<component>)` which looks more like a way to specify some extra information in plain English rather than the `:` without extra spaces which I read as some kind of accession specifier (hence why I was keeping the exact same syntax, just in case...)

aurel-l: Ah, sorry, crossed message.
I do see some "exotic" component names some times, not sure if that's normal
Screenshot 2021-03-02 at 17 15 29

xwatkins: I think the parenthesis would be nicer indeed.

xwatkins: > Ah, sorry, crossed messages.
> I do see some "exotic" component names some times, not sure if that's normal
> Screenshot 2021-03-02 at 17 15 29

Yep, plasmids are circular bits of DNA circulating in Bacterias (but I guess you know that already...) and the Unassembled are from shotgun sequencing (bits they couldn't match to anything)

PR #300, opened 2021-03-05
2021 03 protvista uniparc add component

xwatkins:
## Purpose
Add the protvista tab to the Uniparc entry to display entry features.

## Approach
Using the component from the refactor in the other PR. Convert the data (flatter structure than old website). Add a function to decide on colour based on database name.

## Testing
Will add tests on Monday - but coverage hasn't dropped.

## Checklist


src/uniparc/components/entry/Entry.tsx
```diff
@@ -78,7 +84,45 @@ const Entry: FC = () => {
...
+          <EntryMain transformedData={transformedData} xrefs={xrefDataObject} />
+        </Tab>
+        <Tab
+          title={
+            <Link
+              to={(location) => ({
+                ...location,
+                pathname: `/uniparc/${match.params.accession}/${TabLocation.FeatureViewer}`,
+                hash: undefined,
+              })}
```
dlrice, line 111: Would be good to use the `LocationToPath` config for uniparc and add a `subPage` as we do here:
```
  [Location.UniProtKBEntry]: `/${Namespace.uniprotkb}/:accession/:subPage?`,
```

src/uniparc/components/entry/UniParcFeaturesView.tsx
```diff
@@ -1,8 +1,93 @@
...
+    'interproGroupName'
+  );
+
+// DEfine columns
+const columnConfig = () => ({
+  interproGroup: {
+    label: 'InterPro Group',
+    resolver: (d: UniParcProcessedFeature) =>
+      html`<a
+        href="//www.ebi.ac.uk/interpro/entry/InterPro/${d.interproGroupId}"
```
dlrice, line 47: `InterProEntry` is available in `externalUrls`

src/uniparc/components/entry/UniParcFeaturesView.tsx
```diff
@@ -1,8 +1,93 @@
...
+  data: SequenceFeature[];
+  sequence: string;
+}> = ({ data, sequence }) => {
+  const processedData = useMemo(() => convertData(data), [data]);
+  return (
+    <FeaturesView
+      features={processedData}
+      columnConfig={columnConfig}
+      sequence={sequence}
+      // trackHeight={processedData.length * 10}
```
dlrice, line 89: Do we want to remove this for good?

src/shared/components/views/FeaturesTableView.tsx
```diff
@@ -10,12 +10,14 @@ import useCustomElement from '../../hooks/useCustomElement';
...
 import { ColumnConfig, ProcessedFeature } from './FeaturesView';
+import { UniParcProcessedFeature } from '../../../uniparc/components/entry/UniParcFeaturesView';
 
 export type FeatureColumns = {
   [name: string]: {
     label: string;
     resolver: (
-      d: ProcessedFeature & TransformedVariant
+      // This could be improved
+      d: ProcessedFeature & TransformedVariant & UniParcProcessedFeature
```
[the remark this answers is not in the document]
xwatkins: The issue with this approach is that I would have had to cast as `unknown`. Ended up making the improvement I wanted to make, which was to use generics.

src/shared/utils/color.ts
```diff
@@ -0,0 +1,13 @@
+export const stringToColour = (str: string) => {
```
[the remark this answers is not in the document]
xwatkins: Excellent point. Added a cap, and changed to `rgb` as it's easier to set.

src/uniparc/components/entry/Entry.tsx
xwatkins: Good catch. Just added, all comments should now be addressed.

dlrice (approved):

src/shared/components/views/FeaturesView.tsx
```diff
@@ -78,7 +81,7 @@ const FeaturesView: FC<FeatureProps> = ({
   const ceDefined = navigationDefined && sequenceDefined && managerDefined;
 
   const featureTypes = useMemo(
-    () => uniq(features.map(({ type }) => type.toLowerCase())),
+    () => uniq(features.map(({ type }) => (type as string).toLowerCase())),
```
aurel-l, line 84: Is that needed? TS wasn't complaining before

src/shared/components/views/FeaturesView.tsx
```diff
@@ -28,33 +28,36 @@ export type ProcessedFeature = {
...
-  columnConfig: ColumnConfig;
 };
 
-const FeaturesView: FC<FeatureProps> = ({
-  sequence,
-  features,
-  columnConfig,
-}): JSX.Element | null => {
+// ProcessedFeature | TransformedVariant | UniParcProcessedFeature
+const FeaturesView = <T extends Record<string, unknown>>(
```
aurel-l, line 57:
```suggestion
const FeaturesView = <T extends ProcessedFeature | TransformedVariant | UniParcProcessedFeature>(
```
I believe this might solve the below `as string` comment.

xwatkins: 🤦 of course

src/shared/utils/color.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Convert a random string to an rgb colour
+ * @param str the string to convert
+ * @param cap an optional maximum value for r,g and b
+ * @returns rgb values as a string
+ */
+export const stringToColour = (str: string, cap = 255) => {
```
aurel-l, line 7: Not sure about generating a completely random `rgb` value.
I would ideally prefer to generate something in the `lch` colour space, but support is lower I believe, but `hsl` is widely supported.
Better logic would be to hash the string to generate a hue (0-360) and stick to either fixed saturation and lightness values, or also use the hashing to generate a range within "nice" values for those (40-60% range?)

aurel-l:
Are we expecting the facets to affect the content of the visualisation? I'm think that some domain might disappear if certain facets are selected. Not sure if that's useful as it's the same actual sequence, so InterPro entries should be the same anyway.

If we don't expect the facets to affect the visualisation, then the facets should be hidden? Using the grey empty sidebar?

aurel-l:
Also, not sure why, but this breaks the result pages for supporting data searches. It only does the query for the facets, but not for the actual content of the page

xwatkins:
> Also, not sure why, but this breaks the result pages for supporting data searches. It only does the query for the facets, but not for the actual content of the page

Hadn't merged in the sprint branch, seems to work now

src/shared/utils/color.ts
xwatkins: I think the current implementation is ok for now, but will create a Jira with your suggestions. There are vague plans to unify all feature colours at some point in the future and use luminance or another property to differentiate them.

aurel-l:
OK, and regarding the side bar: User path of UniParc entry page > select facet > feature viewer tab. Is it OK to have the facet in the URL?
Good point: if you click back on the entry page, you end up on the same xref selection.
Negative point: users might expect it to have an effect?

xwatkins:
> OK, and regarding the side bar: User path of UniParc entry page > select facet > feature viewer tab. Is it OK to have the facet in the URL?
> Good point: if you click back on the entry page, you end up on the same xref selection.
> Negative point: users might expect it to have an effect?

I think the visual change might be enough. Might be worth checking with UX - maybe include in testing under "what would you expect to see if you click back on the entry? What do you think the visualisation is showing?

aurel-l:
> > OK, and regarding the side bar: User path of UniParc entry page > select facet > feature viewer tab. Is it OK to have the facet in the URL?
> > Good point: if you click back on the entry page, you end up on the same xref selection.
> > Negative point: users might expect it to have an effect?
> 
> I think the visual change might be enough. Might be worth checking with UX - maybe include in testing under "what would you expect to see if you click back on the entry? What do you think the visualisation is showing?

👍🏽 

I believe this is good like this now, and especially I checked in the BLAST result pages and it's the same behaviour there too at the moment, so at least it's consistent.

aurel-l (approved):

PR #301, opened 2021-03-08
2021 03 footer

aurel-l:
## Purpose
Website footer [TRM-25552](https://www.ebi.ac.uk/panda/jira/browse/TRM-25552)

## Approach
Refactor layout in order to fit the footer in all the pages.
Add footer in a grid that rearranges itself around 3 different breakpoints

## Testing
Snapshot

## Checklist

 - In the end didn't implement a specific small footer for sidebar layout pages as I managed to implement it in a common way for all the pages and it would have been ore complicated to do so. Probably need feedback from UX to see if it's still necessary.
 - I've also hidden the scrolling bar in the facets scrolling area, but would probably need to check if it's OK or not to hide that.
 - Some links raise linting issue, it's normal until we actually handle these (other specific jira for those)
 - Also, included other changes after reviewing Lighthouse audits on the homepage
 - Images have been slightly optimised, but a better approach using webpack loaders might need to be defined at a later point when we have a better overview of how we will use images in the rest of the website.
 - Even if outside of the scope of the jira, implemented tablet view (completely), and _a_ mobile view (but that is just moving things around, not following the mockup that is hiding stuff). Acceptance is based on desktop view only.


src/shared/components/layouts/styles/side-bar-layout.scss
```diff
@@ -37,10 +33,21 @@
 
   &__sidebar {
     @include scroll-shadow;
```
xwatkins, line 35: do we want to add that to `&__content` too?

xwatkins (commented):
Not sure why, but the EMBL-EBI png doesn't look as crisp as the others?

image

aurel-l:
> Not sure why, but the EMBL-EBI png doesn't look as crisp as the others?
> 
> image

Related to
> Images have been slightly optimised, but a better approach using webpack loaders might need to be defined at a later point when we have a better overview of how we will use images in the rest of the website.

For this specific image I ended up using a SVG, even if bigger in size, to make sure to have crisp image, but we might have this issue for any image on the page depending on the size at which it is rendered and the pixel density of the screen it is rendered on.

src/shared/components/layouts/styles/side-bar-layout.scss
aurel-l: Can't really do that as we have 3 different sticky elements (page header, action buttons, table header) that would hide the shadow anyway.

src/shared/components/layouts/styles/footer.scss
```diff
@@ -0,0 +1,165 @@
...
+      white-space: nowrap;
+    }
+
+    img {
+      height: 2.25em;
+    }
+  }
+
+  // tablet breakpoint TODO: define breakpoint (copy from foundation now)
+  @media only screen and (max-width: 64em) {
```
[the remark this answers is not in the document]
aurel-l: I hadn't seen it before, so I changed it to use Franklin's now.
Had to move rules around as the logic was the opposite of what I had. In Franklin it is mobile-first and override incrementally for bigger sizes

src/shared/components/layouts/styles/side-bar-layout.scss
xwatkins: Maybe I was referring to the wrong thing. What is the bit which adds the shadow at the bottom of the sidebar when the footer is in view?

aurel-l: This only adds a shadow to any scrollable area. The scrollable area here is the whole page.

xwatkins (approved):

PR #302, opened 2021-03-08
2021 03 proteomes entry overview

dlrice:
## Purpose
- Create overview section of proteomes entry. 
- [Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-25400)
- Additional sections to follow in subsequent PRs.

## Approach
- Created Proteomes entry with overview section
- Created a few Proteomes components that are shared between the results and the entry page
- Moved entry type icon related code to /shared
- Created `isSameEntry` util fn for entry memoization along with `RequireAtLeastOne` util type.

## Testing
Snapshots

## Checklist


src/proteomes/config/ProteomesEntryConfig.tsx
```diff
@@ -0,0 +1,23 @@
+import Overview from '../components/entry/Overview';
+import { ProteomesUIModel } from '../adapters/proteomesConverter';
+import EntrySection from '../types/entrySection';
+
+const UniRefEntryConfig: {
```
dlrice, line 5: I just noticed this copy-paste relic and it will be fixed in a future PR (Components table).

src/proteomes/components/BuscoAbbr.tsx
```diff
@@ -0,0 +1,3 @@
+export default () => (
```
xwatkins, line 1: As we will have `abbr` for a lot more field names, it might be worth making a map somewhere in shared utils?

src/proteomes/components/GenomeAssemblyAndAnnotationView.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC } from 'react';
```
xwatkins, line 1: I think you could use the `SimpleView` component instead of making a new one.

src/proteomes/components/entry/Overview.tsx
```diff
@@ -0,0 +1,130 @@
...
+              ),
+            },
+            {
+              title: 'Proteome ID',
+              content: data.id,
+            },
+            (data.taxonomy.taxonId || data.taxonomy.scientificName) && {
+              title: 'Taxonomy',
+              content: (
+                <Link
```
xwatkins, line 82: I know it doesn't quite match the mockup, but could we re-use `OrganismDataView` here?

src/proteomes/components/entry/Overview.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  </div>
+                  <BuscoView
+                    report={data.proteomeCompletenessReport?.buscoReport}
+                  />
+                </div>
+              ),
+            },
+          ].filter(Boolean) as InfoData
+        }
+      />
```
xwatkins, line 126: There should also be a description as part of the overview for some proteomes. Is that missing from the data payload?

src/shared/types/utils.ts
```diff
@@ -0,0 +1,7 @@
+export type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<
```
xwatkins, line 1: I wonder if there's a simpler way to achieve the same thing? Is the purpose to enforce the presence of one extra key not in the list of passed keys? I think the name might be a bit misleading too.

Tagging @aurel-l

xwatkins (changes requested):

src/shared/types/utils.ts
aurel-l: Looks like this came from https://docs.microsoft.com/en-us/javascript/api/@azure/keyvault-certificates/requireatleastone?view=azure-node-latest I would trust them on that.
I don't understand what is going on inside either 😅 
Maybe comment where this is from so that we are not scared next time someone stumbles upon this file.

dlrice: My intention here is to make sure that at least one of `primaryAccession` or `id` is provided to the function (ie they can't both be optional because then they could both legally be absent).
Found it on [SO](https://stackoverflow.com/questions/40510611/typescript-interface-require-one-of-two-properties-to-exist) but found it as part of the [type-fest](https://www.npmjs.com/package/type-fest) package which is pretty popular with 53M weekly downloads. Happy to install and use from here rather than having it in our codebase. Thoughts?

src/proteomes/components/BuscoAbbr.tsx
dlrice: Good idea - updated.

src/proteomes/components/GenomeAssemblyAndAnnotationView.tsx
dlrice: I agree - updated. Also moved to the shared as it was being used outside of uniprotkb in a few places.

src/proteomes/components/entry/Overview.tsx
dlrice: Done

src/proteomes/components/entry/Overview.tsx
dlrice: Good catch. This was just an oversight on my part - updated now

src/shared/types/utils.ts
aurel-l: Sounds good 👍🏽 And I see that in this library there are other patterns that might be useful for us.

src/proteomes/components/entry/Overview.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <div className="description">
+        <hr />
```
xwatkins, line 119: We might not want to display this if no description is available?

dlrice: Updated

xwatkins (approved):

PR #306, opened 2021-03-11
2021 03 citations card

xwatkins:
## Purpose
Display citations supporting data results as Cards.

## Approach
- consolidate type definition between KB Entry citations and Literature citations
- re-use same component to display literature citations everywhere
- implement a `CitationCard` component as wrapper

## Testing
Testing `CitationCard`.

The coverage has dropped, likely because I removed some redundant code. I checked the coverage in all components involved in the PR and they seem to all be well tested.

## Checklist


src/supporting-data/citations/adapters/citationsConverter.ts
```diff
@@ -1,4 +1,4 @@
-export type Statistics = {
+export type LiteratureStatistics = {
```
aurel-l, line 1:
```suggestion
export type CitationsStatistics = {
```

Just to keep consistency with name of endpoint

src/supporting-data/citations/adapters/citationsConverter.ts
```diff
@@ -41,15 +41,65 @@ export type Citation = {
...
 export type CitationsUIModel = CitationsAPIModel & {
   // any addition/change by the converter
 };
 
+export const getCitationPubMedId = (citation: Citation) =>
+  citation.citationCrossReferences &&
+  citation.citationCrossReferences.find((xref) => xref.database === 'PubMed');
+
+export const getDoiXref = (citation: Citation) =>
+  citation.citationCrossReferences &&
```
aurel-l, line 81:
```suggestion
  citation.citationCrossReferences?.find((xref) => xref.database === 'DOI')
```

src/supporting-data/citations/adapters/citationsConverter.ts
```diff
@@ -41,15 +41,65 @@ export type Citation = {
...
   citation: Citation;
+  references?: Reference[];
 };
 
 export type CitationsUIModel = CitationsAPIModel & {
   // any addition/change by the converter
 };
 
+export const getCitationPubMedId = (citation: Citation) =>
+  citation.citationCrossReferences &&
```
aurel-l, line 77:
```suggestion
  citation.citationCrossReferences?.find((xref) => xref.database === 'PubMed')
```

src/supporting-data/citations/adapters/citationsConverter.ts
```diff
@@ -41,15 +41,65 @@ export type Citation = {
...
+
+  const pubmedId = pubMedXref && pubMedXref.id;
+
+  const journalInfo = {
+    journal: citation.journal,
+    volume: citation.volume,
+    firstPage: citation.firstPage,
+    lastPage: citation.lastPage,
+    publicationDate: citation.publicationDate,
+    doiId: doiXref && doiXref.id,
```
aurel-l, line 98:
```suggestion
    doiId: doiXref?.id,
```

src/supporting-data/citations/components/results/CitationCard.tsx
```diff
@@ -0,0 +1,46 @@
...
+import LiteratureCitation from '../LiteratureCitation';
+
+const CitationCard: FC<{
+  data: CitationsAPIModel;
+  selected?: boolean;
+  handleEntrySelection?: (rowId: string) => void;
+}> = ({ data, selected, handleEntrySelection }) => {
+  const history = useHistory();
+
+  const key = getIdKeyFor(Namespace.citations)(data);
```
aurel-l, line 18: If you're going to use the curried version, do `const getIdKey = getIdKeyFor(Namespace.citations)` outside this component, and do `const key = getIdKey(data);`

src/supporting-data/citations/components/results/__tests__/CitationCard.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const { history } = renderWithRouter(
+      <CitationCard data={citationData} handleEntrySelection={handleClick} />
+    );
+    fireEvent.click(screen.getByRole('checkbox'));
+    expect(handleClick).toHaveBeenCalled();
+    fireEvent.click(screen.getAllByRole('button')[0]);
+    const {
+      location: { pathname },
+    } = history;
+    expect(pathname).toMatch('/citations/14702039');
```
aurel-l, line 27:
```suggestion
    expect(history.location.pathname).toMatch('/citations/14702039');
```
I think the non-destructured version improved readability

src/uniprotkb/components/entry/EntryPublications.tsx
```diff
@@ -137,7 +137,7 @@ const EntryPublications: FC<{ accession: string }> = ({ accession }) => {
   return (
     <section>
       <h2>Publications for {accession}</h2>
-      <DataListWithLoader<LiteratureResultsAPI>
+      <DataListWithLoader<CitationsAPIModel>
```
aurel-l, line 140:
```suggestion
      <DataListWithLoader
```
I know it was there before, but I think in this case this might not be needed as `allResults` used in the `data` prop is already strongly defined.

src/uniprotkb/components/protein-data-views/UniProtKBEntryPublications.tsx
```diff
@@ -35,21 +39,14 @@ const UniProtKBEntryPublications: FC<{
...
-            <LiteratureCitation
-              title={citation.title}
-              authors={citation.authors}
-              key={`${citation.title}-${citation.citationType}-${citation.journal}`}
-              pubmedId={pubmedId}
-              statistics={statistics}
-              journalInfo={journalInfo}
-            />
-          ))}
+        results.map((citationData) => (
```
aurel-l, line 42:
```suggestion
        results?.map((citationData) => (
```

src/uniprotkb/components/protein-data-views/__tests__/UniProtKBEntryPublications.spec.tsx
```diff
@@ -19,11 +21,16 @@ describe('UniProtKBEntryPublications', () => {
         <UniProtKBEntryPublications pubmedIds={pubMedIds} />
       </Router>
     );
-    expect(useDataApi).toHaveBeenCalledWith(getPublicationsURL(pubMedIds));
+    expect(useDataApi).toHaveBeenCalledWith(
```
aurel-l, line 24: Do we need to check that? We're mocking its implementation a few lines above, if we weren't calling it we would get an empty component, and that would affect the snapshot test in the next line.
At most do a check to see if a UI component depending on the mocked data is in the DOM or not (get a link with a pubmed id in the mocked data and check if it's in the document?)

src/supporting-data/citations/components/LiteratureCitation.tsx
```diff
@@ -193,104 +197,71 @@ const Statistics: FC<StatisticsProps> = ({ statistics, pubmedId }) => {
...
-                  <a
-                    href={`//www.ncbi.nlm.nih.gov/pubmed/${pubmedId}`}
-                    target="_blank"
-                    rel="noopener noreferrer"
-                  >
-                    PubMed
-                  </a>
-                </li>
+  return (
+    <section className="publication">
```
aurel-l, line 209: Maybe let's get rid of all these sections. At most I would keep the topmost one. Also, the topmost section I could argue that using `<article>` might make more sense than a generic section

aurel-l (changes requested):
See single comments scattered here and there

src/supporting-data/citations/adapters/citationsConverter.ts
xwatkins: Had to drop coverage after updating this, I guess because things that shouldn't have been rendered in the tests were.

aurel-l: You have conflicting files with the sprint branch regarding 2 files that are likely to use this converter, maybe resolving this conflict first before adjusting the coverage would be good, as it's likely that you would need additional tweaking of the coverage value anyway afterwards

xwatkins: Yes, addressing comments first and will finish with a merge into this branch so will see then

xwatkins: in the end it went back up. All should be addressed now.

aurel-l: The previous line is not needed anymore

xwatkins: Good catch - I don't think I like using the "commit suggested" that much and prefer to implement the change but maybe that's just me 😄

src/supporting-data/citations/components/results/CitationCard.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const handleCardClick = useCallback(() => {
```
aurel-l, line 20: See implementation inb `UniProtKBCard` to avoid handling clicks on links, buttons, or inputs

aurel-l:
Now testing the actual look and behaviour on the page.
I've raised a specific inline comment about clicking on specific areas of the page.
And visually I'm not sure about the `view abstract` bit. In the rest of the website, the user expect the abstract to be displayed in-place when clicking on it.
If we want it to take you to the entry page in this case, we should name it differently.
Also, maybe some vertical alignment would be good here, or a way to have the card take less vertical space, because there's plenty of unused white space
Screenshot 2021-03-17 at 10 02 30

src/supporting-data/citations/adapters/citationsConverter.ts
aurel-l: Yeah, you don't need to "accept" the suggestion, I'm mainly using it to show the diff, but that only applies to one line. There's often context to it, like in this case, or when using a different import for example

xwatkins:
> Now testing the actual look and behaviour on the page.
> I've raised a specific inline comment about clicking on specific areas of the page.
> And visually I'm not sure about the `view abstract` bit. In the rest of the website, the user expect the abstract to be displayed in-place when clicking on it.
> If we want it to take you to the entry page in this case, we should name it differently.
> Also, maybe some vertical alignment would be good here, or a way to have the card take less vertical space, because there's plenty of unused white space
> Screenshot 2021-03-17 at 10 02 30

This is one of the topics I'd like to discuss in the design meeting this morning. If all the data can be displayed in the results cards for supporting data, why even have a supporting data entry page? I've got an idea (which falls on the back of your old idea) and which could work.

aurel-l:
OK, we'll talk about that in the design meeting then.
A few things to raise is that when we start applying the grid layout with a maximum width, the card will not be that wide, the title will likely wrap (depending on length) and it won't look as stretched with that much white space.
Also, I'm not pushing to actually get the abstract in there, I think it's not its place.
Maybe, in the case of being within a card removing the "cited in" section from the right and putting it instead of the "view abstract" would balance things better (but that does like the pattern with the links at the bottom of the card)

src/supporting-data/citations/components/results/CitationCard.tsx
aurel-l: See the onClick on the checkbox below, which is not needed anymore

xwatkins: Actually wondering...should this logic be in the Card component itself?

aurel-l: I thing we need eventually need a wrapper ResultCard component within the uniprot-website codebase that would abstract the checkbox, the selectedEntry logic, and this.
Maybe a Franklin Card can stick to being a UI component.

xwatkins: Being clickable is part of the definition of the `Card` component, and we don't know whether the consumer of `Card` will have other clickable elements within it, so I'm erring more towards having this logic in Card (maybe with a prop to overwrite the set of `BLOCK_CLICK_ON_CARD`). But not for now anyway

aurel-l: There's more to it though, I commented to Dan already that if we have list with checkboxes, we also need to eventually handle them in context as a user would expect them to behave. That is, click checkbox, then ctrl-click another checkbox, then all the checkboxes in between need to be clicked. This also needs to be handled somewhere.

aurel-l (approved):

src/uniprotkb/components/protein-data-views/UniProtKBEntryPublications.tsx
```diff
@@ -35,21 +39,14 @@ const UniProtKBEntryPublications: FC<{
...
-            <LiteratureCitation
-              title={citation.title}
-              authors={citation.authors}
-              key={`${citation.title}-${citation.citationType}-${citation.journal}`}
-              pubmedId={pubmedId}
-              statistics={statistics}
-              journalInfo={journalInfo}
-            />
-          ))}
+        results?.map((citationData) => (
```
aurel-l, line 42: Forgot to check that one again, the previous line isn't needed anymore either. You can remove from the spring branch directly now I guess

xwatkins: Argh sorry. Doing this now

PR #307, opened 2021-03-12
apply linting and type checking to test files

aurel-l:
## Purpose
Lint and type check test files

## Approach
Modify config in order to enforce linting on *all* test files, and type checking on *some* test files. Files currently ignored by type checks are:
 - "src/query-builder/**/*.spec.ts*",
 - "src/shared/**/*.spec.ts*",
 - "src/tools/**/*.spec.ts*",
 - "src/uniparc/**/*.spec.ts*",
 - "src/uniprotkb/**/*.spec.ts*",
 - "src/uniref/**/*.spec.ts*"

Aim is to include them progressively.

Applied best practices from [https://kentcdodds.com/blog/common-mistakes-with-react-testing-library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) on some of the files I've touched

## Testing
Check that tests kept running, fix tests that were wrong (and I discovered they were wrong because of this refactoring)

## Checklist


src/uniprotkb/components/protein-data-views/__tests__/__snapshots__/XrefView.spec.tsx.snap
```diff
@@ -1,1134 +1,6 @@
...
-          </ul>
-        </div>
-      </div>
-    </li>
-  </ul>
-</DocumentFragment>
-`;
-
-exports[`XRefView should render for SubCellularLocation section 1`] = `
+exports[`XRefView should render section 1`] = `
```
xwatkins, line 3: I remember you asking about this test, just want to check that the drop in lines in the snapshot is expected

xwatkins (commented):
Looks good, I think we should meet to discuss best practices and document it in the coding guidelines (tagging @dlrice ). Some I've noticed in your PR:
- `jest.fn()` instead of `()=>{}`
- `screen` instead of find*/get* from `render()`
- hierarchy of react-testing-library selectors
- `test.each()`
but there must be more to discuss in https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

src/uniprotkb/components/protein-data-views/__tests__/__snapshots__/XrefView.spec.tsx.snap
aurel-l: Yep, no drop in coverage as it was testing the same thing in a loop. All the removed snapshots are identical

aurel-l:
To be fair, about these 2 it's because of eslint that I changed it
>     * `jest.fn()` instead of `()=>{}`
but I could have replaced with `() => {/* */}` I think and eslint would have been happy, but I guess this way it shows more that we might want to eventually test the calls to this function too

>     * test.each()
eslint doesn't like declaring functions (in this case out tests) in loops, so I changed to this. Also, give the potential to have concurrently running tests (but not sure if it's needed)

The rest is from that article of good practices.

xwatkins (approved):
Once conflicts are resolved

src/uniprotkb/components/protein-data-views/__tests__/__snapshots__/XrefView.spec.tsx.snap
dlrice: > Looks good, I think we should meet to discuss best practices and document it in the coding guidelines (tagging @dlrice ). Some I've noticed in your PR:
> 
> * `jest.fn()` instead of `()=>{}`
> * `screen` instead of find*/get* from `render()`
> * hierarchy of react-testing-library selectors
> * `test.each()`
>   but there must be more to discuss in https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

All looks good to me

PR #308, opened 2021-03-15
2021 03 use item select

dlrice:
## Purpose
The same item selection logic being used in several places.

## Approach
* Created a `useItemSelect` hook.
* Replaced `handleSelectedEntries` --> `handleItemSelection` as we are only really handling one item at a time.

## Testing
Wrote useItemSelect tests.

## Checklist

### DIFF
--- a/src/shared/components/results/ResultsContainer.tsx
+++ b/src/shared/components/results/ResultsContainer.tsx
@@ -1,4 +1,4 @@
-import { FC, useMemo, useState } from 'react';
+import { FC, useMemo } from 'react';
 import { useHistory, useLocation } from 'react-router-dom';
 import { PageIntro, Loader } from 'franklin-sites';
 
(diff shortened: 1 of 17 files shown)


aurel-l:
Use of a hook for this looks good.
1 comment, but not really blocking:

> we are only really handling one item at a time

That is something I hope we change in the future. I want the usual pattern you get with checkboxes of selecting the first one, then ctrl-select the 10th, and you get all the checkboxes in between selected. Not sure how to implement it yet, but it's really something I wished we had.

dlrice:
> Use of a hook for this looks good.
> 1 comment, but not really blocking:
> 
> > we are only really handling one item at a time
> 
> That is something I hope we change in the future. I want the usual pattern you get with checkboxes of selecting the first one, then ctrl-select the 10th, and you get all the checkboxes in between selected. Not sure how to implement it yet, but it's really something I wished we had.

Happy if I put: `TODO: handle multiple simultaneous item selections` in the hook for now? And then once implemented we can update the `set` name appropriately.

aurel-l (approved):

PR #309, opened 2021-03-15
2021_03-similar_proteins_nodata

xwatkins:
## Purpose
For some UniProtKB entries (e.g. A0A009EBH2) an error appeared in the Similar Proteins section.

## Approach
When displaying similar proteins, we query for UniRef members of clusters the protein belongs to. We filter out UniParc entries and isoforms, and this sometimes results in an empty list. Because of the groupBy, this translated into an empty object, so the check for empty data before rendering the component was always failing. I've added an early return to fix this issue.

As we are not able to know whether data will be available for similar proteins until the component is created, this causes an issue with the in-page nav. As a workaround it will always display and I've added a message in the component when no data is available.

## Testing
N.A

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
+++ b/src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
@@ -55,6 +55,10 @@ const SimilarProteins: FC<{
         }))
         .filter((item) => item.members.length > 1);
 
+      if (filtered.length <= 0) {
+        return null;
+      }
+
       const clusterTypeGroups: Partial<
         Record<UniRefEntryType, UniRefLiteAPIModel[]>
       > = groupBy(filtered, (cluster) => cluster.entryType);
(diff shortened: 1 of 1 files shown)


src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -82,73 +86,75 @@ const SimilarProteins: FC<{
...
+                      }}
+                    >
+                      View all
+                    </Button>
+                  </Tab>
+                )
+            )}
+          </Tabs>
+        ) : (
+          <em>No similar proteins found.</em>
```
aurel-l, line 156: Not sure about the wording. If we arrived there, it's when we have UniParc, but no UniProt entry, right? I would specify this, something like "No similat UniProt entry found" or something like this.
I believe we should be displaying the UniParc entries too anyway, marked clearly as UniParc, but that's different from the current behaviour of the website though.

xwatkins: I agree that in the long run we should also include UniParc. Will update the text to "No similar UniProtKB entry found" to make less ambiguous.

aurel-l (approved):

PR #312, opened 2021-03-16
2021 03 refactor user settings

aurel-l:
## Purpose
Change approach to user preference storage.

## Approach
Retire `useLocalStorage` custom hook, and replace with a new hook `useUserPreference` which is linked to a React Context.

## Testing
TODO, not passing yet, I know

## Checklist
  - slightly lower branch coverage after moving things around

There were some limitations before:
 - 2 usages of the same key would be out of sync (2 usages within the same page, or across different tabs or windows)
 - There was a delay on load where we would get no value, then the stored value

This tackles the above issues, and adds some improvements:
 - Synchronously get the stored value if any
 - Syncs custom hooks using the same key, even across tabs and windows
 - Still saves the data asynchronously, even after the component is unmounted quickly
 - Lazily load the user setting from local storage, and only once per navigation
 - Using Context avoids props drilling through the complex result components.
 - In this PR, I switched to having the column selection being visible immediately instead of having to save
 - Also, we should probably use that for the basket logic as it syncs well across tabs and windows.

Now the source of truth is a global user preferences state shared through a React Context, lazily initialised from localStorage. Slices of this global state are available through a custom hook (same return values as `useState()`). Modification to this global state are asynchronously saved to localStorage.


src/shared/contexts/UserPreferences.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const stateAndSetState = useSafeState<UserPreferences>({});
+
+  return (
+    <UserPreferencesContext.Provider value={stateAndSetState}>
+      {children}
+    </UserPreferencesContext.Provider>
+  );
+};
+
+export type PossibleUserPreferenceKey =
```
dlrice, line 45: I think `UserPreferenceKey` would be sufficient. Also thinking this might be better situated somewhere in src/app as we currently do for the Locations.

src/shared/components/customise-table/CustomiseTable.tsx
```diff
@@ -40,12 +35,9 @@ const CustomiseTable: FC<CustomiseTableProps> = ({
...
       data-testid="customise-table-form"
     >
-      <ColumnSelect onChange={handleChange} selectedColumns={selectedColumns} />
+      <ColumnSelect onChange={setColumns} selectedColumns={columns} />
       <div className="button-group sticky-bottom-right sliding-panel__button-row">
-        <Button variant="secondary" type="button" onClick={handleCancel}>
-          Cancel
-        </Button>
-        <Button type="submit">Save</Button>
+        <Button type="submit">Close</Button>
```
dlrice, line 40: It's neat to see the real-time updates but have a few reservations:
1. Will this create too many unnecessary requests?
2. Will users miss the ability to revert to what they had before (not just to default)? I guess user testing will only be able to answer this.

dlrice (commented):
Looks good. In the same way you have used contexts do we want to one day replace redux with this and save 25Kb of bundle?

src/shared/components/customise-table/CustomiseTable.tsx
aurel-l: Yeah, I wasn't sure about this behaviour either, hence the draft. I thought it could make sense now that we see both the table and the sidepanel at the same time, but the 2 points you raised are tradeoffs indeed

src/shared/contexts/UserPreferences.tsx
aurel-l: Well, I'm exporting it, but there's actually no need to export it so it just needs to be defined within this file. You get autocomplete wherever you use the hook.
I will still change the name though

aurel-l:
> Looks good. In the same way you have used contexts do we want to one day replace redux with this and save 25Kb of bundle?

Yes, eventually we could generalise the use of context instead of redux. It is what Redux uses under the hood anyway. But there's no rush

PR #313, opened 2021-03-16
Add Release info to header

xwatkins:
Depends on https://github.com/ebi-uniprot/franklin-sites/pull/176

## Purpose
Add the release info in the header and secondary nav

## Approach
Move the Release info to its own component and use the Header Franklin component

## Testing
Update snapshots

## Checklist


src/shared/components/layouts/ReleaseInfo.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const { headers } = useDataApi(
+    `${apiUrls.search(Namespace.uniprotkb)}?query=*&size=0`,
+    fetchOptions
+  );
+  const releaseDate = headers?.['x-release']
+    ? new Date(headers['x-release'])
+    : undefined;
+
+  if (!releaseDate) {
+    return <>Loading release information</>;
```
aurel-l, line 19: It would be good to have the link to Statistics be visible even without the release information yet as this is not network dependent.
Also, I understand this component has been moved here to reuse it, but while I think it is OK to display "Loading release information" in the footer, I believe it might be a bit disruptive to the user to see this happen in the header.
I would rather reserve the space, and slowly make it appear when we have the information (transition opacity)

xwatkins: Yep, it doesn't look great at the moment... I was wondering about using a similar trick as your cursor for "Find my ... " but didn't get a chance to look into it.
Also, how would you go about reserving the space? swap an empty span with as width the expected number of `ch`?

aurel-l: Do you have a screenshot of how it looks like at the moment?

src/shared/components/layouts/styles/release-info.scss
```diff
@@ -0,0 +1,10 @@
+.release-info {
+  &__release_number {
+    transition: filter linear 0.25s;
+    filter: blur(0);
+  }
+
+  &__placeholder {
+    filter: blur(0.1rem);
```
aurel-l, line 8:
```suggestion
    will-change: filter;
    filter: blur(0.1rem);
```
As filter is not necessarily hardware accelerated (even less so a blur which is one of the most complex available filter), I would signal to the browser that this will be animated with this `will-change` so that it has a chance to provide it with its own rendering layer. Not sure if the fact that blur effect bleeds outside of the element and needs information about the background would negate the improvements given by that, but still.

src/uniparc/components/entry/Entry.tsx
```diff
@@ -49,7 +49,7 @@ const Entry: FC = () => {
 
   // if URL doesn't finish with "entry" redirect to /entry by default
   useEffect(() => {
-    if (match && !match.params.subPage) {
+    if (match && !match.params.subPage && accession) {
```
aurel-l, line 52: This is not really part of the PR, so more of a generic comment: Should we have this redirect logic on all the entry page? This way, it's more consitent throughout the website (even if we don't end up having tabs on all the kinds of entry pages).

aurel-l, line 52: This should probably be:
```ts
useEffect(() => {
  if (match && !match.params.subPage) {
    history.replace({
      ...history.location,
      pathname: generatePath(LocationToPath[Location.UniParcEntry], {
        accession: match.params.accession,
        subPage: TabLocation.Entry,
      }),
    });
  }
}, [match, history]);
```
In order to be consistent with the logic on the UniProtKB entry page

src/shared/components/layouts/styles/release-info.scss
xwatkins: Cool, didn't know about that property

src/uniparc/components/entry/Entry.tsx
xwatkins: Did you mean to replace the equivalent of `accession: accession` by `accession: match.params.accession` so that that `useEffect` is the exact same across namespaces?

xwatkins: Yes, or at least pages which use tabs...should it be a custom hook then?

aurel-l: It could be a custom hook, yes.
But yeah, my point is should all entries be /<id>/entry anyway, to make it consistent?

aurel-l: Yes, removes the extra check because accession will always be truthy if match is truthy, and makes it more useful that it's the same logic and thus could be abstracted

xwatkins: I see... I don't think so, as we will still have to provide paths to the entry without that suffix

aurel-l: This is doing redirect anyway for those entries with tabs, so it's the same. The non-`/entry` version would still work for all of them

aurel-l: Same logic in BLAST and Align results too.

src/uniparc/components/entry/Entry.tsx
```diff
@@ -45,22 +45,20 @@ const Entry: FC = () => {
...
-  const accession = match?.params.accession;
-
   // if URL doesn't finish with "entry" redirect to /entry by default
   useEffect(() => {
     if (match && !match.params.subPage) {
       history.replace({
         ...history.location,
         pathname: generatePath(LocationToPath[Location.UniParcEntry], {
-          accession,
+          accession: match?.params.accession,
```
aurel-l, line 54:
```suggestion
          accession: match.params.accession,
```

src/uniparc/components/entry/Entry.tsx
xwatkins: Ok - I think this could be tied in with the tabs too so would require more thought. I'll create a Jira

aurel-l (approved):

PR #314, opened 2021-03-17
2021 03 proteomes entry components

dlrice:
## Purpose
* Add Components section of proteomes entry page
* [Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-25400)

## Approach
* Add Components card
* The `Component Representation` column is not in the API data and never seems to be in the current website. Confirming with backend if this is intentional.
* Add `ComponentsButtons` to handle downloading and viewing the components' UniProtKB entries
* Made `namespace` a prop to `ColumnSelect` and `Download` rather than `useNS` as the proteomes components download is with UniProtKB data while the url indicates proteomes namespace.
* Added an app-level config for user storage - https://github.com/ebi-uniprot/uniprot-website/pull/312 also takes care of this and so will most likely replace the changes in this PR

## Testing
* Unit and snapshots

## Checklist


src/proteomes/components/entry/Components.tsx
```diff
@@ -0,0 +1,94 @@
...
+      render: ({ proteomeCrossReferences }) =>
+        proteomeCrossReferences
+          .filter(({ database, id }) => id && database === genomeAccessionDB)
+          .map<React.ReactNode>(({ id }) => (
+            <ExternalLink url={externalUrls.ENABrowser(id as string)} key={id}>
+              {id}
+            </ExternalLink>
+          ))
+          .reduce(
+            (prev, curr) => (prev === null ? [curr] : [prev, ', ', curr]),
```
aurel-l, line 55: Maybe we should avoid using reduce for this?
A Fragment in the map starting with a `, ` for all but index 0 should be enough

xwatkins:
I'm seeing an error with UP000005640

src/proteomes/components/entry/Components.tsx
dlrice: Done

dlrice:
> I'm seeing an error with UP000005640

Fixed

src/proteomes/components/entry/Components.tsx
aurel-l: 👍🏽
Not sure if it's wanted but now you don't have the extra space after the comma anymore

src/proteomes/components/entry/ComponentsButtons.tsx
```diff
@@ -0,0 +1,112 @@
...
+      }`,
+    [nComponents, id, selectedEntries]
+  );
+
+  const getViewButtonText = useCallback(() => {
+    const nSelected = selectedEntries.length;
+    if (nComponents === 1) {
+      return 'View UniProtKB entry for component';
+    }
+    return `View UniProtKB ${nSelected !== 1 ? 'entries' : 'entry'} for ${
```
xwatkins, line 58: This is a bit wordy - and even if we don't accomodate for individual component selection we need to reproduce the functionality in the "Map to on the left". Do we have the count breakdown for UniProtKB and TrEMBL in the response? If so the options should be (and consistently accross namespaces when we implement them)

View Proteins (dropdown)
Options:
UniProtKB (count)
Reviewed (Swiss-Prot) (count)
Unreviewed (TrEMBL) (count)

src/proteomes/components/entry/ComponentsButtons.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <>
+      {displayDownloadPanel && (
+        <Suspense fallback={null}>
+          <SlidingPanel
+            position={Position.left}
+            onClose={() => setDisplayDownloadPanel(false)}
+            yScrollable
+          >
+            <DownloadComponent
```
xwatkins, line 76: The "Preview n" functionality doesn't seem to work. For FASTA nothing happens and for table outputs I just get the table headers.

src/proteomes/components/entry/Components.tsx
```diff
@@ -0,0 +1,94 @@
...
+        id={id}
+      />
+      <DataTable
+        getIdKey={({ name }) => name}
+        density="compact"
+        columns={columns}
+        data={components}
+        selected={selectedEntries}
+        onSelectRow={handleItemSelection}
+        fixedLayout
```
xwatkins, line 91: The width for the data table checkbox column seems to be overwritten in css. This is in Franklin, I'm having a look at fixing this now. (This is not actually a comment, just a note for myself)

src/proteomes/components/entry/Components.tsx
dlrice: Good catch - I did want that extra space so it's not crowded

src/proteomes/components/entry/ComponentsButtons.tsx
dlrice: >Do we have the count breakdown for UniProtKB and TrEMBL in the response?

Not present in the proteome entry data. If the objective is to create a consistent map to/view proteins experience across the site i would suggest creating a separate component for this that takes a uniprotkb query, makes the requests and populates the dropdown with the counts along with the links.

xwatkins: I think this is already provided in the payload for some namespaces so might be an oversight. I'll make a Jira as this is important functionality. For now maybe just shorten to "View proteins" as a button?

xwatkins: https://www.ebi.ac.uk/panda/jira/browse/TRM-25668

aurel-l:
What's the status on this one and is there anything preventing it to go to sprint branch still?

dlrice:
> What's the status on this one and is there anything preventing it to go to sprint branch still?

Still working on the Preview n functionality. Will most likely have to wait until after the retro planning.

src/proteomes/components/entry/ComponentsButtons.tsx
dlrice: Fixed this. I am relying on protein count which isn't available yet so have hard coded some values as a fallback for now (with TODO to eventually remove). I had to adapt the Download component a bit to accommodate downloading the selected proteome components within the uniprotkb namespace.

xwatkins (approved):

PR #316, opened 2021-03-17
2021 03 taxonomy cards

xwatkins:
## Checklist


src/shared/components/entry/TaxonomyView.tsx
```diff
@@ -9,17 +9,26 @@ import { getEntryPath } from '../../../app/config/urls';
...
 };
 
-export const TaxonomyLineage: FC<{ lineage: string[] }> = ({ lineage }) => (
-  <>{lineage.join(' > ')}</>
-);
+export const TaxonomyLineage: FC<{ lineage: Lineage | string[] }> = ({
+  lineage,
+}) => {
+  if (isLineage(lineage)) {
+    return <></>;
```
aurel-l, line 28: This should probably just be `null`
```suggestion
    return null;
```

src/shared/components/entry/TaxonomyView.tsx
```diff
@@ -9,17 +9,26 @@ import { getEntryPath } from '../../../app/config/urls';
...
+  data: TaxonomyDatum;
 };
 
-export const TaxonomyLineage: FC<{ lineage: string[] }> = ({ lineage }) => (
-  <>{lineage.join(' > ')}</>
-);
+export const TaxonomyLineage: FC<{ lineage: Lineage | string[] }> = ({
+  lineage,
+}) => {
+  if (isLineage(lineage)) {
```
aurel-l, line 27: Not sure the wording is correct in representing the intent.
As it is not, I read it as "if this thing is a lineage, don't render a lineage, but if it's not then render a lineage" 🤔

src/supporting-data/taxonomy/adapters/taxonomyConverter.ts
```diff
@@ -16,19 +19,27 @@ type Strain = {
...
-  mnemonic: string;
+export type TaxonomyDatum = {
+  taxonId: number;
+  scientificName?: string;
+  synonyms?: string[];
+  commonName?: string;
+  mnemonic?: string;
+  evidences?: Evidence[];
+  // Note: eventually should all be an array of TaxonomyData
+  lineage?: Lineage | string[];
```
aurel-l, line 40: Probably should mark it as `@deprecated`, so that in the places where we use it as an array of string, the editor will remind us it is deprecated.
Something like this would cover it:
```ts
type TaxonomyBase = {
  taxonId: number;
  scientificName?: string;
  synonyms?: string[];
  commonName?: string;
  mnemonic?: string;
  evidences?: Evidence[];
  lineage?: Lineage;
  hidden?: boolean;
  rank?: Rank;
};

export type TaxonomyDatum =
  | TaxonomyBase
  | (TaxonomyBase & {
      /**
       * @deprecated "lineage" should be a `Lineage`
       */
      lineage?: string[];
    });
```

src/supporting-data/taxonomy/components/results/TaxonomyCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { FC, useCallback } from 'react';
+import { useHistory } from 'react-router-dom';
+import { getEntryPath } from '../../../../app/config/urls';
+import TaxonomyView from '../../../../shared/components/entry/TaxonomyView';
+import RenderColumnInCard from '../../../../shared/components/results/RenderColumnInCard';
+import { Namespace } from '../../../../shared/types/namespaces';
+import { TaxonomyAPIModel } from '../../adapters/taxonomyConverter';
+import TaxonomyColumnConfiguration, {
+  TaxonomyColumn,
+} from '../../config/TaxonomyColumnConfiguration';
```
aurel-l, line 11:
```suggestion
import { Card } from 'franklin-sites';
import { FC, useCallback } from 'react';
import { useHistory } from 'react-router-dom';

import TaxonomyView from '../../../../shared/components/entry/TaxonomyView';
import RenderColumnInCard from '../../../../shared/components/results/RenderColumnInCard';

import { getEntryPath } from '../../../../app/config/urls';

import { Namespace } from '../../../../shared/types/namespaces';
import { TaxonomyAPIModel } from '../../adapters/taxonomyConverter';
import TaxonomyColumnConfiguration, {
  TaxonomyColumn,
} from '../../config/TaxonomyColumnConfiguration';
```

src/supporting-data/taxonomy/components/results/TaxonomyCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <div className="result-card__right">
+          <h5>
+            <TaxonomyView data={data} />
+          </h5>
+          <div className="result-card__info-container">
+            {mainInfoColumns.map((column) => (
+              <RenderColumnInCard
+                type={column}
+                data={data}
+                columnConfig={TaxonomyColumnConfiguration}
```
aurel-l, line 47: Looking at the API of this now in usage, I'm thinking there's not too much point passing both `type` and `columnConfig` as the full column config where you could just pass, here for example, `TaxonomyColumnConfiguration.get(column)` for the `columnConfig` prop

src/supporting-data/taxonomy/config/TaxonomyColumnConfiguration.tsx
```diff
@@ -84,7 +84,7 @@ TaxonomyColumnConfiguration.set(TaxonomyColumn.lineage, {
   label: 'Lineage',
   // TODO: modify when we have a common approach to represent lineages
   render: ({ lineage }) =>
-    lineage
+    (lineage as Lineage)
       ?.filter(({ hidden }) => !hidden)
```
aurel-l, line 88: By the way, I've noticed that when it's an object, the order should be inverted as the highest-level taxon will be last
```suggestion
      ?.filter(({ hidden }) => !hidden).reverse()
```

src/supporting-data/taxonomy/adapters/taxonomyConverter.ts
```diff
@@ -16,19 +19,29 @@ type Strain = {
   name: string;
 };
 
-type TaxonomyLight = {
-  taxonId: number;
-  scientificName: string;
-  synonyms?: string[];
-  commonName?: string;
-};
+export type Lineage = Array<SetRequired<TaxonomyDatum, 'hidden' | 'rank'>>;
```
aurel-l, line 22: Maybe there are even more fields that can be removed in this case?
```suggestion
export type Lineage = Array<SetRequired<Omit<TaxonomyDatum, 'lineage'>, 'hidden' | 'rank'>>;
```

Thinking there might be more places around this file where the use of `Omit` might be good, so that we don't conditionally do stuff on fields that we should know are going to be `never` (hosts? strains? Anything that is a call to a recursive type of TaxonomyDatum could be candidate)

aurel-l (changes requested):
Apart from the single comments raised here and there, just a general one about casting lineage as either `string[]` or `Lineage` depending on the place. I don't really know what's the best approach here if we want to not have those using `string[]` without us knowing when the backend is changed and starts returning objects. Hopefully the `@deprecated` tag will help us see it in the code more easily, but it doesn't do much more than that.

xwatkins:
I've also merged the `TaxonomyView` and `TaxonomyLightView` components as they were very similar. Lost the "title" attribute on the Link, which could be re-added but I don't think it's necessary because it's the same text as in the anchor.

src/supporting-data/taxonomy/components/results/TaxonomyCard.tsx
aurel-l: Proposal alternative approach in https://github.com/ebi-uniprot/uniprot-website/pull/319

src/proteomes/config/__tests__/__snapshots__/ProteomesColumnConfiguration.spec.tsx.snap
```diff
@@ -72,9 +72,8 @@ exports[`ProteomesColumnConfiguration component should render column "organism":
 <DocumentFragment>
   <a
     href="/taxonomy/9606"
-    title="Homo sapiens (Human), taxon ID 9606"
   >
-    Homo sapiens (Human)
+    Homo sapiens (Human) (synonym)
```
aurel-l, line 76: I think that might not be what is expected here

aurel-l: Actually nevermind, it's in the mock

aurel-l:
> I've also merged the `TaxonomyView` and `TaxonomyLightView` components as they were very similar. Lost the "title" attribute on the Link, which could be re-added but I don't think it's necessary because it's the same text as in the anchor.

OK so the reason why I had this title was to distinguish between taxons with the same name. The only unique thing of a taxon is its ID, so I wanted to display that to screen readers and on mouse over.
See example in https://ebi-uniprot.github.io/uniprot-website/uniparc/UPI00000ED0BA/entry , click "more items" in organisms facet.

src/proteomes/config/__tests__/__snapshots__/ProteomesColumnConfiguration.spec.tsx.snap
xwatkins: I know, I did a double take too :-)

xwatkins:
> > I've also merged the `TaxonomyView` and `TaxonomyLightView` components as they were very similar. Lost the "title" attribute on the Link, which could be re-added but I don't think it's necessary because it's the same text as in the anchor.
> 
> OK so the reason why I had this title was to distinguish between taxons with the same name. The only unique thing of a taxon is its ID, so I wanted to display that to screen readers and on mouse over.
> See example in https://ebi-uniprot.github.io/uniprot-website/uniparc/UPI00000ED0BA/entry , click "more items" in organisms facet.

That looks really weird in the facets but I guess if they have different taxids... will add the `title` back

src/supporting-data/taxonomy/adapters/taxonomyConverter.ts
xwatkins: Not sure, I started playing a bit with it but think it might be more pain than what it gives us, especially in the long run if we have to do this everywhere.

aurel-l: Well I'd do at least the one in the suggestion then, because this is an obvious one.

aurel-l (approved):

src/supporting-data/taxonomy/components/results/TaxonomyCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const TaxonomyCard: FC<{
+  data: TaxonomyAPIModel;
+  selected?: boolean;
+  handleEntrySelection: (rowId: string) => void;
+}> = ({ data, selected, handleEntrySelection }) => {
+  const history = useHistory();
+
+  const handleCardClick = useCallback(() => {
+    history.push(getEntryPath(Namespace.proteomes, data.taxonId));
```
aurel-l, line 26: This is incorrect, this should be `Namespace.taxonomy`.

src/supporting-data/taxonomy/components/results/__tests__/TaxonomyCard.spec.tsx
```diff
@@ -0,0 +1,13 @@
+import renderWithRouter from '../../../../../shared/__test-helpers__/customRender';
+import TaxonomyCard from '../TaxonomyCard';
+
+import mockData from '../../../__mocks__/taxonomyModelData';
+
+describe('Taxonomy Card tests', () => {
```
aurel-l, line 6: This should contain a `'should allow card selection and navigation'` test as for all the other card renderers for other namespaces, this would have caught this issue.

aurel-l (commented):
There is an issue with link the onClick is generating

aurel-l:
> There is an issue with link the onClick is generating

Handled in https://github.com/ebi-uniprot/uniprot-website/pull/319

PR #319, opened 2021-03-22
add CSS modules logic, with TS support

aurel-l:
## Purpose
Support for CSS modules and refactor of RenderColumnInCard

## Approach
 - RenderColumnInCard -> RenderColumnsInCard taking a list of column renderers as prop
 - Support for CSS modules without breaking existing files -> opt-in
 - Support for CSS modules in TypeScript -> autocompletion of class names in IDE!
 - Fix issue in taxonomy card click handler

## Testing
Tests still running, coverage slightly down but would address that after review

## Checklist


src/shared/components/results/RenderColumnsInCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+      );
+    })
+    .filter(Boolean);
+
+  return renderedNodes.length ? (
+    <div className={styles['result-card__info-container']}>{renderedNodes}</div>
+  ) : null;
+};
+
+export default memo(RenderColumnsInCard) as typeof RenderColumnsInCard;
```
dlrice, line 47: In this instance `data` is deeply nested and `renderers` is an array of functions so shallow comparison between previous props and next props won't work and instead reference comparison is used and I don't think the props references will be stable. Providing our own `areEqual` and just looking at the accession/id could be a good comparison. I played around with the facets on proteomes results view to see when it does re-render:

```
const areEqual = (p, n) => {
  console.log(p.data.id, n.data.id, p == n, p.data.id == n.data.id);
  // return p == n;
  return p.data.id == n.data.id;
};
export default memo(
  RenderColumnsInCard,
  areEqual
) as typeof RenderColumnsInCard;

```
One issue in my suggestion is that we won't always have `id` available so we would have to somehow indicate what to compare on or just try id-like attributes one by one (try id and if that's not there try accession and if that's not there try...).

src/uniparc/components/results/__tests__/__snapshots__/UniParcCard.spec.tsx.snap
```diff
@@ -13,7 +13,7 @@ exports[`UniRefCard tests should render and match snapshot 1`] = `
       <div
         class="card__content"
       >
-        <section
+        <div
```
dlrice, line 16: The ` · ` disappears for some reason.
image

src/shared/components/results/RenderColumnsInCard.tsx
aurel-l: About the id, on top of the limitation that there might not always be an `id` field so we would need to have more complexe logic, I'm not a fan of trying to check if the id is the same anyway, as differents requests (with the difference in the fields for example) will return different objects with the same id so we wouldn't detect new objects when we actually do need to.
Also, I'm assuming in the case where our data object is meant to stay the same, it will, as it is the raw object returned by the API through `useDataAPI` and `useDataAPI` is in charge of returning the same actual object at every render.

Regarding the `renderers` props, since this selection of renderers is fixed for any given card, I always define that array outside of the component in order for it to be always the same array.

src/uniparc/components/results/__tests__/__snapshots__/UniParcCard.spec.tsx.snap
aurel-l: This was relying on classes defined in the stylesheet for `RenderColumnsInCard` so now that they're in CSS modules I needed to explicitly import them. Done now

src/proteomes/components/results/__tests__/__snapshots__/ProteomesCard.spec.tsx.snap
```diff
@@ -150,8 +149,8 @@ exports[`ProteomesCard tests should render the card component 1`] = `
                       Missing
                     </span>
                   </span>
+                  :
```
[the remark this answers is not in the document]
aurel-l: Done now, I made it not use the `RenderColumnsInCard` component but just its classes.
Also, extended this specific case to have the BUSCO visualisation spread the whole width of the card.

src/shared/components/results/RenderColumnsInCard.tsx
dlrice: All good points. When should we be preventing a re-render - when will `memo` stop react from re-rendering? Facets will trigger a re-render as it's coming from a different query.

aurel-l: I don't believe the logic within these components is so complex that it would be worth having an extra complex logic to determine if we should re-render or not.
When facets changes, the data object will be new indeed, but likely the resulting list of entries to display would change too anyway so it would be OK.
On the top of my mind the only case where this might be useful is when checking/unchecking the selected checkboxes, that by nature forces the rerendering of all the cards, while most all of their content is unchanged (apart from one specific checkbox). Ideally we'd want the whole rest of the card to be a memoized component and the checkbox to be separated, but that falls a bit further out of the scope of this PR I believe.

dlrice: Good point about the checkboxes

dlrice (approved):

PR #322, opened 2021-03-23
2021 04 peptide search form and peptide search and id mapping submission

aurel-l:
## Purpose
Create a Peptide Search form and implement submission logic. [TRM-25699](https://www.ebi.ac.uk/panda/jira/browse/TRM-25699) and [TRM-25640](https://www.ebi.ac.uk/panda/jira/browse/TRM-25640)
Implement ID Mapping submission logic. [TRM-25700](https://www.ebi.ac.uk/panda/jira/browse/TRM-25700)

## Approach
Same as purpose

## Testing
No specific test while specs are fluid

## Checklist
  - Lots of new logic in job middleware, difficult to unit-test

Current issues and doubts:
 - [ ] unable to use the endpoint, so fall back to current UniProt website as make-do endpoint (will get issue fetching results, seems like they stay forever running unless we do a browser browse of the URL)
 - [ ] As per Jira, supposed to use the existing sequence field and validation, but peptide search doesn't currently support FASTA, and does support smaller sequence -> What we currently have doesn't fit.
 - [ ] We have a component in BLAST and Align forms to fetch and pre-fill sequences from existing entries. Should we have it here too? Might be a tool for the user to fetch the sequence then trim it to the wanted subsequence, but maybe user will think this is the canonical/only way to do peptide searches

Note: To submit an ID Mapping job on this PR, submit any Peptide Search through the form and it will create a predefined ID Mapping job alongisde it


src/tools/id-mapping/types/idMappingServerParameters.ts
```diff
@@ -0,0 +1,23 @@
...
+export type ServerParameters = {
+  from: From;
+  to: To;
+  ids: IDs;
+};
+
+// same, because no email was provided before
+export type PublicServerParameters = ServerParameters;
+
+export type IDMappingTarget =
```
xwatkins, line 19: I guess that list will grow with Dan's PR?

src/tools/peptide-search/components/__tests__/PeptideSearchForm.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+import initialState from '../../../../app/state/rootInitialState';
+
+import { mockSuggesterApi } from '../../../../query-builder/components/__tests__/__mocks__/autocompleteWrapperData';
+
+const mock = new MockAdapter(axios);
+mock.onGet().reply(200, mockSuggesterApi.response);
+
+const aaSequence = 'ABCDEFGHIJKLMNOPQRST';
+const ntSequence = 'ATCGAGCGATAGCGAGGGAC';
+
```
xwatkins, line 19: Should we try with an invalid sequence too?

src/tools/peptide-search/config/PeptideSearchFormData.ts
```diff
@@ -0,0 +1,66 @@
...
+  spOnly = 'Search UniProt Reviewed (Swiss-Prot) only',
+  name = 'Name',
+}
+
+export type PeptideSearchFormValues = Record<
+  PeptideSearchFields,
+  Readonly<PeptideSearchFormValue>
+>;
+
+const formData: Readonly<PeptideSearchFormValues> = Object.freeze({
```
xwatkins, line 33: Can you remind me why the `Object.freeze`?

src/tools/state/getCheckJobStatus.ts
```diff
@@ -13,6 +13,7 @@ import { RunningJob, FinishedJob } from '../types/toolsJob';
 import { Status } from '../types/toolsStatuses';
 import { BlastResults } from '../blast/types/blastResults';
 import { JobTypes } from '../types/toolsJobTypes';
+import { IDMappingTarget } from '../id-mapping/types/idMappingServerParameters';
```
xwatkins, line 16: Not needed (not sure why eslint isn't complaining though, is this file not checked?)

xwatkins (changes requested):

src/tools/peptide-search/config/PeptideSearchFormData.ts
aurel-l: To make sure to not mutate the object while processing it in the form, this is due to the complexity of the form description object and the difficulty to track unexpected side effects related to accidentally mutating it

src/tools/id-mapping/types/idMappingServerParameters.ts
aurel-l: Open to suggestions for a better variable name

xwatkins: Ah I see - `IDMappingResultsNamespace`? `MappingResultsNamespace`?

src/tools/peptide-search/components/__tests__/PeptideSearchForm.spec.tsx
aurel-l: All those tests are skipped now, pending discussion point #2

src/tools/state/getCheckJobStatus.ts
aurel-l: Not sure why that happened either...
Removed now

src/tools/id-mapping/types/idMappingServerParameters.ts
aurel-l: Changed

xwatkins (approved):

PR #325, opened 2021-03-24
2021 04 UI tiles issues

xwatkins:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-25708

## Testing
Visual

## Checklist


src/app/components/styles/home-page-non-critical.scss
```diff
@@ -0,0 +1,6 @@
+.supporting-data-section {
+  h3 {
```
aurel-l, line 2: Not sure if it's the right selector. When looking at the heading flow of the page here in this place we are using an h3 when what we should be using is a h2 so that will need to change at some point.

src/app/components/HomePageNonCritical.tsx
```diff
@@ -28,6 +28,8 @@ import TechDocIllustration from '../../images/tech_doc_illustration.svg';
 import PlaceHolder from './PlaceHolder';
 import { Location, LocationToPath } from '../config/urls';
 
+import './styles/home-page-non-critical.scss';
```
aurel-l, line 31: Should we slowly add new stylesheet files as CSS modules so that we get used to it?

xwatkins: Ah yes, didn't know if that had gone in yet!

src/app/components/styles/home-page-non-critical.scss
aurel-l: Note that if you are going to use `tile__header` from Franklin, but switch this stylesheet to CSS module, you need to use this https://github.com/css-modules/css-modules#exceptions for global class names

aurel-l: Also, should this actually be in franklin?

xwatkins: This is not styling the tile header but making an exception for the supporting data container header. 
I will create a class called `.supporting-data-section__header` and assign put it on the `h3` as this will get the issue of tag specificity if we change the hierarchy one day

aurel-l: Are you saying this is actually for https://www.ebi.ac.uk/panda/jira/browse/TRM-25709 ?

xwatkins: No it's for this title 
image

aurel-l: Screenshot 2021-03-25 at 10 27 58

xwatkins: Sorry, I need to get my glasses 🤓

src/app/components/HomePageNonCritical.tsx
```diff
@@ -97,12 +98,18 @@ const HomePageNonCritical = () => {
       </section>
 
       <HeroContainer
-        className="uniprot-grid uniprot-grid--centered uniprot-grid--with-bleed supporting-data-section"
+        className={cn(
```
aurel-l, line 101: why the extra logic?

xwatkins: Initially to add the `styles[]` which in the end wasn't needed at that level, but also I think classnames makes it more readable - although happy to remove it

src/app/components/styles/home-page-non-critical.scss
aurel-l: OK, so there it still needs to be an h2. If you need it to have the style of an h3 we should be theming it with Franklin mixins' `@include fs-font-settings('H3');`

xwatkins: It's currently an h3 in all other sections - "News" "Analysis Tools" "Need Help etc...

I agree that the hierarchy of titles needs to be revisited across the website (the one that grinds my gears the most is the  entry page tab header being at the same level as the title in the overview) but this will be done after the alpha. I'd like to revisit the type sizing in Franklin too.

aurel-l: I'm saying that because I'm changing the tile component at the moment, and including this in the batch of changes.

aurel-l (approved):

PR #329, opened 2021-03-26
2021 04 fix download columns

xwatkins:
## Purpose
Column selection in download was not applied.

## Approach
The list of columns in the state was not the list passed to the url generator.

## Testing
Added a test to check the url is updated correctly when removing a column.

## Checklist

### DIFF
--- a/src/shared/components/download/__tests__/Download.spec.tsx
+++ b/src/shared/components/download/__tests__/Download.spec.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, screen, waitFor } from '@testing-library/react';
+import { fireEvent, screen } from '@testing-library/react';
 
 import customRender from '../../../__test-helpers__/customRender';
 
--- a/src/shared/components/download/__tests__/Download.spec.tsx
+++ b/src/shared/components/download/__tests__/Download.spec.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, screen } from '@testing-library/react';
+import { act, fireEvent, screen, waitFor } from '@testing-library/react';
 
 import customRender from '../../../__test-helpers__/customRender';
 
@@ -32,9 +32,11 @@ describe('Download component', () => {
   const namespace = Namespace.uniprotkb;
   const query = 'nod2';
   const selectedEntries = ['Q9HC29', 'O43353', 'Q3KP66'];
-  const onCloseMock = jest.fn();
+  let onCloseMock;
 
   beforeEach(() => {
+    onCloseMock = jest.fn();
+
     customRender(
       <Download
         query={query}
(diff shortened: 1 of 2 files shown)


src/shared/components/download/__tests__/Download.spec.tsx
```diff
@@ -90,18 +95,39 @@ describe('Download component', () => {
...
+    async (value, columnSelect) => {
       const formatSelect = screen.getByTestId('file-format-select');
-      fireEvent.change(formatSelect, { target: { value } });
-      const customise = screen.queryByText('Customize data');
-      if (columnSelect) {
-        expect(customise).toBeInTheDocument();
-      } else {
-        expect(customise).not.toBeInTheDocument();
-      }
+      await act(async () => {
```
aurel-l, line 100: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily

src/shared/components/download/__tests__/Download.spec.tsx
```diff
@@ -90,18 +95,37 @@ describe('Download component', () => {
...
+    async (value, columnSelect) => {
       const formatSelect = screen.getByTestId('file-format-select');
       fireEvent.change(formatSelect, { target: { value } });
-      const customise = screen.queryByText('Customize data');
-      if (columnSelect) {
-        expect(customise).toBeInTheDocument();
-      } else {
-        expect(customise).not.toBeInTheDocument();
-      }
+      await waitFor(() => {
```
aurel-l, line 101: Why is this needed? What's the difference with before that now requires this to be wrapped by waitFor?

xwatkins: The change event triggers a state change, so the elements might not be there at the time of check. I'm not sure it worked as expected before?

aurel-l: I'm wondering trying to reuse `customise` variable makes it a bit more complex that it needs to be. Seeing that bit of test looks to me like this bit of the article https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

xwatkins: So I could switch to `await screen.findByText('Customize data');` and then check for `null` instead of `toBeInTheDocument`.

xwatkins: Actually it's not great is it? It shouldn't be waiting for something we know shouldn't be there... 🤔

aurel-l: True. Is there a better question to ask the DOM? Under what condition we are ready to expect that something isn't there not just because it didn't had time to appear, but because it will not be there even after the next render or whatever

xwatkins: So actually it turns out the async/await wasn't needed so I've reverted it back to how it was.

aurel-l (approved):

PR #331, opened 2021-03-30
2021 04 tools idmapping results

xwatkins:
## Purpose
WIP to display results for ID Mapping.

Functionality not part of this PR
- Facets
- Action buttons
- Download
- Make request to `details` endpoint to get "From" and "To" databases
- Generate external links based on "To" database

## Approach
Add column renderers used by the relevant namespaces. Create a new Namespace for `id-mapping`.

## Testing
New test to check presence of relevant data both for `id-mapping` namespace (2 columns) and for one of the other namespaces, `uniprotkb`.

## Checklist


src/tools/id-mapping/adapters/idMappingConverter.ts
```diff
@@ -0,0 +1,17 @@
+import { UniParcAPIModel } from '../../../uniparc/adapters/uniParcConverter';
+import { UniProtkbAPIModel } from '../../../uniprotkb/adapters/uniProtkbConverter';
+import { UniRefAPIModel } from '../../../uniref/adapters/uniRefConverter';
+import { MappingAPIModel, MappingFlat } from '../types/idMappingSearchResults';
+
+const idMappingConverter = (data: MappingAPIModel[]): MappingFlat[] =>
+  data.map((row) => {
```
xwatkins, line 7: Bit of an issue here with types. For `usePagination` to stay generic I would like to return something which extends `APIModel`. That's why I added the `R` as return type for `usePagination` but I think it has problems to match the return type.

src/tools/id-mapping/types/idMappingSearchResults.ts
```diff
@@ -1,13 +1,25 @@
...
+export type MappingTo = {
+  to: string;
+};
+
+export type MappingFrom = {
+  from: string;
+};
+
+// Should this be a subset of APIMode, and if so how do we handle it in usePagination?
+export type MappingFlat = (APIModel | MappingTo) & MappingFrom;
```
xwatkins, line 25: Should this be a subset of APIMode, and if so how do we handle it in usePagination?

aurel-l: If you want usePagination to work with IDMapping search results you should add those definitions to APIModel directly where it's defined I think. The from in MappingFrom should then be optional, and maybe you could add it only to the 4 namespaces supported for ID Mapping.

src/app/config/urls.ts
```diff
@@ -72,8 +72,7 @@ export const LocationToPath: Record<Location, string> = {
   [Location.Blast]: '/blast',
   [Location.PeptideSearchResult]: '/peptide-search/:id/:subPage?',
   [Location.PeptideSearch]: '/peptide-search',
-  // TODO: check final URL for those
-  [Location.IDMappingResult]: '/id-mapping/:id/:subPage?',
+  [Location.IDMappingResult]: '/id-mapping/:id/:targetNS?',
```
aurel-l, line 75: This probably needs to be inverted.
If you don't want an ID to be matched as a targetNS, explicitely state which are the supported namespaces (see below definition of `allSearchResultLocations`)

xwatkins: I've put `targetNS` at the end because it can be empty (when the target is undefined).

aurel-l: Still, since it's the opposite of the URL format given by the server I'm not too keen on going for a different pattern

src/app/config/urls.ts
```diff
@@ -72,7 +72,7 @@ export const LocationToPath: Record<Location, string> = {
   [Location.Blast]: '/blast',
   [Location.PeptideSearchResult]: '/peptide-search/:id/:subPage?',
   [Location.PeptideSearch]: '/peptide-search',
-  [Location.IDMappingResult]: '/id-mapping/:id/:targetNS?',
+  [Location.IDMappingResult]: '/id-mapping/:targetNS?/:id',
```
aurel-l, line 75:
```suggestion
  [Location.IDMappingResult]: '/id-mapping/:targetNS(${IDMappingNamespaces.join('|')})?/:id',
```

src/app/config/urls.ts
```diff
@@ -91,10 +91,20 @@ export const SearchResultsLocations: Record<Namespace, string> = {
...
+  Namespace.uniprotkb,
+  Namespace.uniref,
+  Namespace.uniparc,
+];
+
 // "/:namespace(uniprotkb|uniparc|........)/""
 export const allSearchResultLocations = `/:namespace(${Object.values(
   Namespace
-).join('|')})/`;
+).join('|')}`;
```
aurel-l, line 103:
```suggestion
).join('|')})`;
```

src/shared/hooks/useNS.ts
```diff
@@ -1,20 +1,29 @@
...
+import { allIDMappingTargetLocations } from '../../app/config/urls';
 
 import { Namespace } from '../types/namespaces';
+import { IDMappingNamespace } from '../../tools/id-mapping/types/idMappingServerParameters';
 
-const useNS = (): Namespace | undefined => {
-  const match = useRouteMatch<{ namespace: Namespace }>(
-    allSearchResultLocations
-  );
+const useNS = (): [Namespace, boolean] => {
```
dlrice, line 7: Could we keep `IDMappingNamespace` out of `Namespace` by having `useNS` look at the url to determine if it's at a tool location (id mapping or peptide search) and returning `[namespace, tool]` in that instance.

xwatkins: There is a particularity to `id-mapping` in the sense that it is _also_ a namespace (when the target is just an ID). This means it has its own definition for default columns, primary key, column config etc...
However if this is useful for peptide search, I could return `secondaryNS` instead of a boolean:
- `/id-mapping/12345/uniprot` -> `[uniprot, idmapping]`
- `/id-mapping/12345/uniref` -> `[uniref, idmapping]`
- `/id-mapping/12345` -> `[idmapping, undefined]`
- `/peptidesearch/12345` -> `[uniprot, peptidesearch]`

aurel-l: `/id-mapping/uniprot/12345`*
`/id-mapping/uniref/12345`*

xwatkins: Thanks @aurel-l , edited

dlrice: I've just merged in this branch into my peptide search branch and everything I've implemented so far works fine so need to change your code.

src/tools/id-mapping/components/results/IDMappingResult.tsx
```diff
@@ -0,0 +1,77 @@
...
+  MappingAPIModel,
+  MappingFlat,
+} from '../../types/idMappingSearchResults';
+import { Namespace } from '../../../../shared/types/namespaces';
+
+const jobType = JobTypes.ID_MAPPING;
+const urls = toolsURLs(jobType);
+
+const IDMappingResult = () => {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
aurel-l, line 26: I'm assuming this override is not needed anymore?

dlrice:
LGTM.

How do you plan on tackling download? I'll run through my plan for peptide search. For Download the props are currently:
```
type DownloadProps = {
  query?: string;
  selectedEntries?: string[];
  selectedQuery?: string;
  totalNumberResults: number;
  numberSelectedEntries?: number;
  namespace: Namespace;
  onClose: () => void;
};
```
And I'm thinking of adding an additional `accessions?: string[]` which will then be passed to `getDownloadUrl` and when present will cause a request to the accessions end point. Will you also need this functionality for ID mapping?

xwatkins:
> LGTM.
> 
> How do you plan on tackling download? I'll run through my plan for peptide search. For Download the props are currently:
> 
> ```
> type DownloadProps = {
>   query?: string;
>   selectedEntries?: string[];
>   selectedQuery?: string;
>   totalNumberResults: number;
>   numberSelectedEntries?: number;
>   namespace: Namespace;
>   onClose: () => void;
> };
> ```
> 
> And I'm thinking of adding an additional `accessions?: string[]` which will then be passed to `getDownloadUrl` and when present will cause a request to the accessions end point. Will you also need this functionality for ID mapping?

Haven't looked into it properly yet but I wouldn't need that - the main issue will be with using the "correct namespace" again (ie for uniprotkb/uniref/uniparc target I want to use these as namespace for columns etc... but still need to hit the idmapping endpoint to retrieve the results). So I might need further changes to `useNS`!

xwatkins:
I've added logic to use the new `details` endpoint and generate the corresponding xref external link. Also noticed an issue with the namespace when reloading the page, so will fix that this afternoon.

aurel-l:
> I've added logic to use the new `details` endpoint and generate the corresponding xref external link. Also noticed an issue with the namespace when reloading the page, so will fix that this afternoon.

Since we now have this information available, should we revert to not use `/id-mapping/:targetNS(${IDMappingNamespaces.join('|')})?/:id` as URL and just a plain, more consistent with the other jobs, `/id-mapping/:id`?

xwatkins:
> > I've added logic to use the new `details` endpoint and generate the corresponding xref external link. Also noticed an issue with the namespace when reloading the page, so will fix that this afternoon.
> 
> Since we now have this information available, should we revert to not use `/id-mapping/:targetNS(${IDMappingNamespaces.join('|')})?/:id` as URL and just a plain, more consistent with the other jobs, `/id-mapping/:id`?

I think it would ultimately be good and cleaner, but we would need a way to bypass the use of `useNS` completely in places like `useColumns`, `ResultsFacets`, `ResultsData`, `useNSQuery` etc.... - any ideas on how to do this efficiently?

aurel-l:
> > > I've added logic to use the new `details` endpoint and generate the corresponding xref external link. Also noticed an issue with the namespace when reloading the page, so will fix that this afternoon.
> > 
> > 
> > Since we now have this information available, should we revert to not use `/id-mapping/:targetNS(${IDMappingNamespaces.join('|')})?/:id` as URL and just a plain, more consistent with the other jobs, `/id-mapping/:id`?
> 
> I think it would ultimately be good and cleaner, but we would need a way to bypass the use of `useNS` completely in places like `useColumns`, `ResultsFacets`, `ResultsData`, `useNSQuery` etc.... - any ideas on how to do this efficiently?

Pass the namespace as a prop for the component using it. useNS would return undefined, and could default the one passed as a prop.
It is the same kind of behaviour we'll need for the supporting data map-to results section, to pass a custom endpoint URL for the results component to get data from.
In summary: get from the prop if defined, otherwise use one of the custom hooks to determine the value from context (context being the URL)

xwatkins:
> > > > I've added logic to use the new `details` endpoint and generate the corresponding xref external link. Also noticed an issue with the namespace when reloading the page, so will fix that this afternoon.
> > > 
> > > 
> > > Since we now have this information available, should we revert to not use `/id-mapping/:targetNS(${IDMappingNamespaces.join('|')})?/:id` as URL and just a plain, more consistent with the other jobs, `/id-mapping/:id`?
> > 
> > 
> > I think it would ultimately be good and cleaner, but we would need a way to bypass the use of `useNS` completely in places like `useColumns`, `ResultsFacets`, `ResultsData`, `useNSQuery` etc.... - any ideas on how to do this efficiently?
> 
> Pass the namespace as a prop for the component using it. useNS would return undefined, and could default the one passed as a prop.
> It is the same kind of behaviour we'll need for the supporting data map-to results section, to pass a custom endpoint URL for the results component to get data from.
> In summary: get from the prop if defined, otherwise use one of the custom hooks to determine the value from context (context being the URL)

My issue with this approach was that it's not then obvious to a user of the component that the namespace would be sorted inside the component automatically and might think they _must_ provide the namespace as a prop (even if optional). Maybe a name like `namespaceOverride` would help with that?

src/shared/components/results/ResultsData.tsx
```diff
@@ -32,11 +32,23 @@ const ResultsData: FC<{
...
+  displayIdMappingColumns?: boolean;
+}> = ({
+  resultsDataObject,
+  direct,
+  selectedEntries,
+  handleEntrySelection,
+  namespaceOverride,
+  displayIdMappingColumns,
+}) => {
+  const namespace = useNS() || namespaceOverride || Namespace.uniprotkb;
```
aurel-l, line 45: Naming-wise, now that I see it in context, I'd call that `namespaceFallback` then, as it's not really overriding the internal value so behaviour might be unexpected

xwatkins:
3rd time lucky... new new new approach. @dlrice no more tuple in `useNS`, don't know if you could use a similar approach in your PR? Changes from the PR shouldn't affect yours anyway now.

PR #332, opened 2021-03-31
2021 04 id mapping form

dlrice:
## Purpose
- Implement ID mapping form
- [TRM-25638](https://www.ebi.ac.uk/panda/jira/browse/TRM-25638)

## Approach
- Emulate approach in other tool forms
- Use Franklin's TreeSelect for `From` and `To` database selection

## Testing
Unit tests and snapshot

## Checklist


src/app/components/App.tsx
```diff
@@ -292,6 +296,18 @@ const App = () => (
                 </SingleColumnLayout>
               )}
             />
+            {/* <Route
```
aurel-l, line 299: You should remove the commented out route above then

aurel-l: Or actually put it above the dashboard route to order and group similar routes together

dlrice: Updated

src/tools/id-mapping/components/IDMappingForm.tsx
```diff
@@ -0,0 +1,455 @@
...
+  // refs
+  const fileInputRef = useRef<HTMLInputElement>(null);
+
+  // hooks
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const reducedMotion = useReducedMotion();
+
+  // state
+  const initialFormValues = useMemo(() => {
```
dlrice, line 87: We repeat this in a few places in the code base where the only thing that changes is the types. What do you think about making this a custom hook?

aurel-l: Sounds good, up to you if you want to have this in this PR or do a different one. Now that we have the 4 forms more or less ready it's a good time to start thinking about what can be refactored and put in common

dlrice: Will make a separate PR for this

src/tools/id-mapping/components/IDMappingForm.tsx
```diff
@@ -0,0 +1,465 @@
...
+      // here we should just transform input values into FormParameters,
+      // transformation of FormParameters into ServerParameters happens in the
+      // tools middleware
+      const parameters: FormParameters = {
+        from: fromDb.selected as string,
+        to: toDb.selected as string,
+        ids,
+      };
+
+      if (ruleInfo && ruleInfo?.taxonId && taxID.selected) {
```
aurel-l, line 193:
```suggestion
      if (ruleInfo?.taxonId && taxID.selected) {
```

src/tools/id-mapping/components/IDMappingForm.tsx
```diff
@@ -0,0 +1,465 @@
...
+      const ruleInfo =
+        fromDbInfo?.ruleId && ruleIdToRuleInfo[fromDbInfo.ruleId];
+
+      // here we should just transform input values into FormParameters,
+      // transformation of FormParameters into ServerParameters happens in the
+      // tools middleware
+      const parameters: FormParameters = {
+        from: fromDb.selected as string,
+        to: toDb.selected as string,
+        ids,
```
aurel-l, line 190: OK, so have a think if you want to pass here the cleaned input, or the raw input. If you pass the raw input, the cleaning can be done again inside the `formParametersToServerParameters` function.

aurel-l: If you do use the raw, it would mean `FormParameters['ids']` needs to be defined as `string` instead of `string[]`

aurel-l: But as I said, it's not a requirement for me, it's up to you

aurel-l (approved):
Approve in general, just a small optional chaining change and general comment about what to store

aurel-l (approved):

PR #334, opened 2021-04-06
2021 04 peptide search part 2

aurel-l:
## Purpose
Next iteration on job submission for peptide search and id mapping

## Approach
use fetch instead of axios to avoid redirects, use new peptide search endpoint with correct CORS settings

## Testing
Kept and adapted existing tests

## Checklist

Note: when doing the status check, retrieves the ID Mapping namespace required for the results page and stores it in the job information


src/tools/utils/index.tsx
```diff
@@ -18,33 +18,33 @@ const validServerID: Record<JobTypes, RegExp> = {
...
 
-const peptideSearchJobPattern = /\/peptidesearch\/uniprot\/(?<jobID>[A-Z\d]+)$/;
+const peptideSearchJobPattern = /\/asyncrest\/jobs\/(?<jobID>[A-Z\d]+)$/i;
 
-export const getRemoteIDFromResponse = (
+export const getRemoteIDFromResponse = async (
   jobType: JobTypes,
-  response: AxiosResponse<string | { jobId: string }>
+  response: Response
+  // response: AxiosResponse<string | { jobId: string }>
```
dlrice, line 32: assuming this is no longer needed?
```suggestion
```

src/tools/utils/index.tsx
```diff
@@ -59,39 +59,39 @@ export const getRemoteIDFromResponse = (
...
-  `Job status: (?<status>${statuses.join('|')})`
-);
 
-export const getStatusFromResponse = (
+export const getStatusFromResponse = async (
   jobType: JobTypes,
-  response: AxiosResponse<Status | { jobStatus: Status } | string>
-): [status: Status, idMappingTarget?: IDMappingNamespace] => {
+  response: Response
+  // response: Status | { jobStatus: Status } | string
```
dlrice, line 66:
```suggestion
```

dlrice (approved):

src/tools/utils/index.tsx
aurel-l: Indeed this can be removed, sorry that this slipt through

PR #335, opened 2021-04-07
create dateParse helper function in order to avoid invalid dates

aurel-l:
## Purpose
Fix issue with some date not being parsed correctly and throwing error on some pages (/proteomes/UP000149066) in some browsers (Firefox and Safari)

## Approach
Create a wrapper `parseDate` around date parsing and handles some edge cases that we know of.
Makes sure that even if it cannot parse the date, it doesn't return an invalid date that would cause an issue further down the line.
Replace all bare uses of `new Date()` with the new wrapper.

## Testing
Added tests for wrapper (note that it's running on V8, so would have the same behaviour as Firefox or Safari).
Made sure the rest of tests were running, and that the problematic page was not crashing in Firefox

## Checklist

### DIFF
--- a/src/shared/utils/parseDate.ts
+++ b/src/shared/utils/parseDate.ts
@@ -1,0 +1,50 @@
+const edgeCase1 = /'\w{3}-\d{4}'/;
+
+const months = [
+  'jan',
+  'feb',
+  'mar',
+  'apr',
+  'may',
+  'jun',
+  'jul',
+  'aug',
+  'sep',
+  'oct',
+  'nov',
+  'dec',
+];
+
+/**
+ * Either returns a valid date, or undefined, but no invalid date
+ * @param input date to parse
+ * @returns {Date | undefined} valid date if possible, otherwise undefined
+ */
+const parseDate = (input?: string | number | Date | null): Date | undefined => {
+  if (!input) {
+    return undefined;
+  }
+  if (input instanceof Date) {
+    return input;
+  }
+  // easy case
+  let output = new Date(input);
+  if (!Number.isNaN(output.getTime())) {
+    return output;
+  }
+  /* edge cases */
+  // This one is specific to Firefox and Safari, as of April 2021.
+  // This will not be able to be covered by tests as node is using the V8 engine
+  if (typeof input === 'string' && edgeCase1.test(input)) {
+    const [month, year] = input.split('-');
+    const monthIndex = months.indexOf(month.toLowerCase());
+    output = new Date(+year, monthIndex);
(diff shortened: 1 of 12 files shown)


src/shared/utils/parseDate.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+  // easy case
+  let output = new Date(input);
+  if (!Number.isNaN(output.getTime())) {
+    return output;
+  }
+  /* edge cases */
+  // This one is specific to Firefox and Safari, as of April 2021.
+  // This will not be able to be covered by tests as node is using the V8 engine
+  if (typeof input === 'string' && edgeCase1.test(input)) {
```
dlrice, line 38: Could extract into a separate function (eg `parseDifficultDate`) to run unit tests in V8

src/shared/utils/parseDate.ts
```diff
@@ -0,0 +1,50 @@
...
+    const [month, year] = input.split('-');
+    const monthIndex = months.indexOf(month.toLowerCase());
+    output = new Date(+year, monthIndex);
+    if (monthIndex >= 0 && !Number.isNaN(output.getTime())) {
+      return output;
+    }
+  }
+  // otherwise just return undefined
+  return undefined;
```
dlrice, line 47: How about just returning the provided string as a fallback? I realise this will be a bit tricky as we call `toISOString` on the output of this fn in most places.  One idea then is creating a fn `parseAndFormatDate` which calls `parseDate` and if the return of this is `undefined` returns the provided string and otherwise immediately calls `toISOString`?

dlrice: A console.error that things didn't work out could be useful to find these unsuccessful instances in the future

src/shared/utils/parseDate.ts
aurel-l: Good idea, I did that, and that even allowed me to catch an issue in my regular expression 👍🏽

src/shared/utils/parseDate.ts
aurel-l: I added the console.error, it would be good to other edge cases eventually too.

aurel-l: About returning the provided string, I don't think it is necessary. In the places where we use a string as source of information for a data, we use display the unmodified string to the user. The date is only parsed as an extra information passed to the `<time>` tag shown to the user's assistive technology, or a bot browsing the page. That's also the reason why I thought it was OK to miss parsing some cases, because worst case scenario we still do display the original string to the user and the user "parses" the data in their mind, even if the browser doesn't recognise it as a date

dlrice (approved):

PR #337, opened 2021-04-08
2021 04 peptide search form feedback

aurel-l:
## Purpose
Handle feedback post-UX meeting about peptide search form

## Approach
Added a splitAndTidyText helper function

## Testing
Update snapshots, new tests

## Checklist

### DIFF
--- a/src/shared/hooks/useTextFileInput.ts
+++ b/src/shared/hooks/useTextFileInput.ts
@@ -1,9 +1,15 @@
-import { useEffect, useCallback, useRef, ReactElement } from 'react';
+import {
+  useEffect,
+  useCallback,
+  useRef,
+  ReactElement,
+  MutableRefObject,
+} from 'react';
 
 import useDragNDropFile from './useDragNDropFile';
 
 type UseTextFileInputProps = {
-  input: HTMLInputElement | null;
+  inputRef: MutableRefObject<HTMLInputElement | null>;
   onFileContent: (content: string) => void;
   onError?: (error: TypeError) => void;
   dndOverlay: ReactElement;
@@ -18,7 +24,7 @@ const writableRE = /[\w\s-=+*;><"'/\\.#:;|,]/gi;
  * of files on it and normal file handling on click, parse content
  */
 const useTextFileInput = ({
-  input,
+  inputRef,
   onFileContent,
   onError,
   dndOverlay,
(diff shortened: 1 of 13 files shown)


dlrice (commented):
A few behaviors in our forms that I think could be improved:

1. Should we disable the submit button until input is available in both mapping and pep search? At the moment mapping submit is always enabled. On peptide search it is enabled and then typing a single letter disables it. Typing an additional character then re-enables it.
2. In both forms when I type in some IDs/sequences the suggested name is populated. When the input IDs/sequences are removed the suggested name sticks around while I think it would make more sense to remove it.

aurel-l:
> A few behaviors in our forms that I think could be improved:
> 
>     1. Should we disable the submit button until input is available in both mapping and pep search? At the moment mapping submit is always enabled. On peptide search it is enabled and then typing a single letter disables it. Typing an additional character then re-enables it.

This is also the current behaviour with BLAST and Align too. I think the idea is that we don't want to display disabled buttons until the user actually starts playing with the form, and anyway, even if the submit button isn't disabled it isn't submitting when there is no data in the form.
Do you think we should double check that behaviour with UX?

> 
>     2. In both forms when I type in some IDs/sequences the suggested name is populated. When the input IDs/sequences are removed the suggested name sticks around while I think it would make more sense to remove it.

Good point, I'll modify it

dlrice:
```
This is also the current behaviour with BLAST and Align too. I think the idea is that we don't want to display disabled buttons until the user actually starts playing with the form, and anyway, even if the submit button isn't disabled it isn't submitting when there is no data in the form. Do you think we should double check that behaviour with UX?
```

Could be worth reviewing. Just realised it might also be worth telling the user why the button is disabled in this instance. It may not be clear to the user that they must have a certain number of characters. I know we report when input is invalid (eg wrong chars) but not when there is an insufficient number of chars. This would affect all of our forms so happy if you want to merge this now and then create a new branch when a decision has been made on that.

dlrice (approved):

aurel-l:
> ```
> This is also the current behaviour with BLAST and Align too. I think the idea is that we don't want to display disabled buttons until the user actually starts playing with the form, and anyway, even if the submit button isn't disabled it isn't submitting when there is no data in the form. Do you think we should double check that behaviour with UX?
> ```
> 
> Could be worth reviewing. Just realised it might also be worth telling the user why the button is disabled in this instance. It may not be clear to the user that they must have a certain number of characters. I know we report when input is invalid (eg wrong chars) but not when there is an insufficient number of chars. This would affect all of our forms so happy if you want to merge this now and then create a new branch when a decision has been made on that.

Added a point in the design review meeting document as backlog of discussion. Depending on what is decided we can create the corresponding Jira.

PR #338, opened 2021-04-08
supporting data starting point

aurel-l:
## Purpose
Starting point for supporting data entry pages. Doesn't aim to close any specific jira as none of them fit exactly with the mockups.
This is to build upon!

## Approach
Connect routes for supporting data URLs to components, load data, copy approach for supporting data result page cards, but with more columns rendered.
A lot of copy pasted code, but didn't want to generalise something now when specs are still moving but we need this to be ready yesterday!

## Testing
No specific unit-test for now.
Checked that the pages where rendering

## Checklist

### DIFF
--- a/src/supporting-data/database/components/entry/Entry.tsx
+++ b/src/supporting-data/database/components/entry/Entry.tsx
@@ -1,0 +1,68 @@
+import { RouteChildrenProps } from 'react-router-dom';
+import { Loader, Card } from 'franklin-sites';
+
+import SingleColumnLayout from '../../../../shared/components/layouts/SingleColumnLayout';
+import ErrorHandler from '../../../../shared/components/error-pages/ErrorHandler';
+import RenderColumnsInCard from '../../../../shared/components/results/RenderColumnsInCard';
+
+import useDataApi from '../../../../shared/hooks/useDataApi';
+
+import apiUrls from '../../../../shared/config/apiUrls';
+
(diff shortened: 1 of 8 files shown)


src/supporting-data/database/components/entry/Entry.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (loading || !data) {
+    return <Loader progress={progress} />;
+  }
+
+  return (
+    <SingleColumnLayout>
+      <h2>Database - {data.abbrev}</h2>
+      <Card>
+        <RenderColumnsInCard
+          renderers={DatabaseColumnConfiguration.get(DatabaseColumn.name)}
```
dlrice, line 44: A potentially premature comment: could simplify the code a bit here by mapping over `[DatabaseColumn.name, DatabaseColumn.server,...]` and calling `RenderColumnsInCard`.

dlrice (approved):

src/supporting-data/database/components/entry/Entry.tsx
aurel-l: Reused same code everywhere, but I'm expecting this to be changed/overriden when handling the "card content" subtask in each of the 6 supporting data jiras. Not sure RenderColumnsInCard is the best approach for these, as the mockup requirements are slightly different (for example, full blue bar on the left, or 2 columns within the card for some namespaces like taxonomy).

PR #341, opened 2021-04-13
more supporting data entry pages content using infolist

aurel-l:
## Purpose
Nicer content in supporting data entry pages

## Approach
- Use useDataApiWithStale (in order to avoid flash when jumping across entries of same type)
- Use InfoList
- Modify some column renderers in order to return undefined when no data (so that InfoList doesn't display anything)
- Move is-stale classes to a helper file in CSS module

## Testing
Updated snapshots

## Checklist

Anything not covered by this PR has had a Jira created in https://www.ebi.ac.uk/panda/jira/browse/TRM-25517 (that would be for more complex/visualisation bits)

### DIFF
--- a/src/supporting-data/taxonomy/config/TaxonomyColumnConfiguration.tsx
+++ b/src/supporting-data/taxonomy/config/TaxonomyColumnConfiguration.tsx
@@ -143,7 +143,12 @@ TaxonomyColumnConfiguration.set(TaxonomyColumn.strain, {
   render: ({ strains }) =>
     strains?.length && (
       <ExpandableList descriptionString="strains" displayNumberOfHiddenItems>
-        {strains.map((strain) => strain.name)}
+        {strains.map((strain) => (
+          <>
+            {strain.name}
+            {strain.synonyms?.length && ` (${strain.synonyms.join(', ')})`}
+          </>
+        ))}
       </ExpandableList>
     ),
 });
--- a/src/supporting-data/taxonomy/config/TaxonomyColumnConfiguration.tsx
+++ b/src/supporting-data/taxonomy/config/TaxonomyColumnConfiguration.tsx
@@ -125,7 +125,7 @@ TaxonomyColumnConfiguration.set(TaxonomyColumn.otherNames, {
 TaxonomyColumnConfiguration.set(TaxonomyColumn.parent, {
   label: 'Parent',
   render: ({ parentId }) =>
-    parentId && <Link to={getEntryPath(parentId)}>{parentId}</Link>,
+    parentId ? <Link to={getEntryPath(parentId)}>{parentId}</Link> : undefined,
 });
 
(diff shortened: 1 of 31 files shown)


xwatkins:
Functionality review below (haven't looked at the code yet). You're probably aware of some of them, if so please make Jiras explaining the issue(s).

## Taxonomy
1. Taxonomy root node seems to display a rogue '0' in the right column `taxonomy/1`
2. We need to have some kind of "taxonomy navigation". Also do we only get the parent taxid, not the name?

## Diseases
1. There's already a disclaimer at the bottom of uniprotkb entries, could be worth using the same styling for both (whatever that is!)
2. Cross-references should be links

## Sub-cell
1. References are not on the old website (`SL-0002` for instance), is this new data? If so I guess it should be urls?

## Database
1. Category should be a link to a query

aurel-l:
> ## Taxonomy
> 
>     1. Taxonomy root node seems to display a rogue '0' in the right column `taxonomy/1`
Fixed
> 
>     2. We need to have some kind of "taxonomy navigation". Also do we only get the parent taxid, not the name?
 - Isn't that what the lineage is for?
 - Going down, we don't have children data
 - Yes, just the tax ID for now, was part of that feedback document, hopefully we'll have the object whenever the change is deployed.
> 
> ## Diseases
> 
>     1. There's already a disclaimer at the bottom of uniprotkb entries, could be worth using the same styling for both (whatever that is!)
Created share component and used it
> 
>     2. Cross-references should be links
Commented before
> 
> ## Sub-cell
> 
>     1. References are not on the old website (`SL-0002` for instance), is this new data? If so I guess it should be urls?
Commented before (but could remove it to avoid asking backend for changes now)
> 
> ## Database
> 
>     1. Category should be a link to a query
Changed

src/supporting-data/taxonomy/config/TaxonomyColumnConfiguration.tsx
```diff
@@ -115,17 +111,21 @@ TaxonomyColumnConfiguration.set(TaxonomyColumn.mnemonic, {
...
   label: 'Other names',
-  render: ({ otherNames }) => (
-    <ExpandableList descriptionString="other names" displayNumberOfHiddenItems>
-      {otherNames}
-    </ExpandableList>
-  ),
+  render: ({ otherNames }) =>
+    otherNames?.length && (
+      <ExpandableList
+        descriptionString="other names"
```
xwatkins, line 117: "More/Less other names" sounds a bit funny. Maybe just "More/Less names"?

aurel-l, line 117:
```suggestion
        descriptionString="names"
```

aurel-l: Changed

xwatkins (approved):

PR #342, opened 2021-04-13
2020_05-query_hook

xwatkins:
## Purpose
Better separation of concerns with custom hooks handling query string and columns

## Approach
Created custom hooks.

## Testing
Had to skip one of the ResultsContainer tests as it was hanging (seems like the loader gets displayed and the results don't appear so the await on element selector would not resolve). Decided to skip instead of fixing as tests will need an overhall post refactor, and these tests will likely move to the facet component.

## Checklist


src/shared/hooks/useNSQuery.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { getParamsFromURL } from '../../uniprotkb/utils/resultsUtils';
+import { getAPIQueryUrl } from '../config/apiUrls';
+import { Column } from '../config/columns';
+
+import useNS from './useNS';
+
+const useNSQuery = ({
+  queryColumns,
+  facets,
```
aurel-l, line 12: Couldn't we get those 2 from inside the hook?

aurel-l: facets are derived from the location, right? And queryColumns from the namespace and the userPreference

xwatkins: - `queryColumns` also depend on the `ViewMode` so I wanted to wait until later to see if it can be moved.
- `facets`, I seem to have somehow completely lost in the refactor, having a look!

aurel-l: >     * `queryColumns` also depend on the `ViewMode` so I wanted to wait until later to see if it can be moved.
Yep, that's what I meant when I said userPreference, so you could call that hook in here too and that should give that information to you.

src/shared/components/results/ResultsView.tsx
```diff
@@ -224,107 +154,46 @@ const cardRenderer = (
...
-    namespace,
-    query,
-    // TODO: Do similar things for the rest of namespaces
-    columns: queryColumns,
-    selectedFacets,
-    // Not really interested in the facets here, so try to reduce payload
-    facets: null,
-    sortColumn,
-    sortDirection,
+  const { url: initialApiUrl, direct } = useNSQuery({
```
dlrice, line 193: When switching from table to card view initialApiUrl is updated however it's not getting into this `useEffect`:
```
  useEffect(() => {
    setAllResults([]);
    setMetaData({ total: 0, nextUrl: undefined });
    setUrl(initialApiUrl);
  }, [initialApiUrl]);
```
to reset and fetch all of the columns needed for the card view.

src/shared/hooks/useNSQuery.ts
xwatkins: Ah yes of course! I weirdly had assumed it was in the state, not the user preferences

src/shared/components/results/ResultsView.tsx
xwatkins: I've simplified the "hacky logic" to force a re-render of the loader, and I think this is now addressed. @aurel-l could you have a look if you don't mind as I think you'd added this logic in the first place?

aurel-l: Using the initial URL to detect when to force a refresh is good, you should also be able to drop the `prevViewMode.current !== viewMode` bit too I believe.

xwatkins: I tried that and there's a race condition and the `useEffect` in `useNSQuery.ts` hasn't run by the time it gets to the check. Not sure there's an easy way around that

src/shared/hooks/useNSQuery.ts
```diff
@@ -0,0 +1,76 @@
...
+import { ViewMode } from '../components/results/ResultsContainer';
+import { Namespace } from '../types/namespaces';
+
+const useNSQuery = ({
+  size,
+  withFacets = true,
+}: {
+  size?: number;
+  withFacets?: boolean;
+}) => {
```
aurel-l, line 21:
```suggestion
} = {}) => {
```
Would avoid requiring you to pass pass an empty object where you use this hook

xwatkins: 👍  done

aurel-l (approved):
No other comment from me, but I just had a look at the code here, haven't run it.
I guess @dlrice 's review is more important as this overlaps his work.

dlrice (approved):

PR #343, opened 2021-04-13
Add pagination hook

xwatkins:
## Purpose
Extract logic to do with handling pagination

## Approach
Create a custom hook

## Testing
N/A

## Checklist

### DIFF
--- a/src/shared/components/results/ResultsView.tsx
+++ b/src/shared/components/results/ResultsView.tsx
@@ -33,6 +33,7 @@ import useNS from '../../hooks/useNS';
 import useNSQuery from '../../hooks/useNSQuery';
 import useUserPreferences from '../../hooks/useUserPreferences';
 import useColumns from '../../hooks/useColumns';
+import usePagination from '../../hooks/usePagination';
 
 import { getIdKeyFor } from '../../utils/getIdKeyForNamespace';
 import { getEntryPathFor } from '../../../app/config/urls';
@@ -43,7 +44,6 @@ import { ViewMode } from './ResultsContainer';
 
 import './styles/warning.scss';
 import './styles/results-view.scss';
-import usePagination from '../../hooks/usePagination';
 
 const cardRenderer = (
   namespace: Namespace,
--- a/src/shared/components/results/ResultsView.tsx
+++ b/src/shared/components/results/ResultsView.tsx
@@ -166,7 +166,7 @@ const ResultsView: FC<ResultsTableProps> = ({
   const { url: initialApiUrl, direct } = useNSQuery();
   const {
     allResults,
-    loading,
+    initialLoading,
     handleLoadMoreRows,
     hasMoreData,
     progress,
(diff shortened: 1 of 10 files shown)


src/shared/components/results/ResultsView.tsx
```diff
@@ -204,27 +191,15 @@ const ResultsView: FC<ResultsTableProps> = ({
...
-      total: +(headers?.['x-totalrecords'] || 0),
-      nextUrl: getNextURLFromHeaders(headers),
-    }));
-  }, [data, headers]);
+  }, [history, direct, hasMoreData, allResults, getEntryPathForEntry]);
 
   if (
     // if loading the first page of results
-    (loading && url === initialApiUrl) ||
+    loading ||
```
aurel-l, line 202: Losing the other part of the test isn't it causing the page to flash on new data being added in? As far as I can see the current version of the hook sets loading to true when any new network call is being made.

xwatkins: I had compared it with https://ebi-uniprot.github.io/uniprot-website and didn't notice a difference...although there is a bit of a jump there too, which I guess is because of the "jump to top" PR?

aurel-l: What I'm talking about is when scrolling down to get more data, that shouldn't interact with the jump-to-top functionality because there is no client URL change.
If there is no flash when scrolling down to get more data then all good

xwatkins: Actually if I slow the network down the results disappear indeed (reason I was talking about the jump to top is that because it does that jump it was not noticeable - but we should fix that too).
Actually because the infinite scroll table handles the display of the loader in that case, I'm thinking to change "loading" to "initialLoading" in the hook and have the url checking logic there. What do you reckon?

aurel-l: `initialLoading` sounds good 👍🏽 

Regarding the jump to top issue that you are referring to, I can't seem to see it, maybe I'm not looking at the right place. Do you mind sending me a video of the issue?

xwatkins: done

src/shared/components/results/ResultsView.tsx
```diff
@@ -46,6 +43,7 @@ import { ViewMode } from './ResultsContainer';
 
 import './styles/warning.scss';
 import './styles/results-view.scss';
+import usePagination from '../../hooks/usePagination';
```
aurel-l, line 46: Can you move this up with the rest of the custom hooks?

xwatkins: Damn, I really need to write something to check my imports order 😬

aurel-l: I mean, I believe that's what the `I have reviewed my own code` bit is for. For me I do this by having a look at the GitHub side-by-side diff view that you get even during the process of creating a PR and go line by line to see if there's anything I've missed.

aurel-l (approved):

xwatkins:
Ended up removing react-spring. Tested the message popup by adding a `setInterval` and it looked similar.

aurel-l:
> Ended up removing react-spring. Tested the message popup by adding a `setInterval` and it looked similar.

From a quick look it looks like you haven't cover the sliding panel cases when coming from the side (Download, QueryBuilder)

aurel-l (approved):

PR #344, opened 2021-04-14
2021 05 refactor results container

dlrice:
## Purpose
[ResultsContainer change location of API calls](https://www.ebi.ac.uk/panda/jira/browse/TRM-25832)

## Approach
- `Results`: holds `ResultsData` and `ResultsFacets`
- `ResultsData`: actual table/cards of results. All related fetching now happens in here.
- `ResultsFacets`: All related fetching now happens in here.
- `resultsCardRenderers`: A config file that returns the correct card renderer for the namespace
- `SideBarLayout`: Removed action buttons as this was only used for results
- Get `total` from `usePagination`

## Testing
- Updated existing
- Updated download to remove act warnings

## Checklist


src/shared/components/results/ResultsFacets.tsx
```diff
@@ -1,21 +1,52 @@
...
+    headers,
+    status,
+    isStale,
+  } = useDataApiWithStale<Response['data']>(initialApiUrl);
+
+  if (error || !(loading || data) || !namespace) {
+    return <ErrorHandler status={status} />;
+  }
+
+  const total = headers?.['x-totalrecords']
```
xwatkins, line 37: If there's 0 results there should be no facets as they are calculated based on the results set. So this could be removed.

aurel-l: Is this comment in the right place in the code?

xwatkins: L37 - I'm wondering whether the check of `total` L45 is necessary and if not we could simplify the logic a bit.

src/shared/components/results/ResultsData.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+const ResultsData: FC = () => {
+  const namespace = useNS() || Namespace.uniprotkb;
+  const [viewMode] = useUserPreferences<ViewMode>('view-mode', ViewMode.CARD);
+  const history = useHistory();
+  const [columns, updateColumnSort] = useColumns();
+  const [selectedEntries, handleEntrySelection] = useItemSelect();
+  const { search: queryParamFromUrl } = useLocation();
+  const { query, selectedFacets, sortColumn, sortDirection } = getParamsFromURL(
```
xwatkins, line 45: Could this be done in `ResultsButtons` instead?

src/shared/components/results/ResultsFacets.tsx
dlrice: When there are no results the facets still exist:
```
[
  {
    "label": "Status",
    "name": "reviewed",
    "allowMultipleSelection": false
  },
  {
    "label": "Model organisms",
    "name": "model_organism",
    "allowMultipleSelection": true
  },
  {
    "label": "Proteins with",
    "name": "proteins_with",
    "allowMultipleSelection": true
  },
  {
    "label": "Protein Existence",
    "name": "existence",
    "allowMultipleSelection": true
  },
  {
    "label": "Annotation Score",
    "name": "annotation_score",
    "allowMultipleSelection": true
  },
  {
    "label": "Sequence length",
    "name": "length",
    "allowMultipleSelection": true,
    "values": [
      {
        "label": "1 - 200",
        "value": "[1 TO 200]",
        "count": 0
      },
      {
        "label": "201 - 400",
        "value": "[201 TO 400]",
        "count": 0
      },
      {
        "label": "401 - 600",
        "value": "[401 TO 600]",
        "count": 0
      },
      {
        "label": "601 - 800",
        "value": "[601 TO 800]",
        "count": 0
      },
      {
        "label": ">= 801",
        "value": "[801 TO *]",
        "count": 0
      }
    ]
  }
]
```
We could try to interpret this object to decide if there are not facets but I think it's easier to condition on `total` as it's immediately available.

xwatkins: Ah of course! 😭

src/shared/components/results/ResultsData.tsx
dlrice: Good point, I did that then realised it could be moved further down to Download. Updated.

dlrice (commented):
In my testing this morning I'm seeing some issues that I still need to address in this PR:

1. Changing from one namespace to another will sometimes cause it to have all the data it needs to render a card.
2. If I leave the tab open at the results page and come back in 10 min I see an error:
```
Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

src/shared/components/download/__tests__/Download.spec.tsx
```diff
@@ -95,19 +95,23 @@ describe('Download component', () => {
...
       if (columnSelect) {
-        expect(customise).toBeInTheDocument();
+        await waitFor(() => {
+          expect(customise).toBeInTheDocument();
+        });
       } else {
-        expect(customise).not.toBeInTheDocument();
+        await waitFor(() => {
+          expect(customise).not.toBeInTheDocument();
+        });
```
aurel-l, line 109: This (`waitFor` + `not.toBe`) feels a bit strange. I think that if we need to wait because something is loading, we need to check for another signal which is not the thing that we're going to do the assertion on.
Either the `waitFor` is unnecessary, or we need to wait for an update of the state of the component before getting the `Customize data` (through `act`, or by `waitFor` that the selected value is the one we changed it for).

dlrice: Good catch - it wasn't needed. I got carried away with the `waitFor`s.

aurel-l (approved):
For me it's good to go.

I still see the warning about setting the state multiple times even after all these refactors so please have a look at https://github.com/ebi-uniprot/uniprot-website/pull/348 and see if the approach looks valid and if the error still happens (for me it doesn't anymore, but it's hard to pinpoint anyway)

xwatkins:
> For me it's good to go.
> 
> I still see the warning about setting the state multiple times even after all these refactors so please have a look at #348 and see if the approach looks valid and if the error still happens (for me it doesn't anymore, but it's hard to pinpoint anyway)

Will check #348 as soon as I've made sense of my ID mapping results branch. @dlrice do you want to do the merge into `main`?

PR #347, opened 2021-04-20
optimise job polling

aurel-l:
## Purpose
Optimise job polling

## Approach
Do not schedule a new job polling loop if there are no jobs that need to be checked

## Testing
Made sure the job runner was still checking newly created jobs, and running jobs even after a refresh of the page

## Checklist


src/tools/state/toolsMiddleware.ts
```diff
@@ -39,16 +39,25 @@ const toolsMiddleware: Middleware<Dispatch<AnyAction>, RootState> = (store) => {
...
 
     // reset flag
     pollJobs.scheduled = false;
 
     await sleep(POLLING_INTERVAL);
-    // eslint-disable-next-line @typescript-eslint/no-use-before-define
-    pollJobs.schedule();
+
+    jobsToCheck = Object.values(getState().tools).filter(
+      (job) => job.status === Status.CREATED || job.status === Status.RUNNING
```
dlrice, line 54: A small stylistic comment: how about extracting a fn here and calling on L42 and L53

aurel-l: Good point, I updated the code

dlrice (approved):

PR #348, opened 2021-04-20
new approach to useUserPreferences

aurel-l:
## Purpose
Rewrite useUserPreferences to avoid crazy amount of calls to setState

## Approach
Removed use of one global state in react's context. Removed async wrappers around calls to localStorage's setItem and getItem.
Note that I'm not sure if there's gonna be an impact to calling setItem and getItem synchronously without scheduling them, but if it solves the setState issue I guess it's better. We need to make sure we limit the amount of time we call these 2 methods to the bare minimum

## Testing
Updated all tests using it, and the hook's tests

## Checklist


xwatkins (commented):
LGTM

src/shared/components/download/__tests__/Download.spec.tsx
```diff
@@ -35,6 +35,10 @@ describe('Download component', () => {
   let onCloseMock;
 
   beforeEach(() => {
+    window.localStorage.setItem(
```
xwatkins, line 38: Thinking out loud here, but wondering if this could be done in `customRender` and the `window.localStorage.clear();` triggered when the Wrapper component unmounts (through a `useEffect` return maybe?).

xwatkins (approved):
Just a suggestion which might not be possible, approved otherwise

src/shared/components/download/__tests__/Download.spec.tsx
aurel-l: `useEffect` wouldn't work because it happens after the first render. But I liked the idea and modified it. Basically this way the customRender keeps the same signature than before the refactor, but we set the localStorage in the Wrapper's constructor (yes, had to use classes, as there's no hook for logic _before_ the first render), and cleanup in `componentWillUnmount()`.
Now in latest commit

xwatkins: Shame about the cleanup but it does look neater 👍

xwatkins: Oh actually yeah cleanup is good too 😆

PR #350, opened 2021-04-23
2021 05 peptide search result

dlrice:
## Purpose
Create peptide search results [jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-25641)

## Approach
Use and augment newly refactored hooks.

Todo:
1. Test
2. Fix download of results

## Testing
Not yet

## Checklist

### DIFF
--- a/src/app/components/App.tsx
+++ b/src/app/components/App.tsx
@@ -138,12 +138,12 @@ const IDMappingForm = lazy(
     )
 );
 
-// const PeptideSearchResult = lazy(
-//   () =>
-//     import(
-//       /* webpackChunkName: "peptide-search-result" */ '../../tools/peptide-search/components/results/PeptideSearchResult'
-//     )
-// )
+const PeptideSearchResult = lazy(
+  () =>
+    import(
+      /* webpackChunkName: "peptide-search-result" */ '../../tools/peptide-search/components/PeptideSearchResult'
+    )
+);
 
 const PeptideSearchForm = lazy(
   () =>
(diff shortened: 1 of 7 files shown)


aurel-l:
Looks good so far.

I would customise the header maybe, in order to display that it's a Peptide Search, either instead, or additionally to the current big `UniProtKB` title.
Maybe `ResultsDataHeader` should have a `children` ReactNode prop in order to customise a bit the title. In this case you could pass `found in peptide search ${jobId}` and it would render `UniProtKB 14 results found in peptide search 38984udduf894u `.
This could be reused in ID mapping and in the future map to section at the bottom of the supporting data entry pages (see existing mockup for this).

Do check with @xwatkins if your changes overlap with his

xwatkins:
> Looks good so far.
> 
> I would customise the header maybe, in order to display that it's a Peptide Search, either instead, or additionally to the current big `UniProtKB` title.
> Maybe `ResultsDataHeader` should have a `children` ReactNode prop in order to customise a bit the title. In this case you could pass `found in peptide search ${jobId}` and it would render `UniProtKB 14 results found in peptide search 38984udduf894u `.
> This could be reused in ID mapping and in the future map to section at the bottom of the supporting data entry pages (see existing mockup for this).
> 
> Do check with @xwatkins if your changes overlap with his

This sounds good, although ID Mapping also needs a container to display the list of unmatched IDs (guess this could be passed in the children too) and the buttons are different when there is no namespace (Download only).

PR #353, opened 2021-04-28
2021 05 map to button

aurel-l:
## Purpose
Add map to/view proteins pattern to supporting data entry pages and search pages ([TRM-25729](https://www.ebi.ac.uk/panda/jira/browse/TRM-25729))

## Approach
 - Using statistics object
 - Temporarily use a facet query for higher-level taxons (can be removed in the future when we have the data)
 - Update the Card component in Franklin and update its usage everywhere
 - Fix/change heading levels on the components touched
 - Higher typescript and linting coverage on test files (added whole UniRef and UniParc)
 - Added back the references in the UniProtKB publication pages (was lost at some point in previous refactor when trying tho use the CitationsCard everywhere)
 - Removed al lot of `FC` when we didn't use children (that's why we didn't see the issue in the previous point)
 - Fixed the main search selector that was not switching when navigating to another namespace ([TRM-25557](https://www.ebi.ac.uk/panda/jira/browse/TRM-25557))

## Testing
Update existing tests and snapshots

## Checklist


src/app/components/styles/app.scss
```diff
@@ -15,3 +17,24 @@
 .loader-container + .loader-container {
   display: none;
 }
+
+.card__header {
+  padding-top: $global-padding;
```
xwatkins, line 22: Should this be defined in Franklin instead?

src/proteomes/components/entry/Publications.tsx
```diff
@@ -1,28 +1,36 @@
...
+import { Namespace } from '../../../shared/types/namespaces';
+
+const getIdKey = getIdKeyFor(Namespace.citations);
+
 const Publications: FC<Pick<ProteomesAPIModel, 'citations'>> = ({
   citations,
 }) => {
-  if (!citations?.length) {
+  const data = useMemo<CitationsAPIModel[]>(
+    () => citations?.map((citation) => ({ citation })),
```
xwatkins, line 18: Could you please add a comment in the code to explain why this step is needed?

src/supporting-data/citations/components/LiteratureCitation.tsx
```diff
@@ -1,11 +1,14 @@
-import { useState, FC, Fragment } from 'react';
+import { createElement, useState, FC, Fragment, HTMLAttributes } from 'react';
```
xwatkins, line 1: image
Not sure when this was lost, but the "Cited for/Source etc.." section on the left used to be right under the "view abstract" (to gain a bit of vertical space)

xwatkins (commented):
Are backend aware of the count discrepancy e.g. for http://localhost:8080/citations/17062754 (13 mapped, only 2 in results)

.eslintrc
```diff
@@ -74,6 +74,7 @@
     "react/no-did-update-set-state": "off",
     "react/prop-types": "off",
     "react/react-in-jsx-scope": "off",
+    "react/require-default-props": "off",
```
xwatkins, line 77: I guess with TS we don't need it?

src/uniref/types/entrySection.ts
```diff
@@ -21,7 +21,7 @@ export const getEntrySectionNameAndId = (
       };
     case EntrySection.Members:
       return {
-        name: 'Members',
+        name: 'members',
```
xwatkins, line 24: if there's no way to conditionally add the 's' if more than 1 member, this should say "member(s)"

xwatkins (commented):
The problem with the new approach is that we have to manually define the header level in each of the entry page sections which could lead to inconsistencies. I guess it's ok in most cases because you can see that the following titles are `h3` but could anything stop someone from adding a new section and accidentally starting at level `h3` instead of `h2`?  Aside from having a `EntrySectionHeader` content which would just return `<h2>{children}</h2>`

src/app/components/styles/app.scss
aurel-l: Actually I had forgotten to make it more specific, it is now covering headers but only containing small text

src/proteomes/components/entry/Publications.tsx
aurel-l: Added a comment

src/supporting-data/citations/components/LiteratureCitation.tsx
aurel-l: Not really corresponding to that part of the code, but I went and changed lower in this file where the child was injected.
Also, didn't really know beforehand where it was supposed to go because I didn't remember seeing it in the past 😅

src/uniref/types/entrySection.ts
aurel-l: Changed it to say member, and append "s" in the same bit where I prepended to count

aurel-l:
> Are backend aware of the count discrepancy e.g. for http://localhost:8080/citations/17062754 (13 mapped, only 2 in results)

I raised it again, not sure if they added solved this everywhere or not already (if they did, then there's a bug somewhere).

> The problem with the new approach is that we have to manually define the header level in each of the entry page sections which could lead to inconsistencies. I guess it's ok in most cases because you can see that the following titles are h3 but could anything stop someone from adding a new section and accidentally starting at level h3 instead of h2? Aside from having a EntrySectionHeader content which would just return `<h2>{children}</h2>`

I would say there was a bigger problem with the previous approach anyway, because we couldn't change the heading level at all. At least now we have control. There is no magic way to "remember" which heading level to put where apart from seeing it in context (note that I've added a helper class in a CSS file to uncomment when we're debugging this).
We could have a "magic" `<Heading>` component at some point which would decide which heading level to render where depending on its context (probably would need to use React's Context too), but that won't be straightforward and that's way out of the scope of this PR.
Deciding the hierarchy of headings is a manual task that needs to be done by a human anyway, so we need to get used to it. Heading levels depend on content and context together. Humans need to decide which heading to put where.

xwatkins:
> > Are backend aware of the count discrepancy e.g. for http://localhost:8080/citations/17062754 (13 mapped, only 2 in results)
> 
> I raised it again, not sure if they added solved this everywhere or not already (if they did, then there's a bug somewhere).
> 
> > The problem with the new approach is that we have to manually define the header level in each of the entry page sections which could lead to inconsistencies. I guess it's ok in most cases because you can see that the following titles are h3 but could anything stop someone from adding a new section and accidentally starting at level h3 instead of h2? Aside from having a EntrySectionHeader content which would just return `<h2>{children}</h2>`
> 
> I would say there was a bigger problem with the previous approach anyway, because we couldn't change the heading level at all. At least now we have control. There is no magic way to "remember" which heading level to put where apart from seeing it in context (note that I've added a helper class in a CSS file to uncomment when we're debugging this).
> We could have a "magic" `<Heading>` component at some point which would decide which heading level to render where depending on its context (probably would need to use React's Context too), but that won't be straightforward and that's way out of the scope of this PR.
> Deciding the hierarchy of headings is a manual task that needs to be done by a human anyway, so we need to get used to it. Heading levels depend on content and context together. Humans need to decide which heading to put where.

I'd be against a magic component but would prefer some sort of safeguard regarding the hierarchy of titles on the page, which your debug codes gives us but it would be good to have something automated in the future to flag any unwanted hierarchy "jumps" - maybe as part of puppeteer reporting, dunno 🤷

aurel-l:
> > > Are backend aware of the count discrepancy e.g. for http://localhost:8080/citations/17062754 (13 mapped, only 2 in results)
> > 
> > 
> > I raised it again, not sure if they added solved this everywhere or not already (if they did, then there's a bug somewhere).
> > > The problem with the new approach is that we have to manually define the header level in each of the entry page sections which could lead to inconsistencies. I guess it's ok in most cases because you can see that the following titles are h3 but could anything stop someone from adding a new section and accidentally starting at level h3 instead of h2? Aside from having a EntrySectionHeader content which would just return `<h2>{children}</h2>`
> > 
> > 
> > I would say there was a bigger problem with the previous approach anyway, because we couldn't change the heading level at all. At least now we have control. There is no magic way to "remember" which heading level to put where apart from seeing it in context (note that I've added a helper class in a CSS file to uncomment when we're debugging this).
> > We could have a "magic" `<Heading>` component at some point which would decide which heading level to render where depending on its context (probably would need to use React's Context too), but that won't be straightforward and that's way out of the scope of this PR.
> > Deciding the hierarchy of headings is a manual task that needs to be done by a human anyway, so we need to get used to it. Heading levels depend on content and context together. Humans need to decide which heading to put where.
> 
> I'd be against a magic component but would prefer some sort of safeguard regarding the hierarchy of titles on the page, which your debug codes gives us but it would be good to have something automated in the future to flag any unwanted hierarchy "jumps" - maybe as part of puppeteer reporting, dunno 🤷

In this case what you want is already part of Lighthouse testing, in the accessibility section.

src/supporting-data/citations/components/LiteratureCitation.tsx
xwatkins: Sorry I wasn't clear (and without having seeing it before, you're coding in the dark!) - this is how it should be (still left column but right under "View abstract"
image

aurel-l: Ah I see, I put it back below the abstract then

xwatkins (approved):

PR #354, opened 2021-04-29
Console.log git information for build.

dlrice:
## Purpose
Show git info for sanity check/debugging for deployed sites.

## Approach
Created some new variables in webpack and console.log in index.tsx

## Testing
None

## Checklist


aurel-l (approved):
Works fine in my setup.
Not sure if we'll ever get a non-clean state because this is only injected for builds (I mean, if we do, it would be really worrying, so it's good to have it).
Will be useful to debug issues, possible could use it to prepulate bug reports from users and automatic analytics reports.

xwatkins (commented):
Just one thing, wondering whether this should use `console.debug` instead of log as it isn't visible by default

PR #355, opened 2021-04-29
clean text

aurel-l:
## Purpose
Clean up text imported before injecting in the page

## Approach
Use `sanitize-html` library, wrap a helper function around it with our sensible defaults

## Testing
Added tests

## Checklist


src/shared/utils/cleanText.ts
```diff
@@ -0,0 +1,43 @@
...
+import sanitizeHtml, {
+  defaults,
+  simpleTransform,
+  IOptions,
+} from 'sanitize-html';
+
+import styles from './styles/clean-text.module.scss';
+
+// List of tags to remove from the library default accepted set
+const blacklist = new Set(['a']);
```
dlrice, line 14:
```suggestion
const excludedTags = new Set(['a']);
```
I always confuse `blacklist` and `whitelist` - could we use `excludedTags`?

dlrice (approved):

src/shared/utils/cleanText.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const cleanTextDefaultOptions: IOptions = {
+  // https://github.com/apostrophecms/sanitize-html/blob/main/index.js#L691-L710
+  allowedTags: defaults.allowedTags.filter((tag) => !blacklist.has(tag)),
```
dlrice, line 18:
```suggestion
  allowedTags: defaults.allowedTags.filter((tag) => !excludedTags.has(tag)),
```

src/shared/utils/cleanText.ts
aurel-l: Oh yeah, sounds good. I'd always wondered if these terms had similar bad connotations as the "master" branch naming for example. This avoids this issue altogether 👍🏽

PR #356, opened 2021-04-29
2021 06 lighthouse report

aurel-l:
Add Lighthouse CI to project

Work in progress

Notes and examples to potentially follow:
 - https://medium.com/tuimm/measure-performance-web-app-with-lighthouse-ci-in-a-gitlab-pipeline-dd292842e40d


.gitlab-ci.yml
```diff
@@ -3,6 +3,7 @@ stages:
   - build_static_assets
   - build_docker_image
   - deploy
+  - lighthouse
```
dlrice, line 6: If lighthouse fails I'm assuming we don't want to deploy? If so, `lighthouse` wants to go above the `deploy` stage

.gitlab-ci.yml
```diff
@@ -88,3 +89,20 @@ deploy-fallback-only:
...
+      - main
+  stage: lighthouse
+  before_script:
+    - git --version
+    - node --version
+    - npm --version
+    - yarn --version
+    - yarn
+  script:
+    # Could we use the artifacts from before instead of rebuilding?
```
dlrice, line 106: Yeah, you're right. From the `build_static_assets` job:
```
  artifacts:
    paths:
      - build/
```
`build` directory will have it there

.lighthouserc.js
```diff
@@ -0,0 +1,30 @@
+// https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md
+module.exports = {
+  ci: {
+    collect: {
+      numberOfRuns: 1,
+      // Replace with correct URLs depending on context
+      url: [
+        'https://beta.uniprot.org',
```
dlrice, line 8: This is probably my lack of knowledge - are we testing the build or the actual live site?

aurel-l: To run the full audit, we have to test against an actual deploy of the website, could be internal URL though (available through some Gitlab environment variable?).
This is because it's also testing server settings (like server compression, and stuff like that).
That's why I thought it should run only on the main branch, *after* it has been deployed.

We could have an other subset of tests (like for accessibility and some of the tests that are not depending on the server) to run on a localhost server just running for the job. That could allow us to test at least a subset of audits and make the build fail if this subset fails.

So yeah, basically I think we'd need 2 kinds of lighthouse audit, full, or subset.

.gitlab-ci.yml
```diff
@@ -88,3 +89,20 @@ deploy-fallback-only:
   extends: .kubectl
   script:
     - ./deploy/deploy.sh HX # fallback
+
+lightouse:
+  image: cypress/browsers:node14.15.0-chrome86-ff82
+  only:
+    refs:
+      - main
```
dlrice, line 97: Are lighthouse tests expensive? If so do we want to run them at some regular interval rather than on each push to main? If so, we could create a branch `lighthouse` and push to this whenever we want a test or maybe [schedule it](https://docs.gitlab.com/ee/ci/pipelines/schedules.html).

.gitlab-ci.yml
aurel-l: So for what I said in https://github.com/ebi-uniprot/uniprot-website/pull/356#discussion_r623153110 I'd say:
```
  - test (<- include lighthouse_subset, all commits)
  - build_static_assets
  - build_docker_image
  - deploy
  - lighthouse (<- lighthouse full on the deployed version, only on main)
```

.gitlab-ci.yml
aurel-l: I think the only thing we'd test by running them regularly is if the API is down, which is not really what we're supposed to test for frontend, they have their own monitoring.
Associating lighthouse runs to specific commits should make it easier to track when (as in "in which specific commit") we added a library that slowed down the page for example.
Also, eventually, I'd like to have something like this set-up https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/server.md Which works better when associated with specific commits

.lighthouserc.js
xwatkins: It would be good to test against the internal URL as this would mean we wouldn't test the EBI routing infrastructure as part of the tests.
Good idea about the subset, although we'd have to carefully chose what we want to include in there.

aurel-l: To be fair, I think the point of some of the audits here is to test as close as possible to what a user would see, so the "full lighthouse" I believe should be run on the public-facing URL. Because if we get a "First Contentful Paint" of "1.5s" internally we might think it's fine, but if for a real user it's "20s" we have a problem to tackle, either with the EBI routing, or optimising even more on our side, or both.
But I'd say we'd want to know for sure!

xwatkins: While it's true that we need to test the "whole experience" maybe it is not something we want to check with every push to master, while something we did in the code which would affect the performance is.

.gitlab-ci.yml
dlrice: I'm sure we can know which commit a scheduled lighthouse test was run against. We would however lose specificity because we will then have a set of commits that could have caused a loss in performance (or the server health at that point in time).

PR #357, opened 2021-04-30
remove operator for first clause

aurel-l:
## Purpose
Remove the first operator of the query builder [TRM-21994](https://www.ebi.ac.uk/panda/jira/browse/TRM-21994)

## Approach
If it's the first term, remove the logical operator selector, and put an empty span filling the space to keep everything aligned.
Set the underlying state to "AND", happens also if the 2nd clause was "OR" and the first clause gets removed, the value gets changed to "AND" internally

## Testing
 - Updated snapshot, added snapshot with one clause, and another one with 2.
 - Activated TypeScript checks for the whole `src/query-builder` codebase subtree and fixed all the type issues in the tests. Also tried to follow more the guidelines to write tests

## Checklist


xwatkins:
Just eyeballed it but looking at the old website it looks like "NOT" is still a valid option for the first operator.

aurel-l:
Updated

PR #360, opened 2021-05-05
2021_05-idmapping-facets

xwatkins:
## Purpose
Allow filtering from ID Mapping results

## Approach
- Get the list of facets depending on the target namespace
- Create a new request to get facets
- Parse facets from url and pass to id mapping url

## Testing
Testing that the facets are displayed properly and that clicking a value updates the history correctly

## Checklist

### DIFF
--- a/src/tools/id-mapping/components/results/__mocks__/UniProtkbMapping.ts
+++ b/src/tools/id-mapping/components/results/__mocks__/UniProtkbMapping.ts
@@ -1,4 +1,189 @@
 export default {
+  facets: [
+    {
+      label: 'Status',
+      name: 'reviewed',
+      allowMultipleSelection: false,
+      values: [
+        {
+          label: 'Reviewed (Swiss-Prot)',
+          value: 'true',
+          count: 4,
+        },
+      ],
+    },
+    {
+      label: 'Model organisms',
+      name: 'model_organism',
+      allowMultipleSelection: true,
+      values: [
+        {
+          value: 'Human',
+          count: 1,
+        },
+        {
+          value: 'Mouse',
+          count: 1,
+        },
+        {
+          value: 'S. cerevisiae',
+          count: 1,
+        },
+      ],
+    },
+    {
+      label: 'Proteins with',
+      name: 'proteins_with',
+      allowMultipleSelection: true,
+      values: [
+        {
+          label: 'Active site',
+          value: 'act_site',
+          count: 4,
+        },
+        {
+          label: 'Binding site',
+          value: 'binding',
+          count: 4,
+        },
+        {
+          label: 'Catalytic activity',
+          value: 'catalytic_activity',
+          count: 4,
+        },
+        {
+          label: 'Chain',
+          value: 'chain',
+          count: 4,
+        },
+        {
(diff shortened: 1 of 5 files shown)


src/tools/id-mapping/components/results/__mocks__/UniProtkbMapping.ts
```diff
@@ -1,4 +1,189 @@
...
+          count: 3,
+        },
+        {
+          label: '>= 801',
+          value: '[801 TO *]',
+          count: 1,
+        },
+      ],
+    },
+  ],
```
aurel-l, line 186: Could you add a comment at the top with the URL of where these values are coming from?

Also, it would be good to specify to type of the mocks in the mock files. It lets us know if there's something wrong directly here:
```
const mock: TypeOfMyMock = {...}

export default mock;
```

src/tools/id-mapping/components/results/__tests__/IDMappingResult.spec.tsx
```diff
@@ -38,13 +37,16 @@ describe('IDMappingResult tests', () => {
...
+    const { history } = customRender(<IDMappingResult />, {
       route: '/id-mapping/id2',
       initialUserPreferences: {
         'view-mode': ViewMode.TABLE, // This should eventually be removed
       },
     });
-    expect(await (await screen.findAllByText('Q9Z0H0')).length).toBe(2);
+    expect((await screen.findAllByText('Q9Z0H0')).length).toBe(2);
+    const facetLink = screen.getByText(/Reviewed \(Swiss-Prot\)/);
+    fireEvent.click(facetLink);
```
aurel-l, line 49: This could be something like `screen.getByRole('link', {name: /Reviewed/})`

aurel-l (changes requested):
Small changes, but overall works fine 👍🏽

Wondering if it makes sense to display the Clusters facet for UniRef, as it will always be one type of cluster and it's the only facet for this result type.
Also, should it include the custom taxonomy facet or is it not part of the requirement for jobs? It would be a bit complicated because it ties in with the query builder logic

aurel-l:
Actually, when switching facets the whole page flashes, which is not really the behaviour of the normal result page and it should now be possible to avoid this with the latest changes on the results component.

Also wondering if facet change should push or replace history. It currently replaces, but it means clicking the back button takes you back to the dashboard instead of the previous unfacetted result set. Not sure what's the best behaviour for the user. It surprised me a bit. but doing the replace has been done on purpose in the facets component in franklin, so I guess it was a decision taken a while back for some specific reason.

aurel-l (approved):
Facet seem to work fine. I haven't been able to try with a lot of combinations though because the service is really slow right now.

2 comments on things that I've seen, even though they're not part of this PR so probably shouldn't be fixed here:
 - Warning about react keys -> I guess it needs to use a combination of from + to instead of just the from as the from value could appear multiple times
  - The sticky header has a gap. If the idea was to have buttons there as part of another PR then disregard this comment, otherwise, the spacing needs to be adjusted at some point

xwatkins:
Thanks - I'm aware of these 2 things 👍

PR #361, opened 2021-05-06
2021 06 query builder double double quotes

aurel-l:
## Purpose
Fix issue with double set of quotes https://www.ebi.ac.uk/panda/jira/browse/TRM-25888

## Approach
Give responsibility to quote to the stringifier, not the field
Also, "fixed" mock data, and behaviour for "*" search for cross-ref, but the resulting query doesn't seem to be valid anyway (was following the test descriptions 🤷🏼‍♂️)

## Testing
Modify tests, other tests keep working

## Checklist

Built on top of #357


src/query-builder/components/TextField.tsx
```diff
@@ -23,10 +18,12 @@ const TextField: FC<{
     if (trimmed.length) {
       if (field.valuePrefix && field.term === 'xref' && trimmed === '*') {
         // Query is faster with this hack
-        handleChange({ xref: `${field.valuePrefix}` });
+        // NOTE: Is this even working?..
```
xwatkins, line 21: You can remove that "optimisation hack" as it's actually not working...it should use `database` instead. Let's just remove and do `xref: myxref-*`

xwatkins (commented):
Just need to remove the "optimisation hack" :-)

src/query-builder/components/TextField.tsx
aurel-l: Shoudn't that change be part of the jira you took this morning anyway?

aurel-l: I don't understand how xref queries are supposed to work anyway, especially because I haven't seen any working fine yet, so it's hard to implement against...

xwatkins: Ok merge it and I’ll take it from there

xwatkins (approved):

PR #362, opened 2021-05-06
Fix xref search "hack"

xwatkins:
## Purpose
Optimised xref query wasn't doing anything

## Approach
change the logic to handle this special case - replacing "xref" by "database" and the prefix value as value. Also added the parsing counterpart, which is only temporary until backend add the "database" field.

## Testing
Updated test case for xref and '*' search

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "lines": 83.19,
         "statements": 83.13,
         "functions": 83.38,
-        "branches": 73.77
+        "branches": 73.76
       }
     },
     "transform": {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "lines": 83.19,
         "statements": 83.13,
         "functions": 83.38,
-        "branches": 73.76
+        "branches": 73.51
       }
     },
     "transform": {
--- a/src/query-builder/components/QueryBuilder.tsx
+++ b/src/query-builder/components/QueryBuilder.tsx
@@ -241,6 +241,10 @@ const QueryBuilder: FC<Props> = ({ onCancel, fieldToAdd }) => {
           </Button>
           <Button type="submit">Search</Button>
         </div>
+        <small>
+          Type * in the search box to search for all values for the selected
+          field.
+        </small>
       </form>
     </>
   );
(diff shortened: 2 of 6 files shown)


aurel-l (approved):
Approved.

Would be good to have feedback about the warning sentence at the bottom at some point, after some users have tested it.
Wondering if that should be in a message pattern from Franklin, informational level.

xwatkins:
> Approved.
> 
> Would be good to have feedback about the warning sentence at the bottom at some point, after some users have tested it.
> Wondering if that should be in a message pattern from Franklin, informational level.

That's a good point...we have the hero container but I'm not sure that's the right pattern here.

aurel-l:
> > Approved.
> > Would be good to have feedback about the warning sentence at the bottom at some point, after some users have tested it.
> > Wondering if that should be in a message pattern from Franklin, informational level.
> 
> That's a good point...we have the hero container but I'm not sure that's the right pattern here.

Was talking about this component https://ebi-uniprot.github.io/franklin-sites/?path=/story/layout-message--default-message

xwatkins:
> > > Approved.
> > > Would be good to have feedback about the warning sentence at the bottom at some point, after some users have tested it.
> > > Wondering if that should be in a message pattern from Franklin, informational level.
> > 
> > 
> > That's a good point...we have the hero container but I'm not sure that's the right pattern here.
> 
> Was talking about this component https://ebi-uniprot.github.io/franklin-sites/?path=/story/layout-message--default-message

I see. I don't think that's the right pattern either, messages are more to inform the user that something is happening

xwatkins:
> > > > > Approved.
> > > > > Would be good to have feedback about the warning sentence at the bottom at some point, after some users have tested it.
> > > > > Wondering if that should be in a message pattern from Franklin, informational level.
> > > > 
> > > > 
> > > > That's a good point...we have the hero container but I'm not sure that's the right pattern here.
> > > 
> > > 
> > > Was talking about this component https://ebi-uniprot.github.io/franklin-sites/?path=/story/layout-message--default-message
> > 
> > 
> > I see. I don't think that's the right pattern either, messages are more to inform the user that something is happening
> 
> We use this in the warning message in the BLAST phylotree when we have negative values, as information in the dashboard when no job is present, for the "Caution" message in the UniProtKB entry page function section, etc. It's quite versatile already

Maybe we could have another session about the "purpose" and "function" of components in the next weeks - as I think the definitions are missing for a lot of new components. These really help deciding what to use where/when

PR #365, opened 2021-05-11
2021 05 idmapping results

xwatkins:
## Purpose
Implement ID Mapping results

## Approach
- Depending on the target database, override the namespace used to display results and facets
- `SearchableNamespace` is a subset of `Namespace`. id mapping is in effect a namespace when the mapping "to" is not a namespace, but it is not searchable.
- Flatten the data structure returned by the API so that the end result for target namespaces is just an extra "from" column
- Create column renderer for the "to" column to extract xref link

## Testing
- Testing 2 types of mappings are rendered properly, with a namespace as target and with just ids as target.

## Checklist


dlrice:
When I map UniProtKB_AC-ID → UniProtKB-Swiss-Prot the results are `From` and `To` columns but I would have thought this should be UniProtKB (with all of the table columns). Example here:

http://localhost:8080/id-mapping/101984ea56046397fc72ca5ca6b59327360ec3b0

dlrice:
Could be useful to tell the user what sort of results (UniProtKB, UniParc or UniRef) they are looking at in the title eg in peptide search we have:

Screenshot 2021-05-17 at 16 55 32

xwatkins:
> When I map UniProtKB_AC-ID → UniProtKB-Swiss-Prot the results are `From` and `To` columns but I would have thought this should be UniProtKB (with all of the table columns). Example here:
> 
> http://localhost:8080/id-mapping/101984ea56046397fc72ca5ca6b59327360ec3b0

Good catch! This is now fixed.

xwatkins:
> Could be useful to tell the user what sort of results (UniProtKB, UniParc or UniRef) they are looking at in the title eg in peptide search we have:
> 
> Screenshot 2021-05-17 at 16 55 32

Looking into this and the new `ResultsDataHeader` I think this could be looked into when implementing the buttons for ID mapping

src/tools/id-mapping/components/results/__tests__/IDMappingResult.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+import IDMappingResult from '../IDMappingResult';
+
+import SimpleMappingData from '../__mocks__/SimpleMapping';
+import SimpleMappingDetails from '../__mocks__/SimpleMappingDetails';
+import UniProtkbMapping from '../__mocks__/UniProtkbMapping';
+import UniProtkbMappingDetails from '../__mocks__/UniProtkbMappingDetails';
+
+import { ViewMode } from '../../../../../shared/components/results/ResultsData';
+
+const mock = new MockAdapter(axios);
```
dlrice, line 15: Non-critical and can be done at a later time: can combine all of these mock statements as is done in https://github.com/ebi-uniprot/uniprot-website/blob/de90a1a8b66ba42d4c65b7b57ef29cd1b74b8e59/src/uniprotkb/components/entry/__tests__/Entry.spec.tsx#L34

dlrice (approved):

PR #368, opened 2021-05-13
2021 06 facet icons

xwatkins:
## Purpose
Add icons for reviewed / unreviewed (UniProkb) and reference proteome.

## Approach
Check facet name and values to add the icon to the ReactNode. Currently not very restrictive in checking and assignment, ok for now but might cause issues in the future if values change.

## Testing
Added a render which takes into account the new code

## Checklist

### DIFF
--- a/src/shared/components/results/ResultsFacets.tsx
+++ b/src/shared/components/results/ResultsFacets.tsx
@@ -1,16 +1,10 @@
 import { FC } from 'react';
-import {
-  Facets,
-  Facet,
-  Loader,
-  SwissProtIcon,
-  TremblIcon,
-  ReferenceProteomeIcon,
-} from 'franklin-sites';
+import { Facets, Facet, Loader } from 'franklin-sites';
 
 import useNS from '../../hooks/useNS';
 
 import TaxonomyFacet from './TaxonomyFacet';
+import EntryTypeIcon from '../entry/EntryTypeIcon';
 
 import { mainNamespaces } from '../../types/namespaces';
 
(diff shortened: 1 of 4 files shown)


src/shared/components/results/ResultsFacets.tsx
```diff
@@ -8,11 +15,44 @@ import TaxonomyFacet from './TaxonomyFacet';
...
+            height={ICON_HEIGHT}
+            className="icon--reference-proteome"
+          />
+          {facetValue.label}
+        </>
+      );
+    default:
+      return facetValue.label;
+  }
+};
```
aurel-l, line 54: Maybe investigate using the EntryTypeIcon with the label to get an Icon (or nothing) and inject it in front of the label

xwatkins: We didn't think about the fact that `facetValue.label` is actually a `ReactNode`, not a string so `EntryTypeIcon` doesn't apply

aurel-l: At the moment you receive it from the server it can only be a string.
I guess we switched the type to node so that we could use the same type for server-sent labels and client-side or customised label.
A type check of `typeof label === 'string'` should be enough to make TS happy

xwatkins: Good idea, will try that.

src/shared/components/results/__tests__/ResultsFacets.spec.tsx
```diff
@@ -43,7 +48,7 @@ describe('ResultsFacets', () => {
     });
   });
 
-  it('should deselect a facet', async () => {
+  test('should deselect a facet', async () => {
```
aurel-l, line 51: Looking at this change I realised there were 221 occurrences of ` it('` and 221 of ` test('` so we don't really have a consistent approach in naming. I guess it's fine to call them 'it' when the name of the test makes sense ('...should') but I don't really care for one or the other.
Just commenting here, not asking to revert

src/shared/components/results/ResultsFacets.tsx
```diff
@@ -4,15 +4,29 @@ import { Facets, Facet, Loader } from 'franklin-sites';
...
+  const { label } = facetValue;
+  if (typeof label === 'string') {
+    return (
+      <>
+        <EntryTypeIcon entryType={label} />
+        {facetValue.label}
+      </>
+    );
+  }
+  return facetValue.label;
```
aurel-l, line 27: Not sure why destructuring the argument here but then using twice the destructured variable, and twice the original argument of the function.
I guess you should use one, or the other, but not a mixed approach.

aurel-l:
Just one small comment about destructuring or not, but other than that it looks good, it should be good to go afterwards.

aurel-l (approved):

aurel-l:
I remember now that we are using the same facets in BLAST, so probably we should have the same facets with icons displayed there too.

xwatkins:
> I remember now that we are using the same facets in BLAST, so probably we should have the same facets with icons displayed there too.

Indeed. Actually I think we could use `ResultsFacets` instead of:

```
        {data?.facets?.map((facet: FacetObject) => (
          <Facet
            key={facet.name}
            data={facet}
            className={isStale ? helper.stale : undefined}
          />
        ))}
```

PR #369, opened 2021-05-13
2021 06 protein names display more

aurel-l:
## Purpose
Don't show all protein and gene names by default https://www.ebi.ac.uk/panda/jira/browse/TRM-25971, and part of https://www.ebi.ac.uk/panda/jira/browse/TRM-25949

## Approach
Stop using the complex name components as they were not flexible enough for that, just dump all the names, starting by the recommended, as a list of unique names (strings) and display that in a ExpandableList.
Did the same approach for gene names as it's quite similar and I assumed that would have been raised at some point in the future anyway.
It also removed the evidence tags as part of the changes since we're not displaying anything complex anymore

## Testing
Updated snapshots

## Checklist
  - lower coverage


xwatkins:
We're missing some names (including the Enzyme EC id if available) - see difference with old website below. You can also check when downloading as TSV.
image
image

aurel-l:
> We're missing some names (including the Enzyme EC id if available) - see difference with old website below. You can also check when downloading as TSV.
> image
> image

Fixed now

xwatkins (approved):

PR #373, opened 2021-05-17
Remove sticky

xwatkins:
## Purpose
When clicking on a protvista tooltip, it would be displayed offset on the x and y axis.

## Approach
Following a suggestion from @aurel-l , remove the `position: sticky` on the content.

## Testing
- long facet
- infinite scroll still works
- sticky action buttons
- wide tables (already an issue about this)

## Checklist

### DIFF
--- a/src/shared/components/layouts/styles/side-bar-layout.scss
+++ b/src/shared/components/layouts/styles/side-bar-layout.scss
@@ -41,9 +41,6 @@
   }
 
   &__content {
-    flex: 1;
-    top: $header-height;
-    right: 0;
     width: calc(100vw - 30ch);
   }
 }
--- a/src/shared/components/layouts/styles/side-bar-layout.scss
+++ b/src/shared/components/layouts/styles/side-bar-layout.scss
@@ -42,7 +42,6 @@
 
   &__content {
     flex: 1;
-    position: sticky;
     top: $header-height;
     right: 0;
     width: calc(100vw - 30ch);


src/shared/components/layouts/styles/side-bar-layout.scss
```diff
@@ -42,7 +42,6 @@
 
   &__content {
     flex: 1;
-    position: sticky;
     top: $header-height;
     right: 0;
     width: calc(100vw - 30ch);
```
aurel-l, line 47: I think none of this is needed now.
`flex: 1` only if we want the table to stretch when not a lot of columns but a really wide screen, which is questionable

src/shared/components/layouts/styles/side-bar-layout.scss
```diff
@@ -42,7 +42,6 @@
 
   &__content {
     flex: 1;
-    position: sticky;
     top: $header-height;
     right: 0;
```
aurel-l, line 46: I think none of this is needed now.
flex: 1 only if we want the table to stretch when not a lot of columns but a really wide screen, which is questionable

(misplaced the previous comment which I just deleted)

xwatkins: I think you're right...so this was an attempt at fixing the overflowing table?

aurel-l: I don't really know anymore, maybe

aurel-l (approved):

PR #374, opened 2021-05-18
2021 06 id mapping buttons

xwatkins:
## Purpose
Add logic to display relevant buttons when looking at ID mapping results.

## Approach
Pass down props with `namespaceFallback` and a flag to disable the card view switch.

## Testing
Visual testing only.

## Checklist

### DIFF
--- a/src/shared/components/results/ResultsButtons.tsx
+++ b/src/shared/components/results/ResultsButtons.tsx
@@ -34,6 +34,7 @@ type ResultsButtonsProps = {
   total: number;
   accessions?: string[];
   namespaceFallback?: Namespace;
+  base?: string;
   disableCardToggle?: boolean; // Note: remove if we have card view for id mapping
 };
 
@@ -42,6 +43,7 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
   selectedEntries,
   accessions,
   namespaceFallback,
+  base,
   disableCardToggle = false,
 }) => {
   const [displayDownloadPanel, setDisplayDownloadPanel] = useState(false);
@@ -72,6 +74,7 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
               totalNumberResults={total}
               onClose={() => setDisplayDownloadPanel(false)}
               namespace={namespace}
+              base={base}
             />
           </SlidingPanel>
         </Suspense>
--- a/src/shared/components/results/ResultsButtons.tsx
+++ b/src/shared/components/results/ResultsButtons.tsx
@@ -129,8 +129,9 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
             })}
           />
         </Button>
-        {(viewMode === ViewMode.TABLE || disableCardToggle) &&
-          namespace !== Namespace.idmapping && <CustomiseButton />}
+        {viewMode === ViewMode.TABLE && namespace !== Namespace.idmapping && (
+          <CustomiseButton />
+        )}
       </div>
     </>
   );
--- a/src/shared/components/results/ResultsButtons.tsx
+++ b/src/shared/components/results/ResultsButtons.tsx
@@ -33,15 +33,19 @@ type ResultsButtonsProps = {
   selectedEntries: string[];
   total: number;
   accessions?: string[];
+  namespaceFallback?: Namespace;
+  disableCardToggle?: boolean; // Note: remove if we have card view for id mapping
 };
 
 const ResultsButtons: FC<ResultsButtonsProps> = ({
   total,
   selectedEntries,
   accessions,
+  namespaceFallback,
+  disableCardToggle = false,
 }) => {
   const [displayDownloadPanel, setDisplayDownloadPanel] = useState(false);
-  const namespace = useNS() || Namespace.uniprotkb;
+  const namespace = useNS() || namespaceFallback || Namespace.uniprotkb;
   const [viewMode, setViewMode] = useUserPreferences<ViewMode>(
     'view-mode',
     ViewMode.CARD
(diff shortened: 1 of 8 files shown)


src/shared/components/results/ResultsButtons.tsx
```diff
@@ -109,6 +113,7 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
           title={`Switch to "${
             viewMode === ViewMode.CARD ? 'table' : 'card'
           }" view`}
+          disabled={namespace === Namespace.idmapping || disableCardToggle}
```
dlrice, line 116: I'm thinking that having `disableCardToggle` at our disposal means we don't need to additionally check `namespace === Namespace.idmapping`

src/shared/components/results/ResultsButtons.tsx
```diff
@@ -121,7 +126,8 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
             })}
           />
         </Button>
-        {viewMode === ViewMode.TABLE && <CustomiseButton />}
+        {(viewMode === ViewMode.TABLE || disableCardToggle) &&
+          namespace !== Namespace.idmapping && <CustomiseButton />}
```
dlrice, line 130: As we are already using props to change what is displayed we could could provide a `disableCustomise` and keep exceptional Namespace behaviour in a parent component eg `IDMappingResult`

src/shared/components/results/ResultsData.tsx
```diff
@@ -96,6 +96,33 @@ const ResultsData: FC<{
...
   }
 
+  const dataTableWithLoader =
+    namespace !== Namespace.idmapping ? (
+      <DataTableWithLoader
+        getIdKey={getIdKey}
+        columns={columns}
+        data={allResults}
+        loading={loading}
+        selected={selectedEntries}
```
dlrice, line 106: In the same way that we call `ResultsDataHeader` with different props to deal with idmapping specific behaviour how about having a `disableRowSelection` prop for `ResultsData`?

src/shared/components/results/ResultsButtons.tsx
xwatkins: You're right, it's just that as per my comment I thought we would eventually remove `disableCardToggle` so I'd left the `idmapping` check. However now that I'm working on the download I realise we will have to pass an "isIDMapping" flag down so might change that...

src/shared/components/results/ResultsData.tsx
xwatkins: Not sure, quite like the way it is at the moment in that if you provide one you have to provide the other. I think an extra attribute would be unnecessary, unless we end up having to do this lots of times.

src/shared/components/results/ResultsButtons.tsx
aurel-l: We have the card view for peptide-search, so again raising the point of:
 - is it worth it to add edge-cases here and there to force only ID mapping to only display table, when it might be easier to just display the cards and have greater consistency across the website?

xwatkins: I'm still not convinced cards make sense for a mapping to which is not a namespace such as Ensembl etc... but agree that eventually we could have cards for UniProtkb/UniRef/UniParc (which is why I added the note here https://github.com/ebi-uniprot/uniprot-website/pull/374/files/081ac813a5467323d4c187181c81f4009cfffb86#diff-e631b19484cab2883e9c24ccd391fb48f52a2a72ce4588e82bbe44eef6addd60R37

aurel-l: Agree that it isn't an improvement as it doesn't _add_ anything visually, but:
 - it adds consistency
 - it removes code complexity (the big issue for me)
 - it's not less useful than a 2-column table

Should we involve UX?

xwatkins: Yes, definitely, but the current aim is to get this functionality out (replicating what is in the current website) as quickly as possible. This is why I'm not adding cards right now.

src/shared/components/results/ResultsData.tsx
dlrice: I think I've put this comment in the wrong place. This should be about the check at L100:

```
namespace !== Namespace.idmapping ? (
```

and wether or not we should instead do:

```
disableRowSelection ?
```

More broadly, shall we just do:

* the `namespace === 'id-mapping'` check for the all of the exceptional behaviour throughout each of the components

or

* Provide additional disable props throughout (eg `disableCardToggle`) which is activated within `IDMappingResult`

but try to avoid both

xwatkins:
I've pushed the changes to allow download. Tried to add card view to remove the `disableCardToggle` but hit a wall because of what looks like a race condition with `useUserPreferences` - as it's not a hard requirement, not doing the card view right now.

I'm not the original PR comments still apply, but let me know if I missed anything. Ta!

dlrice:
I think it would make sense to not show the "Customise data" in the download when viewing just `from' and `to` mappings:
Screenshot 2021-05-20 at 12 06 57

src/shared/components/results/ResultsButtons.tsx
```diff
@@ -121,7 +129,8 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
             })}
           />
         </Button>
-        {viewMode === ViewMode.TABLE && <CustomiseButton />}
+        {(viewMode === ViewMode.TABLE || disableCardToggle) &&
```
dlrice, line 132: Does `disableCardToggle` need to be here?

xwatkins:
> I think it would make sense to not show the "Customise data" in the download when viewing just `from' and `to` mappings:
> Screenshot 2021-05-20 at 12 06 57

Good catch! removed

dlrice (approved):

PR #375, opened 2021-05-18
display less evidences in uniprotkb results

aurel-l:
## Purpose
Remove the evidence tags on the UniProtKB results columns [TRM-25949](https://www.ebi.ac.uk/panda/jira/browse/TRM-25949)

## Approach
Remove the evidence tags on *some* of the UniProtKB results columns, use plain text expandable lists instead.
Will need a complete review of all the columns anyway because a lot are using the same components as on the Entry page when it doesn't really make sense within a table cell. And some columns just don't have an implementation yet.

## Testing
Updated snapshots

## Checklist

Check the columns that have been modified (see diff view of the column config file)

### DIFF
--- a/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
+++ b/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
@@ -3,6 +3,7 @@ import { ExpandableList, LongNumber, Sequence } from 'franklin-sites';
 import { Link } from 'react-router-dom';
 
 import SimpleView from '../../shared/components/views/SimpleView';
+import { ECNumbersView } from '../components/protein-data-views/ProteinNamesView';
 import TaxonomyView, {
   TaxonomyLineage,
 } from '../../shared/components/entry/TaxonomyView';
@@ -412,7 +413,7 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccPolymorphism, {
   label: 'Polymorphysm',
   render: (data) => {
     const { polymorphysm } = data[EntrySection.Sequence];
-    return polymorphysm && <FreeTextView comments={polymorphysm} />;
+    return <FreeTextView comments={polymorphysm} noEvidence />;
   },
 });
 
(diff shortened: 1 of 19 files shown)


src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
```diff
@@ -204,82 +201,68 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.length, {
...
-              geneData.geneName && (
-                <div key={geneData.geneName.value}>
-                  {geneData.geneName.value}
-                </div>
-              )
-          )}
+
+    const names = geneNamesData?.map((geneNames) => geneNames.geneName?.value);
+
+    return names?.length ? (
```
dlrice, line 210: I see the pattern `predicate ? true-branch : null` for all of the column modifications in this PR - is there a reason this is used rather `predicate && true-branch` (which I personally find a little bit easier to read)?

dlrice (approved):

src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
aurel-l: This is because these are lengths, so you could end up displaying "0".
I could also test on the array itself but not its length, but I think the Expandable list renders an empty list, but I want it to render nothing, not an empty `<ul>`.

dlrice: thoughts on?
```
    return (
      Boolean(names?.length) && (
        <ExpandableList descriptionString="gene names" numberCollapsedItems={1}>
          {names}
        </ExpandableList>
      )
    );
```

aurel-l: I would prefer:
```
return (
      names && (
        <ExpandableList descriptionString="gene names" numberCollapsedItems={1}>
          {names}
        </ExpandableList>
      )
    );
```
And modify ExpandableList in Franklin to return null when the number of children is 0. What do you think?

dlrice: That sounds good - also names could potentially include `undefined` elements so could determine if this is the case within franklin too

dlrice (approved):

PR #378, opened 2021-05-21
2021 06 ellipsis reveal

xwatkins:
## Purpose

## Approach

## Testing

## Checklist


src/supporting-data/citations/components/__tests__/LiteratureCitation.spec.tsx
```diff
@@ -25,6 +26,18 @@ describe('Publication component', () => {
...
+  test('should display 1 author', async () => {
+    const mockData: CitationsAPIModel = {
+      citation: {
+        id: '14702039',
+        authors: ['Ota T.'],
+      },
+    };
+    rendered = customRender(<LiteratureCitation data={mockData} />);
+    const author = await screen.findByText('Ota T.');
+    expect(author).toBeTruthy();
```
aurel-l, line 38: `author` is a promise, it will always be truthy

aurel-l:
OK with the approach 👍🏽 , I guess it can be generalised to all the columns using ExpandableList.

Another point is that the current website either shows *all* the hidden text in all the cells, or none. So while this will make happy the people that want tidy tables and aligned columns, if other users depends on the whole of the text, we will in the same way get "strong" feedback if someone has to click all the instances of `[...]` to display the information they may need as in the current website.

xwatkins:
> OK with the approach 👍🏽 , I guess it can be generalised to all the columns using ExpandableList.
> 
> Another point is that the current website either shows _all_ the hidden text in all the cells, or none. So while this will make happy the people that want tidy tables and aligned columns, if other users depends on the whole of the text, we will in the same way get "strong" feedback if someone has to click all the instances of `[...]` to display the information they may need as in the current website.

I think you're right and we're going to have to add `[...]` to the corresponding columns. Fun fun fun.

I've noticed an issue with gene names sometimes only showing `[...]` so will look into fixing this before merging. Also so comment on Franklin branch about removing `[...]` from "Show abstract"

src/supporting-data/citations/components/__tests__/LiteratureCitation.spec.tsx
```diff
@@ -30,18 +30,18 @@ describe('Publication component', () => {
...
       citation: {
         id: '14702039',
-        authors: ['Ota T.'],
+        authors: ['Smith X.'],
       },
     };
     rendered = customRender(<LiteratureCitation data={mockData} />);
-    const author = await screen.findByText('Ota T.');
-    expect(author).toBeTruthy();
+    const author = await screen.findByText('Smith X.');
```
aurel-l, line 37: Is there a specific need for these `await ... findBy` instead of a plain `getBy`? When reading the test it feels like there is some asynchronous rendering happening, which is not the case, is it?
See https://testing-library.com/docs/queries/about/#types-of-queries for reference

xwatkins: No need, I think I had a `fireEvent` in there before and forgot to update it.

aurel-l: I went ahead and modified it in the others places where this was used too, and modified a bit more the rest of the test file

src/supporting-data/citations/components/LiteratureCitation.tsx
```diff
@@ -53,15 +53,24 @@ const Authors: FC<AuthorProps> = ({ authors, limit = 10 }) => {
...
+        <EllipsisReveal>
+          {', '}
+          {hiddenAuthors.map((author, index) => (
+            // eslint-disable-next-line react/no-array-index-key
+            <Fragment key={index}>
+              {index !== 0 && ', '}
+              <Link to={getLinkToAuthor(author)}>{author}</Link>
+            </Fragment>
+          ))}
+        </EllipsisReveal>
```
aurel-l, line 66:
```suggestion
        <EllipsisReveal>
          {hiddenAuthors.map((author, index) => (
            // eslint-disable-next-line react/no-array-index-key
            <Fragment key={index}>
              , <Link to={getLinkToAuthor(author)}>{author}</Link>
            </Fragment>
          ))}
        </EllipsisReveal>
```

Do double check in a local instance though, I did that in GitHub but unsure if the whitespace will be respected after the ",", otherwhise `{", "}`

src/supporting-data/citations/components/__tests__/LiteratureCitation.spec.tsx
```diff
@@ -19,16 +23,25 @@ describe('Publication component', () => {
...
+  });
+
+  test('should display 1 author', async () => {
+    const mockData: CitationsAPIModel = {
+      citation: {
+        id: '14702039',
+        authors: ['Smith X.'],
+      },
+    };
+    rendered = customRender(<LiteratureCitation data={mockData} />);
```
aurel-l, line 38:
```suggestion
    rendered.rerender(<LiteratureCitation data={mockData} />);
```
That's the comment I did a while back that if you just keep rendering during one test, the previous render (the one in the beforeEach) will be kept, alongside this one. So you either need to unmount it, or rerender on top of the previous render, or not have a beforeEach render

aurel-l: Was not causing an issue here, but could avoid unexpected bugs later on

aurel-l (approved):

PR #380, opened 2021-05-24
Update query for backend

xwatkins:
## Purpose
Backend asked us to optimise queries

## Approach
- Use query instead of facet to get the SwissProt/TrEMBL * results from the tile
- Remove facets when not needed

## Testing
Update snapshots

## Checklist

### DIFF
--- a/src/shared/hooks/useNSQuery.ts
+++ b/src/shared/hooks/useNSQuery.ts
@@ -40,13 +40,8 @@ const useNSQuery = ({
   }
 
   const { search: queryParamFromUrl } = location;
-  const {
-    query,
-    selectedFacets,
-    sortColumn,
-    sortDirection,
-    direct,
-  } = getParamsFromURL(queryParamFromUrl);
+  const { query, selectedFacets, sortColumn, sortDirection, direct } =
+    getParamsFromURL(queryParamFromUrl);
 
   const url = useMemo(() => {
     if (!query && !accessions.length) {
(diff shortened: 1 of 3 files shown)


aurel-l (approved):
Added my own change for facets optimisation

src/shared/hooks/useNSQuery.ts
```diff
@@ -57,7 +52,7 @@ const useNSQuery = ({
       query,
       columns: withColumns ? queryColumns : undefined,
       selectedFacets,
-      facets: withFacets ? undefined : undefined,
+      facets: withFacets ? undefined : [],
```
xwatkins, line 55: I was looking at that and thinking it's not obvious that passing `undefined` means it will get the default for the current namespace. As there's not many calls of `getAPIQueryUrl`, I wonder if we should pass the default at the time of the call instead? What do you think?

aurel-l: Probably we should eventually consolidate the approach we take for parameters for all of these URL builder functions in general. If we change something for getAPIQueryUrl we also need to change the way getAccessionsURL handles facets, because it's not consistent either.

xwatkins: Ok, so we merge?

aurel-l: I guess so.
We might need a Jira for review of all of these later on, or we might need to tackle that part of the BLAST against UniParc/UniRef as it might affect getAccessionsURL (we'll use accession endpoint, but it's assuming default facets for UniProtKB at the moment)

xwatkins: Good idea, I'll add a subtask to the issue.

PR #389, opened 2021-06-08
2021 07 uniparc entry columns

aurel-l:
## Purpose
UniParc entry columns https://www.ebi.ac.uk/panda/jira/browse/TRM-25540 and https://www.ebi.ac.uk/panda/jira/browse/TRM-25726

## Approach
 - Avoid using all the logic in the results components because it was too much for this specific case which was quite reduced anyway (no need to order columns for example). So take a similar approach without using all the custom hooks, and had to generalise in some places anyway in order to keep types happy.
 - Modified and reused the column selector.
 - Split the payload for the UniParc entry page in 2, one for the first render (sequence, name, basic things), and one for the xrefs themselves (used in the facets and in the xrefs section)
   - Regarding this, I'm using a hack to not get any cross-reference for the first payload (asking for the cross-references linked to tax id 0). Hopefully backend will eventually see a use case for a cross-reference endpoint (as far as I remember that was refused before) similar to the members endpoint in UniRef.
 - Create a new Timeline visual component (Tried to use the same approach than the BUSCO one at first, but then reverted to something more basic as the requirements were not as complex as BUSCO)

## Testing
Unit tests for all the columns. Updated UniParc mock with latest data and fields and updated corresponding snapshots

## Checklist

~Note: as of this morning (8/06) the dev API doesn't work for UniParc entries. Need to use the beta API, or wait for the issue to be resolved.~


src/uniparc/config/UniParcXRefsColumnConfiguration.tsx
```diff
@@ -0,0 +1,322 @@
...
+  EntryTypeIcon,
+} from '../../shared/components/entry/EntryTypeIcon';
+import TaxonomyView from '../../shared/components/entry/TaxonomyView';
+
+import { getEntryPath } from '../../app/config/urls';
+
+import parseDate from '../../shared/utils/parseDate';
+
+import { Namespace } from '../../shared/types/namespaces';
+import { ColumnConfiguration } from '../../shared/types/columnConfiguration';
```
dlrice, line 17: Type imports want to happen last according to our conventions

src/uniparc/config/UniParcXRefsColumnConfiguration.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+import {
+  databaseToEntryType,
+  UniParcXRef,
+  XRefsInternalDatabasesEnum,
+} from '../adapters/uniParcConverter';
+import { ColumnDescriptor } from '../../shared/hooks/useColumns';
+import evidenceUrls from '../../uniprotkb/config/evidenceUrls';
+import { processUrlTemplate } from '../../uniprotkb/components/protein-data-views/XRefView';
```
dlrice, line 26: This wants to precede type imports

src/uniparc/adapters/uniParcConverter.ts
```diff
@@ -27,14 +27,11 @@ export type UniParcXRef = Partial<{
...
-  // "properties" only there when database is "EMBL", and not for all of them.
-  // We shouldn't rely on that to display anything in the frontend, and it
-  // wasn't used in the previous website anyway
-  properties?: Array<{ key: string; value: string }>;
   proteinName: string;
   proteomeId: string;
   component: string;
-  version?: number; // might not always be there (e.g., for PRF xrefs)
+  chain: string;
+  version: number; // might not always be there (e.g., for PRF xrefs)
```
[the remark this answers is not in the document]
aurel-l: I think I removed all optionnals within this object just because we already always wrapping the whole thing as a Partial. Could revert, but then most of the keys would be `?:` anyway.

src/uniparc/config/UniParcXRefsColumnConfiguration.tsx
aurel-l: Good catch, I let that one though, should be fixed now

src/shared/components/column-select/ColumnSelect.tsx
```diff
@@ -38,27 +38,29 @@ type ColumnSelectProps = {
...
-  const defaultColumns = nsToDefaultColumns[namespace];
+  const primaryKeyColumns = nsToPrimaryKeyColumns(namespace, isEntryPage);
+  const defaultColumns = nsToDefaultColumns(namespace, isEntryPage);
 
   // remove the entry field from the choices as this must always be present
   // in the url fields parameter when making the search request ie
   // don't give users the choice to remove it
   const removableSelectedColumns = selectedColumns.filter(
-    (column) => column !== primaryKeyColumn
+    (column) => !primaryKeyColumns.includes(column)
```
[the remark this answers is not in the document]
aurel-l: I see, I didn't know that function existed in lodash. Even though the non-lodash version is not too big, the lodash one does make the intent clearer. It is now using the helper you suggested

src/uniparc/components/entry/Entry.tsx
```diff
@@ -78,7 +95,7 @@ const Entry: FC = () => {
 
   const transformedData = uniParcConverter(dataObject.data);
 
-  const entrySidebar = <XRefsFacets xrefs={dataObject} />;
+  const entrySidebar = <XRefsFacets xrefs={xrefsDataObject} />;
```
dlrice, line 98: Noticed that the Organisms facets disappear:
1. http://localhost:8080/uniparc/UPI000000000B/entry
2. Choose an organism from the Organisms facet
3. The Organisms facet will disappear

src/uniparc/adapters/uniParcConverter.ts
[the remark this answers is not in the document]
dlrice: Ah I see. I didn't catch the use of Partial. And also the comment in the line below it is connected and it all makes sense now. So all good.

src/uniparc/components/entry/Entry.tsx
aurel-l: 🤔 mmh, I can't really reproduce on my side.
If you want ping me whenever you want on Slack and we can go over this specific issue

aurel-l: Should be fixed now

dlrice (approved):

PR #391, opened 2021-06-15
useBasket

aurel-l:
## Purpose
useBasket hook [TRM-26119](https://www.ebi.ac.uk/panda/jira/browse/TRM-26119)

## Approach
Rename useUserPreferences to useLocalStorage, and wrap it with a useBasket hook for the specific case of the basket (serialise/deserialise Set to/from arrays).

## Testing
New unit tests

## Checklist

### DIFF
--- a/src/shared/hooks/__tests__/useBasket.spec.ts
+++ b/src/shared/hooks/__tests__/useBasket.spec.ts
@@ -1,5 +1,5 @@
 import { renderHook, act } from '@testing-library/react-hooks';
-import { JsonValue } from 'type-fest';
+
 import { Namespace } from '../../types/namespaces';
 
 import useBasket from '../useBasket';
--- a/src/shared/hooks/__tests__/useBasket.spec.ts
+++ b/src/shared/hooks/__tests__/useBasket.spec.ts
@@ -1,0 +1,108 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import { JsonValue } from 'type-fest';
+import { Namespace } from '../../types/namespaces';
+
+import useBasket from '../useBasket';
+
+describe('useBasket hook', () => {
+  afterEach(() => {
+    window.localStorage.clear();
+  });
+
+  test('get value, basic, first time', async () => {
+    const { result } = renderHook(() => useBasket());
+
+    expect(result.current[0]).toEqual({
+      [Namespace.uniprotkb]: new Set(),
+      [Namespace.uniref]: new Set(),
+      [Namespace.uniparc]: new Set(),
+    });
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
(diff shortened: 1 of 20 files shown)


src/shared/hooks/useBasket.ts
```diff
@@ -0,0 +1,63 @@
...
+type StringifiableBasket = {
+  [Namespace.uniprotkb]: Array<string>;
+  [Namespace.uniref]: Array<`UniRef${50 | 90 | 100}_${string}`>;
+  [Namespace.uniparc]: Array<`UPI${string}`>;
+};
+
+const deserialise = (stringifiableBasket: StringifiableBasket): Basket => {
+  const entries = Object.entries(stringifiableBasket);
+  const object = Object.fromEntries(
+    entries.map(([key, value]) => [key, new Set(value)])
```
xwatkins, line 22: Wondering if it would be worth filtering out potentially empty/null values?

src/shared/hooks/useBasket.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) as Basket;
+  return object;
+};
+
+const serialise = (basket: Basket): StringifiableBasket => {
+  const entries = Object.entries(basket);
+  const object = Object.fromEntries(
+    entries.map(([key, value]) => [key, Array.from(value)])
```
xwatkins, line 30: Same here, there shouldn't be empty/null/undefined values, but might be worth checking?

src/shared/hooks/__tests__/useBasket.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      JSON.stringify({
+        [Namespace.uniprotkb]: ['P05067'],
+        [Namespace.uniref]: [],
+        [Namespace.uniparc]: ['UPI0000000001', 'UPI0000000002'],
+      })
+    );
+    const { result } = renderHook(() => useBasket());
+
+    act(() =>
+      result.current[1]({
```
xwatkins, line 58: Confused by this. why `current[1]`?

xwatkins (changes requested):
Just a suggestion for additional checks and a question but good otherwise 👍

src/shared/hooks/__tests__/useBasket.spec.ts
aurel-l: current is the result of the hook. Since it's a tuple of `[basket, setBasket]`, `current[1]` is `setBasket`.
All the tests in this file are a copy of the ones from useLocalStorage but adapted a bit, so it's the same logic there.

src/shared/hooks/useBasket.ts
aurel-l: done

src/shared/hooks/useBasket.ts
aurel-l: done

xwatkins (approved):

PR #392, opened 2021-06-17
End of sprint tasks

dlrice:
## Purpose
End of sprint tasks

## Approach
* You may have to `rm -r node_modules/.cache` due to the eslint upgrade
* I've incorporated yalc into gitignore and jest config. Using yalc is a workaround as I don't understand what causes:
  ```
  TypeError: fixRegExpWellKnownSymbolLogic is not a function
  ```
  when using yarn link. To install yalc:
  ```
  yarn global add yalc
  ```
  To use yalc:
  In franklin, after building
  ```
  yalc publish
  ```
  In uniprot-website: 
  ```
  yalc link franklin-sites
  ```
  When changes happen in franklin that you want to appear in the linked uniprot-website, in franklin:
  ```
   yalc push
  ```
* `"testEnvironment": "jsdom"` due to jest 27 now using node as the default environment
* Upgraded puppeteer and fixed paths to visual and functional tests
* Depends on https://github.com/ebi-uniprot/franklin-sites/pull/204

The following have not been upgraded as [per notes](https://github.com/ebi-uniprot/coding-guidelines/blob/main/sprint_release.md):
```
 devDependencies
   name                           range   from        to            url
 ◯ @types/d3                      latest  5.16.3   ❯  6.7.0         https://github.com/DefinitelyTyped/DefinitelyTyped.git
 ◯ eslint-plugin-uniprot-website  latest  1.0.0    ❯  exotic        file:./eslint-custom-rules
 ◯ sass                           latest  1.32.13  ❯  1.35.1        https://github.com/sass/dart-sass
 ◯ sass-loader                    latest  10.1.1   ❯  12.1.0        https://github.com/webpack-contrib/sass-loader

 dependencies
   name                           range   from        to            url
 ◯ d3                             latest  5.16.0   ❯  7.0.0         https://d3js.org
 ◯ history                        latest  4.10.1   ❯  5.0.0         https://github.com/ReactTraining/history#readme
 ◯ lit-html                       latest  1.3.0    ❯  1.4.1         https://lit-html.polymer-project.org/
 ◯ protvista-uniprot              latest  2.6.1    ❯  2.7.0-beta.3
 ◯ protvista-variation-adapter    latest  3.1.2    ❯  3.2.14        https://ebi-webcomponents.github.io/nightingale/
```

## Testing
N/A

## Checklist


src/tools/id-mapping/components/IDMappingForm.tsx
```diff
@@ -354,7 +354,9 @@ const IDMappingForm = () => {
                     });
                   }}
                   label={fromDbInfo.displayName}
-                  defaultActiveNodes={[fromDb.selected]}
+                  defaultActiveNodes={
+                    fromDb.selected ? [fromDb.selected.toString()] : undefined
```
xwatkins, line 358: Why the `toString()` here?

src/tools/id-mapping/components/IDMappingForm.tsx
```diff
@@ -368,7 +370,9 @@ const IDMappingForm = () => {
                     setToDb((toDb) => ({ ...toDb, selected: id }));
                   }}
                   label={toDbInfo.displayName}
-                  defaultActiveNodes={[toDb.selected]}
+                  defaultActiveNodes={
+                    toDb.selected ? [toDb.selected.toString()] : undefined
```
xwatkins, line 374: Same as above

package.json
```diff
@@ -57,10 +57,12 @@
       "<rootDir>/node_modules/(?!lit-html|franklin-sites|p-map|aggregate-error|clean-stack).+\\.js"
     ],
     "testPathIgnorePatterns": [
-      "/__mocks__/"
+      "/__mocks__/",
+      "/.yalc/"
```
xwatkins, line 61: I'm not sure about adding a reference to `yalc` in package.json if it's not a dependency. I don't think it's a problem to have it in `.gitignore` however.

src/tools/id-mapping/components/IDMappingForm.tsx
dlrice: Good question, it's not needed. I have used `as string` instead.

package.json
dlrice: I wasn't sure either but I didn't know how else to handle this issue. Without including `.yalc` in `testPathIgnorePatterns` and `coveragePathIgnorePatterns` I observer errors. Any ideas to avoid its inclusion? Should we remove the `.yalc` references from `package.json` to force devs to remove any yalc stuff themselves (`yalc remove my-package`)?

xwatkins: Where do the .yalc files reside? I find it weird that `testPathIgnorePatterns` and `coveragePathIgnorePatterns` ignores other hidden directories and files (`.husky`, `.vscode` etc...)

aurel-l: > ignores other hidden directories and files (.husky, .vscode etc...)

Might be because they don't have javascript files within

xwatkins: Might be!
In that case, have you tried making `transform` more specific by adding the `src/` path in the regex if that's possible?

xwatkins: Or a more generic `transformIgnorePatterns` to ignore directories at the root starting with "."
Maybe `<rootDir>\\..+\\.js`?

dlrice: That's a good idea. Latest commit does this.

xwatkins (approved):

PR #393, opened 2021-06-17
2021 08 use useBasket

aurel-l:
## Purpose
use useBasket around the website [TRM-26120](https://www.ebi.ac.uk/panda/jira/browse/TRM-26120) excluding the subtask about using the bubble component (need franklin changes).

## Approach
 - Change basket from object of Sets, to Map of Sets (was helping with types and simplifying the code)
 - Add a Map as a caching layer on top of localStorage to avoid reading from it too much (especially in result pages where each accession cell uses a useBasket hook)
 - Add a basket icon next to most of the accessions around the website
   - In entry pages, on the right of the title
   - In result page, on the left of the accession in the table view
   - In result page, on the top right of the card in the card view 
 - Clicking on the basket icon marking the presence of the accession removes it from the basket
 - Use more `AccessionView` around the website
 - Other fixes here and there

## Testing
 - Update snapshots

## Checklist

### DIFF
--- a/src/proteomes/config/__tests__/__snapshots__/ProteomesColumnConfiguration.spec.tsx.snap
+++ b/src/proteomes/config/__tests__/__snapshots__/ProteomesColumnConfiguration.spec.tsx.snap
@@ -302,7 +302,7 @@ exports[`ProteomesColumnConfiguration component should render column "protein_co
 exports[`ProteomesColumnConfiguration component should render column "upid": upid 1`] = `
 <DocumentFragment>
   <div
-    class="accession-view"
+    class="no-wrap"
   >
     <span
       class="entry-title__status icon--reference-proteome"
--- a/src/shared/components/BasketStatus.tsx
+++ b/src/shared/components/BasketStatus.tsx
@@ -1,0 +1,44 @@
+import { Button, BasketIcon } from 'franklin-sites';
+
+import useBasket from '../hooks/useBasket';
+
+import accessionToNamespace from '../utils/accessionToNamespace';
+
+import styles from './styles/basket-status.module.scss';
+
+type Props = { id?: string };
+
+const BasketStatus = ({ id }: Props) => {
+  const [basket, setBasket] = useBasket();
+
+  if (!id) {
+    return null;
+  }
+
+  const namespace = accessionToNamespace(id);
+
+  const inBasket = Boolean(basket.get(namespace)?.has(id));
+
+  if (!inBasket) {
+    return null;
+  }
+
+  return (
+    <Button
+      className={styles['basket-status']}
+      onClick={() => {
+        setBasket((currentBasket) => {
+          const newSet = new Set(currentBasket.get(namespace));
+          newSet.delete(id);
+          return new Map([...currentBasket, [namespace, newSet]]);
+        });
+      }}
+      variant="tertiary"
+      title={`Remove ${id} from basket`}
+    >
+      <BasketIcon width="1ch" height="1ch" />
+    </Button>
+  );
+};
+
+export default BasketStatus;
--- a/src/shared/components/BasketStatus.tsx
+++ b/src/shared/components/BasketStatus.tsx
@@ -9,7 +9,7 @@ import styles from './styles/basket-status.module.scss';
 type Props = { id?: string };
 
 const BasketStatus = ({ id }: Props) => {
-  const [basket, setBasket] = useBasket();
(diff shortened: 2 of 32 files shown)


xwatkins (changes requested):
Code looks great.
Functionality/UI I'm unsure about a few things, some of which we should probably discuss with UX:
- The basket icon is a bit lost in the cards. I think it would be good to have a consistent way to display the icon with the accession, which is why I was suggesting after the accession in the results table, cards and entry
- Entry shouldn't display an active Add to basket button if the entry is in the basket already
- The basket icon in the entry should match the header size (if there's a way, we should look into having icon sizes for `.medium`, `.large` etc...)
- I'm not sure about the click to remove behaviour. I would have expected to open the basket instead (but that's just me).

xwatkins:
> Code looks great.
> Functionality/UI I'm unsure about a few things, some of which we should probably discuss with UX:
> 
> * The basket icon is a bit lost in the cards. I think it would be good to have a consistent way to display the icon with the accession, which is why I was suggesting after the accession in the results table, cards and entry
> * Entry shouldn't display an active Add to basket button if the entry is in the basket already
> * The basket icon in the entry should match the header size (if there's a way, we should look into having icon sizes for `.medium`, `.large` etc...)
> * I'm not sure about the click to remove behaviour. I would have expected to open the basket instead (but that's just me).

Update: had a chat with Marija and she is making mockups.

aurel-l:
OK, I guess I'll wait for the next design meeting to do any further change then

xwatkins:
I think she'll just post to Slack as soon as she has something ready.
I would say you can go ahead and merge, and we can touch up afterwards.
One thing I was wondering, wouldn't it be less expensive to make just one call in the `Results` component and pass down the list of basket items instead of calling the (cached) dataset for every row?

aurel-l:
> I would say you can go ahead and merge, and we can touch up afterwards

OK, will do, so that we can get the changed basket model to the main branch and start playing with the basket.

> One thing I was wondering, wouldn't it be less expensive to make just one call in the Results component and pass down the list of basket items instead of calling the (cached) dataset for every row?

I also think it _might_ be less expensive (might be the same actually thanks to the cache optimisation, but will certainly not be worse), but the refactor at different levels was not worth it at the time as it might require a model change (pass the basket status alongside the data to the data table) and possibly changes to the data table props if we need some different behaviour (we cannot easily remove from the table at the moment). I'm expecting some changes to the required specs once we start playing with it so I wanted to keep it simple for now, and do optimisations later.

> Entry shouldn't display an active Add to basket button if the entry is in the basket already

For example this is the kind of changes I assumed would come afterwards when playing with it. At the moment I tried to follow the current behaviour (an entry in the basket still has that button on its entry page) but I agree with you.
For example I would maybe want to untick the entries added to the basket after clicking on the button to add them, but I want more people to play with it before making that change and I'm sticking with what we have on the website at the moment.

xwatkins:
> > I would say you can go ahead and merge, and we can touch up afterwards
> 
> OK, will do, so that we can get the changed basket model and the cache optimisation to the main branch and start playing with the basket.
> 
> > One thing I was wondering, wouldn't it be less expensive to make just one call in the Results component and pass down the list of basket items instead of calling the (cached) dataset for every row?
> 
> I also think it _might_ be less expensive (might be the same actually thanks to the cache optimisation, but will certainly not be worse), but the refactor at different levels was not worth it at the time as it might require a model change (pass the basket status alongside the data to the data table) and possibly changes to the data table props if we need some different behaviour (we cannot easily remove selection from the table logic at the moment). I'm expecting some changes to the required specs once we start playing with it so I wanted to keep it simple for now, and do optimisations later.
> 
> > Entry shouldn't display an active Add to basket button if the entry is in the basket already
> 
> For example this is the kind of changes I assumed would come afterwards when playing with it. At the moment I tried to follow the current behaviour (an entry in the basket still has that button on its entry page) but I agree with you.
> For example I would maybe want to untick the entries added to the basket after clicking on the button to add them, but I want more people to play with it before making that change and I'm sticking with what we have on the website at the moment.

👍  - I think the "add to basket" still being enabled on the current website is a bug BTW ;-)

xwatkins (approved):

aurel-l:
Merging, but keeping the branch around to do more changes covered by the corresponding jira (will further PRs if needed though).
I've commented out the "remove from basket" logic in `BasketStatus` so that you @xwatkins can replace it with the logic to open the basket panel.

PR #395, opened 2021-06-22
follow first design proposals for basket status icon

aurel-l:
## Purpose
Follow up of https://github.com/ebi-uniprot/uniprot-website/pull/393 implementing design and behaviour proposals

## Approach
Change icon, size, and positionning.
Add logic to remove from basket on entry pages when already in the basket.

## Testing
Updated snapshots

## Checklist

### DIFF
--- a/src/shared/components/action-buttons/AddToBasket.tsx
+++ b/src/shared/components/action-buttons/AddToBasket.tsx
@@ -9,24 +9,37 @@ import accessionToNamespace from '../../utils/accessionToNamespace';
 import { pluralise } from '../../utils/utils';
 import { Namespace } from '../../types/namespaces';
 
-type AddToBasketButtonProps = { selectedEntries: string[] };
+type AddToBasketButtonProps = { selectedEntries: string | string[] };
 
 const AddToBasketButton: FC<AddToBasketButtonProps> = ({ selectedEntries }) => {
-  const [, setBasket] = useBasket();
+  const [basket, setBasket] = useBasket();
 
-  const n = selectedEntries.length;
+  const isSingleEntry = !Array.isArray(selectedEntries);
+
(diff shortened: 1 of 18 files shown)


src/shared/components/action-buttons/AddToBasket.tsx
```diff
@@ -9,24 +9,37 @@ import accessionToNamespace from '../../utils/accessionToNamespace';
...
 const AddToBasketButton: FC<AddToBasketButtonProps> = ({ selectedEntries }) => {
-  const [, setBasket] = useBasket();
+  const [basket, setBasket] = useBasket();
 
-  const n = selectedEntries.length;
+  const isSingleEntry = !Array.isArray(selectedEntries);
+
+  const accessions = Array.isArray(selectedEntries)
+    ? selectedEntries
+    : [selectedEntries];
```
dlrice, line 21:
```suggestion
  const accessions = isSingleEntry
    ? [selectedEntries]
    : selectedEntries;
```

dlrice (approved):

src/shared/components/action-buttons/AddToBasket.tsx
aurel-l: I did think about that first, but TS forgets the implication of this boolean when it's used after being set. So at that point, if I were to use that boolean, it would believe `selectedEntries` is `string | string[]` in both places, while re-doing the test in the same statement allows us to have `string` in the first case and `string[]` in the second one.

aurel-l: Coming back to this comment as I rememberd this when reading the upcoming features in the next version of TS.
In the future this case should be handled as we both expected it
https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#cfa-aliased-conditions

PR #396, opened 2021-06-23
improve UX when a new version of the app is available

aurel-l:
## Purpose
Improve UX when getting a new version of the website

## Approach
Replace the pop up with a forced refresh whenever the user changes path. The reload will happen on changing the page so it should barely be noticeable by the user as they already expect the page to change.
Using a shared module/flag to mark the application as in need of reloading, and did the similar approach as `useScrollToTop` to detect when is a good time to trigger the forced refresh.

## Testing
Local production deploys and test update flow when forcing a change in the codebase

## Checklist

### DIFF
--- a/src/app/components/App.tsx
+++ b/src/app/components/App.tsx
@@ -10,6 +10,7 @@ import GDPR from '../../shared/components/gdpr/GDPR';
 import history from '../../shared/utils/browserHistory';
 
 import useScrollToTop from '../../shared/hooks/useScrollToTop';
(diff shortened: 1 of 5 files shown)


dlrice:
What's a good way to test this locally? Not sure how to force a change in the codebase?

aurel-l:
> What's a good way to test this locally? Not sure how to force a change in the codebase?

`yarn run build:prod` and serve locally (for example with `npx serve build`). Then do a change (add a console.log for example) and redo the same process to deploy a new version.

dlrice (approved):

PR #398, opened 2021-06-24
Merge and test on PRs.

dlrice:
## Purpose
Travis informed us of when a PR would fail if merged. This PR tries to reproduce the same functionality.

## Approach

1. Condition on `merge_request_event`
2. Checkout the target and merge in with a dummy commit message
3. If conflicts, bail
4. Else, test

If it works will copy this to franklin.

Also stopped replaced `buildkite/puppeteer` with `node:15` as we weren't using it.

## Testing
N/A

## Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -21,8 +21,9 @@ test:
     - yarn test
 
 merge_and_test:
-  only:
-    - merge_requests
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
+      when: always
   extends: .js
   stage: test
   script:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -17,6 +17,7 @@ test:
   extends: .js
   stage: test
   script:
+    - env
     - echo $CI_PIPELINE_SOURCE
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -5,23 +5,33 @@ stages:
...
+      when: always
+  extends: .js
+  stage: test
+  script:
+    - echo Merging $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME into $CI_MERGE_REQUEST_TARGET_BRANCH_NAME to test upstream effects
+    - git checkout $CI_MERGE_REQUEST_TARGET_BRANCH_NAME
+    - git merge $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME -m "Merge and test"
+    - test -n "$(git status --porcelain)" && exit 1 # Exit if merge has conflicts
     - yarn
     - yarn test
```
aurel-l, line 36: Don't you need to add this somewhere:
```
only:
    - external_pull_requests
```

dlrice: Was trying out [rules](https://docs.gitlab.com/ee/ci/merge_request_pipelines/#use-rules-to-run-pipelines-for-merge-requests) but if this doesn't work then we should try your suggestion.

PR #404, opened 2021-06-28
2021 08 publication and annotation links

dlrice:
## Purpose
"Add a publication" and "Community annotation" links ([jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26094))

## Approach
* All entries now have an `Add a publication` link
* Only those entries that have community curation submissions show a link. The vast majority of entries don't have community annotations (https://beta.uniprot.org/uniprotkb?query=(community_pubmed_id:*) has 1,425 results) so I opted to not display any sort of loading indication when querying for community submission count

## Testing
Unit test and snapshot

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/Entry.tsx
+++ b/src/uniprotkb/components/entry/Entry.tsx
@@ -13,8 +13,6 @@ import {
   DropdownButton,
   Tabs,
   Tab,
-  Button,
-  CommunityAnnotationIcon,
 } from 'franklin-sites';
 
 import EntrySection, {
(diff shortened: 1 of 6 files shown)


src/uniprotkb/components/entry/Entry.tsx
```diff
@@ -63,6 +64,7 @@ import { EntryType } from '../../../shared/components/entry/EntryTypeIcon';
 
 import '../../../shared/styles/sticky.scss';
 import '../../../shared/components/entry/styles/entry-page.scss';
+import externalUrls from '../../../shared/config/externalUrls';
```
aurel-l, line 67: This needs to go higher

aurel-l (approved):
Feel free to merge whenever the import has been moved to the right position in the file (see other comment)

src/uniprotkb/components/entry/Entry.tsx
dlrice: good catch!

PR #405, opened 2021-06-28
header icons with bubbles

aurel-l:
## Purpose
add bubbles with counts in headers https://www.ebi.ac.uk/panda/jira/browse/TRM-26160

## Approach
Count items in basket
Count unseen jobs for dashboard (update job model)

## Testing
Updated tests

## Checklist


src/shared/components/layouts/SecondaryItems.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+import useBasket from '../../hooks/useBasket';
+
+import { pluralise } from '../../utils/utils';
+
+import { LocationToPath, Location } from '../../../app/config/urls';
+import { RootState } from '../../../app/state/rootInitialState';
+
+import styles from './styles/secondary-items.module.scss';
+import { Status } from '../../../tools/types/toolsStatuses';
```
dlrice, line 20:
```suggestion

import { RootState } from '../../../app/state/rootInitialState';
import { Status } from '../../../tools/types/toolsStatuses';

import styles from './styles/secondary-items.module.scss';
```

src/shared/components/layouts/SecondaryItems.tsx
```diff
@@ -0,0 +1,120 @@
...
+};
+
+export const Basket = () => {
+  const [basket] = useBasket();
+
+  const count = useMemo(
+    () =>
+      Array.from(basket.values())
+        .map((ns) => ns.size)
+        .reduce((total, current) => total + current, 0),
```
dlrice, line 73: My brain is not adept at understanding `reduce` so I tend to use lodash's `sum` in instances like this, but that's just personal preference

dlrice (approved):

src/shared/components/layouts/SecondaryItems.tsx
aurel-l: Not a big fan of reduce myself, and doing sums is the only acceptable case for me. But that's a good suggestion and going with lodash I went a bit further and used `sumBy` that allowed me to drop also drop the `.map`

PR #409, opened 2021-07-06
Duplicate xrefs bug

dlrice:
## Purpose
There is duplication of EnsemblBacteria cross-reference links in both the current site and the new site:

- https://www.uniprot.org/uniprot/P0A879#sequences
- https://beta.uniprot.org/uniprotkb/P0A879/entry#sequence

Which arises because of the inclusion of additional properties in the link
```
[
   {
      "database":"EnsemblBacteria",
      "id":"BAA14793",   // <-- 1
      "properties":[
         {
            "key":"ProteinId",
            "value":"BAA14793" // <-- 2
         },
         {
            "key":"GeneId",
            "value":"BAA14793" // <-- 3
         }
      ]
   }
]
```
The objective of this PR is to prevent duplicate links arising.
## Approach

- Create a list of xref link attributes for properties and the main link then check for duplicates
- Noticed and fixed a typo in `urlsAreEqual`. Tested this in the Downloads where it's used and is working as before anyway.

## Testing
- Created unit test

## Checklist


src/uniprotkb/components/protein-data-views/__tests__/XrefView.spec.tsx
```diff
@@ -1,26 +1,28 @@
...
 
-import xrefUIData from './__mocks__/xrefUIData.json';
+import xrefs from './__mocks__/xrefUIData';
 
 describe('XRefView', () => {
   test(`should render section`, () => {
-    const xrefs = xrefUIData as XrefUIModel[];
     const { asFragment } = render(
-      <XRefView xrefs={xrefs} primaryAccession="P01234" />
+      <XRefView xrefs={xrefs.normal} primaryAccession="P01234" />
```
aurel-l, line 18:
```suggestion
      <XRefView xrefs={xrefs.standard} primaryAccession="P01234" />
```
I think there was an issue here because the corresponding snapshot had become empty

aurel-l (approved):

PR #410, opened 2021-07-06
job dashboard

aurel-l:
## Purpose
Modify the job dashboard panel as specified in design meetings and in https://www.ebi.ac.uk/panda/jira/browse/TRM-26230

## Approach
- Add `seen` flag to all kinds of jobs instead of just finished
- Just use one colour with no number in the website header
- Click on `seen` indicator in dashboard to mark job as seen even without opening it
- Any change of URL when the panel is open will close it (Should probably eventually be handled by the SlidingPanel, but since now it's in another codebase I didn't want to touch it)

## Testing
Updated mocks, visual checks

## Checklist


src/tools/dashboard/components/Dashboard.tsx
```diff
@@ -34,6 +37,18 @@ const Dashboard = ({ inPanel }: { inPanel?: boolean }) => {
     }
   );
 
+  // All of this should probably part of the sliding panel logic
```
dlrice, line 40: Would be handy to have a TODO and jira reference for integrating these changes into franklin

src/shared/components/layouts/SecondaryItems.tsx
```diff
@@ -52,23 +36,16 @@ const getArrowX = (element: HTMLSpanElement) => {
...
-      unseenJobs.length,
-      // at the moment, will always be false
-      unseenJobs.some((job) => job.status === Status.FAILURE),
-    ];
-  });
+const ToolsDashboard = ({ display, close }: Props) => {
+  const count = useSelector<RootState, number>(
+    (state) =>
+      Object.values(state.tools).filter(
+        (job) => 'seen' in job && job.seen === false
```
dlrice, line 46:
```suggestion
     (job) =>
          'seen' in job &&
          job.seen === false &&
          [Status.FINISHED, Status.FAILURE, Status.ERRORED].includes(job.status)
```
Noticed that the green dot appears when I have in my jobs:
1. Successful and seen jobs
2. Running (unfinished) jobs

aurel-l: Ah, good catch, that was left over from the previous approach and I forgot to adjust this bit of code. I have now followed your suggestions (but extracting the statuses to check outside of this function) so that should be fixed now.

src/tools/dashboard/components/Dashboard.tsx
aurel-l: I'll create that Jira over the course of the day then 👍🏽

src/shared/components/layouts/SecondaryItems.tsx
dlrice: For some reason when mapping the same ID twice it won't remove the seen status when the duplicate job is viewed:
https://user-images.githubusercontent.com/4357962/125056859-912d8200-e0a0-11eb-9edd-834968d25a00.mov

dlrice: Looks like the it doesn't get marked as seen I'm guessing because the URL for both jobs is the same.

aurel-l: Addressed.
Now links to a job set a location state containing the internalID that this link refers to in order to only mark this specific one as read. In the really rare case where this state would not be set (if coming directly to the result URL but still owning the job, so with it being in the user's storage), it would mark all the jobs whose remoteID is the one in the URL as read.

aurel-l: Having @xwatkins review this as Dan is on holiday now. Dan reviewed the behaviour with me in a previous call, feel free to review only the implementation in this specific commit https://github.com/ebi-uniprot/uniprot-website/pull/410/commits/58bb93cdbf83ac66f335eff43f8fd3b3159a6075 as the rest as already been reviewed (or feel free to review more if you want! 😅 )

aurel-l: Resolving this comment and merging this PR in as it holds back on further development on the Basket because of refactoring being done on the different contents of the sliding panel

PR #411, opened 2021-07-07
2021 09 direct results

aurel-l:
## Purpose
Redirect to entry page https://www.ebi.ac.uk/panda/jira/browse/TRM-26172

## Approach
Different behaviour to the one in the Jira as difficult to define what is a "subsequent query".
Redirect to the entry page if only one result and if the search matches the accession or the ID of that result.

Also, fixed `?direct` approach that wasn't passed down the components as needed. Done by removing prop drilling and getting the value of that flag only where needed.

## Testing
Added tests for all 3 cases of redirecting to entry page. Testing redirect (through `history.location.pathname`)

Can check behaviour by searching for an accession ("P05067"), an ID ("KVD30_HUMAN", some ID searches like "A4_HUMAN" return multiple results so won't work for all IDs), or an ID of any namespace ("9606" in taxonomy).

## Checklist

### DIFF
--- a/src/shared/components/results/Results.tsx
+++ b/src/shared/components/results/Results.tsx
@@ -17,7 +17,7 @@ const Results = () => {
   const [selectedEntries, handleEntrySelection] = useItemSelect();
 
   // Query for facets
-  const { url: initialApiFacetUrl } = useNSQuery({
+  const initialApiFacetUrl = useNSQuery({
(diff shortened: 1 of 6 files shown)


xwatkins (approved):
Been comparing with the current website and I think your assessment of when the redirect should happen is correct.

PR #412, opened 2021-07-07
Increase jest test timeout.

dlrice:
Seeing an issue where our [tests are timing out](https://gitlab.ebi.ac.uk/uniprot/uniprot-website/uniprot-website/-/jobs/495940) on the gitlab CI runners I’m guessing because:

- not all runners have the specs
- the machines are shared and maybe another process is stealing resources

This PR just pushes the timeout to 30s.

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -6,7 +6,7 @@ const nodeCrypto = require('crypto');
 
 global.crypto = {
   ...global.crypto,
-  getRandomValues: function (buffer) {
+  getRandomValues(buffer) {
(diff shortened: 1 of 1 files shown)


jest.setup.ts
```diff
@@ -6,7 +6,7 @@ const nodeCrypto = require('crypto');
 
 global.crypto = {
   ...global.crypto,
-  getRandomValues: function (buffer) {
+  getRandomValues(buffer) {
```
dlrice, line 9: This change was automatic formatting

aurel-l (approved):

PR #413, opened 2021-07-08
Submission protein names

dlrice:
## Purpose
Jira: [Protein names missing "subNames"](https://www.ebi.ac.uk/panda/jira/browse/TRM-26211)

## Approach
Gene Names and Protein Names column renderers doing very similar things so:

1. Created `deepFindAllByKey` function to extract all of the `value` names for both genes and proteins
2. Created `CSVView` component to render this (including the missing submission names)

## Testing

- Create unit tests for `CSVView` and `deepFindAllByKey`
- Update `UniProtKBColumnConfiguration` snapshot tests to also check to see if there is a "more" button and if so click this and create a new snapshot from this.

## To view changes
As the jira states visit http://localhost:8080/uniprotkb?query=uncharacterized%20protein%20koala and look at the protein names column

## Checklist

### DIFF
--- a/src/shared/utils/utils.ts
+++ b/src/shared/utils/utils.ts
@@ -89,3 +89,20 @@ export const isSameEntry = (
   prev?.primaryAccession !== undefined
     ? prev.primaryAccession === next.primaryAccession
     : prev.id === next.id;
+
+export const deepFindAllByKey = (
+  o: Record<string, unknown> | Record<string, unknown>[],
+  predicateKey: string,
+  accum: string[] = []
+) => {
+  if (typeof o === 'object') {
+    Object?.entries(o).forEach(([key, value]) => {
+      if (key === predicateKey) {
+        accum.push(value as string);
+      } else {
+        deepFindAllByKey(value as Record<string, unknown>, predicateKey, accum);
+      }
+    });
+  }
+  return accum;
+};
--- a/src/shared/utils/utils.ts
+++ b/src/shared/utils/utils.ts
@@ -90,19 +90,21 @@ export const isSameEntry = (
     ? prev.primaryAccession === next.primaryAccession
     : prev.id === next.id;
 
-export const deepFindAllByKey = (
-  o: Record<string, unknown> | Record<string, unknown>[],
-  predicateKey: string,
-  accum: string[] = []
-) => {
-  if (typeof o === 'object') {
-    Object?.entries(o).forEach(([key, value]) => {
+export function* deepFindAllByKey<T = string>(
+  input: unknown,
+  predicateKey: string
+): Generator<T, void, never> {
+  if (Array.isArray(input)) {
+    for (const item of input) {
+      yield* deepFindAllByKey<T>(item, predicateKey);
+    }
+  } else if (input && typeof input === 'object') {
+    for (const [key, value] of Object.entries(input)) {
       if (key === predicateKey) {
-        accum.push(value as string);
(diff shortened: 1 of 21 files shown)


src/shared/utils/utils.ts
```diff
@@ -89,3 +89,20 @@ export const isSameEntry = (
...
+    Object?.entries(o).forEach(([key, value]) => {
+      if (key === predicateKey) {
+        accum.push(value as string);
+      } else {
+        deepFindAllByKey(value as Record<string, unknown>, predicateKey, accum);
+      }
+    });
+  }
+  return accum;
+};
```
aurel-l, line 108: If we have to do recursivity I'd rather avoid passing an accumulator around and mutate it, and instead would go for a generator.
See suggestion, feel free to use as is or modify.
```suggestion
export function* deepFindAllByKey<T = string>(
  input: unknown,
  predicateKey: string
): Generator<T, void, never> {
  if (Array.isArray(input)) {
    for (const item of input) {
      yield* deepFindAllByKey<T>(item, predicateKey);
    }
  } else if (input && typeof input === 'object') {
    for (const [key, value] of Object.entries(input)) {
      if (key === predicateKey) {
        yield value;
      } else {
        yield* deepFindAllByKey(value, predicateKey);
      }
    }
  }
}
```
Wasn't sure about the types, so I changed it a bit. Now accepting anything, and returning the correct type without having to cast inside the function. The expected type can optionally be set by the user on calling the generator, but assuming it is a string by default.

Note that for its use you'd need to wrap the current places where the function is called in this way: `Array.from(deepFindAllByKey(/* */))`.

Although... If the use case for this function is to only be used in the context of unique strings, then I'd keep the current approach, and directly use a `Set` as an accumulator, not an array, so that you don't have to wrap it with `lodash`'s `uniq`

src/uniprotkb/components/protein-data-views/__tests__/CSVView.spec.tsx
```diff
@@ -0,0 +1,13 @@
+import { screen, render } from '@testing-library/react';
+import CSVView from '../CSVView';
+
+describe('CSVView', () => {
+  it('should return nothing if no data provided', () => {
+    const { container } = render(<CSVView data={undefined} />);
+    expect(container.firstChild).toBeNull();
```
aurel-l, line 7:
```suggestion
    expect(container).toBeEmptyDOMElement();
```
Is also available for this

src/uniprotkb/config/__tests__/UniProtKBColumnConfiguration.spec.tsx
```diff
@@ -32,6 +34,13 @@ describe('UniProtKBColumnConfiguration component', () => {
         <>{column.render(uniProtKbConverter(data))}</>
       );
       expect(asFragment()).toMatchSnapshot(key);
+      const showMore =
+        screen.queryByTitle('Show more') ||
+        screen.queryByTestId('expandable-message');
+      if (showMore) {
+        showMore.click();
+        expect(asFragment()).toMatchSnapshot(`${key}-expanded`);
+      }
```
aurel-l, line 43: Nice addition to the tests. Would you mind adding this pattern to the other namespaces' column renderer tests as part of this PR? Not sure that it applies to many of the columns of the other namespaces, but some do and some might in the future, so it's good to have it part of the tests already.

aurel-l (approved):

PR #414, opened 2021-07-08
Fix interaction column renderer

xwatkins:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-24893
In the UniProtkb results table, the "interacts with" columns always displayed the "current" accession instead of the interactor.

## Approach
Add logic to pick either interactantOne or interactantTwo as the interactant to display.

## Testing
Updated the example data to reflect reality, as well as corresponding snapshots.

## Checklist

### DIFF
--- a/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
+++ b/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
@@ -671,6 +671,8 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccInteraction, {
 
     interactionComments?.forEach((interactionCC) =>
       interactionCC.interactions.forEach((interaction) => {
+        // Note: currently the interaction type is missing from the request.
+        // Check again when it has been added.
         if (
           interaction.type === InteractionType.SELF ||
           (interaction.interactantOne.uniProtKBAccession === primaryAccession &&
--- a/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
+++ b/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
@@ -71,12 +71,11 @@ import EntryTypeIcon, {
 
 import { getEntryPath } from '../../app/config/urls';
 import { fromColumnConfig } from '../../tools/id-mapping/config/IdMappingColumnConfiguration';
+import { sortInteractionData } from '../utils/resultsUtils';
 
 import { Namespace } from '../../shared/types/namespaces';
 import { ColumnConfiguration } from '../../shared/types/columnConfiguration';
 import AccessionView from '../../shared/components/results/AccessionView';
-
-import { Interactant } from '../adapters/interactionConverter';
 
 import helper from '../../shared/styles/helper.module.scss';
 
@@ -741,18 +740,7 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccInteraction, {
       })
     );
 
-    const sortedInteractions = Array.from(interactionDataMap.values()).sort(
-      (a, b) => {
-        if (typeof a !== 'string' && typeof b !== 'string') {
-          return (
-            (a as Interactant).geneName?.localeCompare(
-              (b as Interactant).geneName || ''
-            ) || -1
-          );
-        }
-        return -2;
-      }
-    );
+    const sortedInteractions = sortInteractionData(interactionDataMap);
 
     return (
       <ExpandableList displayNumberOfHiddenItems>
--- a/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
+++ b/src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
@@ -2,10 +2,12 @@
 import {
   EllipsisReveal,
   ExpandableList,
+  ExternalLink,
   LongNumber,
   Sequence,
 } from 'franklin-sites';
 import { Link } from 'react-router-dom';
+import { Fragment } from 'react';
 
 import SimpleView from '../../shared/components/views/SimpleView';
 import { ECNumbersView } from '../components/protein-data-views/ProteinNamesView';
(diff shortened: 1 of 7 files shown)


src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
```diff
@@ -703,33 +703,67 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccInteraction, {
...
+        })
+      )
+      .flat();
+
+    return (
+      interactionElements && (
+        <ExpandableList displayNumberOfHiddenItems>
+          {interactionElements.map((element) => element)}
+        </ExpandableList>
+      )
```
aurel-l, line 766:
```suggestion
      <ExpandableList displayNumberOfHiddenItems>
        {interactionElements}
      </ExpandableList>
```
ExpandableList takes care of checking if its children are truthy or not, if it's an empty array, or even if there's no children at all, so I don't think this extra code is needed

src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
```diff
@@ -703,33 +703,67 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccInteraction, {
...
-                      to={getEntryPath(
-                        Namespace.uniprotkb,
-                        interaction.interactantOne.uniProtKBAccession
-                      )}
-                    >
-                      {interaction.interactantOne.uniProtKBAccession}
-                    </Link>
+    const { primaryAccession } = data;
+
+    const interactionElements = interactionComments
```
aurel-l, line 708: This needs to return a unique set of links. I'd rather have it return a `Set` of accessions, or the specific string `'self'`, and then generate a link out of those. This should also remove the need for these complex keys in there, and the `div` look like they are not necessary

aurel-l (changes requested):
Screenshot 2021-07-09 at 09 16 40
It would be good to dedupe the accessions displayed in the cell, see related in-code comment.
The screenshot is from the search `"p21802"`

src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
```diff
@@ -703,33 +703,67 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccInteraction, {
...
+                  to={getEntryPath(
+                    Namespace.uniprotkb,
+                    interaction.interactantTwo.uniProtKBAccession
                   )}
-            </div>
-          ))
-        )}
-      </ExpandableList>
+                >
+                  {interaction.interactantTwo.uniProtKBAccession}
```
aurel-l, line 729: I see the entry type is in the data, maybe it would be good to display the reviewed/unreviewed icon maybe through the `src/shared/components/results/AccessionView.tsx` component?

xwatkins:
Made the suggested improvements and also added extra information for each interactant - however this means there still is the duplicate because of a different Intact ID. I've emailed Sandra to ask for her opinion on whether we should display the extra information or not (it won't be in the download).

xwatkins:
Heard back from Sandra, she says only the gene name would be useful, and asked if it could be sorted. Had forgotten `Set` only works on primitive types, so using a `Map` instead with the accession as `key`.

src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
aurel-l: Any opinion on this then?

xwatkins: Forgot to comment on this - I haven't seen "type" as a reviewed/unreviewed indication (at least for P05067), but it should be about the interaction type (binary, xeno or self). However I noticed that this information is not properly propagated from the flatfile, so contacted Leo and Jie about it and they are going to look into it as it should be there.

aurel-l: > I see the entry type is in the data
Ignore that, this is when I looked in the devtools, the alignment of the field names made me think it was part of the interaction comment I was seeing, but it's actually on the top-level entry.

aurel-l:
Now that we have the gene names in there it does look a bit strange when the column is narrow (happens quite quickly though):
Screenshot 2021-07-19 at 14 18 12
Should the gene names be grouped somehow (e.g. `<accesion> (<gene_name>)`, or `<accession> - <gene_name>) and making sure the line doesn't break by using the `.no-wrap` class in the helper css file?

Also, what happens if there are 2 genes of one entry interacting with the target entry? I would expect something like `<accession> (<gene1>, <gene2>)` but I'm not sure if that ever happens.

Last thing about "self". The current website renders the string "itself" and this PR renders the accession as a link. Not sure if that's a requirement I don't know, should it be "itself" too? or "itself" as a link to the entry, or "<accession> (itself)"?

src/uniprotkb/config/UniProtKBColumnConfiguration.tsx
```diff
@@ -703,31 +707,77 @@ UniProtKBColumnConfiguration.set(UniProtKBColumn.ccInteraction, {
...
+        if (typeof a !== 'string' && typeof b !== 'string') {
+          return (
+            (a as Interactant).geneName?.localeCompare(
+              (b as Interactant).geneName || ''
+            ) || -1
+          );
+        }
+        return -2;
+      }
+    );
```
aurel-l, line 755: It would be good to specify what is the sorting algorithm here, in plain English in a comment. As far as I can see in the implementation it gives:
- All without a gene name (in the order returned by the backend)
- "Itself"
- All with a gene name (in the order return by the backend)
Is that correct and expected?

xwatkins: In the end decided to add a test case - works as a better explanation than a comment.

aurel-l: Screenshot 2021-07-19 at 18 00 12
I'm still seeing this (for this search http://localhost:8080/uniprotkb?query=%22Q00266%22) which seems inconsistent that's why I was asking for the ordering in the first place.

aurel-l: Possible change here https://github.com/ebi-uniprot/uniprot-website/pull/419

gives me this in the previous example search

Screenshot 2021-07-19 at 18 25 02

xwatkins:
> Also, what happens if there are 2 genes of one entry interacting with the target entry? I would expect something like `<accession> (<gene1>, <gene2>)` but I'm not sure if that ever happens.

Checking with Leo but `geneName` is a string, so my guess is they just pick the first one...

aurel-l (approved):

PR #417, opened 2021-07-15
2021 08/09 basket

aurel-l:
## Purpose
Add basket content in the sliding panel https://www.ebi.ac.uk/panda/jira/browse/TRM-26226
Add basket content as full view https://www.ebi.ac.uk/panda/jira/browse/TRM-26121
Align button https://www.ebi.ac.uk/panda/jira/browse/TRM-26123
BLAST button https://www.ebi.ac.uk/panda/jira/browse/TRM-26124
Download button https://www.ebi.ac.uk/panda/jira/browse/TRM-26125
ID Mapping button https://www.ebi.ac.uk/panda/jira/browse/TRM-26122

## Approach
Try to reuse parts of the Results page approach, including some hooks that need to be overridden in some aspects.
Unify approach of Dashboard and Basket in sliding panel.
Reuse search result buttons & other components
Add ID Mapping button in a similar way as the other action buttons
Rename all namespaceFallback and columnsFallback to namespaceOverride and columnsOverride because it is manually overriding whatever value the logic would have given by default, rather than falling back on this value if the logic were unable to determine anything.

## Testing
Only tested interaction and behaviour manually
Updated snapshots

## Checklist
  - yeah, no, at beginning yes, but not anymore
  - Reused and refactored a lot of code, so coverage didn't fall too much even though no test was added

Note: buggy `useItemSelect` when deleting items (still kept in the internal state), or when switching namespace in the basket -> need to be rewritten more like a proper `useState` hook to be able to remove stuff from it (created Jira)


src/basket/svgs/empty-basket.svg
```diff
@@ -0,0 +1,45 @@
+<svg width="295" height="190" viewBox="-25 0 320 190" fill="none" xmlns="http://www.w3.org/2000/svg">
```
xwatkins, line 1: Shouldn't this be in franklin?

src/basket/styles/basket-mini-view.module.scss
```diff
@@ -0,0 +1,28 @@
+@import 'franklin-sites/src/styles/settings';
```
xwatkins, line 1: Maybe for another time, but I wonder if there should be modifiers defined in franklin for this. We already have the "compact" table.

aurel-l:
> My opinion would be to not have the "Delete" from basket option available on the full view, and treat it simply as an accession search.

So also remove the remove/bin icon column there?

xwatkins (commented):
> > My opinion would be to not have the "Delete" from basket option available on the full view, and treat it simply as an accession search.
> 
> So also remove the remove/bin icon column there?

I would say so - another option would be to "reset" the view if there are no results in the current result set, but that might be a bit weird

src/shared/hooks/useNSQuery.ts
```diff
@@ -44,7 +46,7 @@ const useNSQuery = ({
     getParamsFromURL(queryParamFromUrl);
 
   const url = useMemo(() => {
-    if (!query && !accessions.length) {
+    if (!(query || accessions)) {
```
xwatkins, line 49: Do we want to also check that `accession` is not an empty list?

src/basket/svgs/empty-basket.svg
aurel-l: no-blast-result was in uniprot website so I put that one here too

aurel-l: We also have all the illustrations in `src/images`. I'd assume icons in franklin, big specific illustrations in uniprot-website

src/basket/styles/basket-mini-view.module.scss
aurel-l: For the settings? I'm not sure what you are referring to

xwatkins (commented):
I mean most of the file - it seems that it's related to making the table even more compact, removing margins etc... so thought it could be something useful in other places too and could be part of the component definition.

src/basket/styles/basket-mini-view.module.scss
aurel-l: > I mean most of the file - it seems that it's related to making the table even more compact, removing margins etc... so thought it could be something useful in other places too and could be part of the component definition.

Following the thread here otherwise it's hard to follow.

The issue is that it's overriding styles in sliding-panel, cards, tabs (tabs headers specifically), data-table, and button-groups. I agree the content of this file shouldn't have to exist, but I don't know where else we'd put it as it covers so many different franklin components interacting with each other, and just for this specific basket use case.

src/shared/hooks/useNSQuery.ts
aurel-l: I removed that at some point because when accessing an empty basket, it would end up returning the URL for a star query somehow. But that is not the case anymore, because I'm just not even rendering a table when the basket (or subbasket) is empty, and also probably because I've touched some other code somewhere else I guess.
I have now modified as suggested.

src/shared/components/results/ResultsButtons.tsx
```diff
@@ -132,9 +142,9 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
             })}
           />
         </Button>
-        {viewMode === ViewMode.TABLE && namespace !== Namespace.idmapping && (
-          <CustomiseButton />
-        )}
+        {!notCustomisable &&
+          viewMode === ViewMode.TABLE &&
+          namespace !== Namespace.idmapping && <CustomiseButton />}
```
xwatkins, line 147: I think we should be consistent and either use `!notCustomisable` everywhere (id mapping included, which would involve passing the prop through `ResultsDataHeader`, or only by using the namespace (although this might be more tricky?)

src/shared/components/results/ResultsData.tsx
```diff
@@ -108,6 +119,35 @@ const ResultsData: FC<{
     prevViewMode.current !== viewMode ||
     prevNamespace.current !== namespace;
 
+  const finalColumns = useMemo(() => {
```
xwatkins, line 122: This use case is very similar to the ID Mapping "extra" column, so I wonder if it would be possible to put that logic in the same place in `useColumns`? see the use of the `displayIdMappingColumns` flag here: https://github.com/ebi-uniprot/uniprot-website/pull/417/files?file-filters%5B%5D=.tsx&hide-deleted-files=true#diff-2574fa071caa07abbdde4d341fd363f480b8b0d44fbbe032721f33b5c18d57f4R146

src/shared/components/results/ResultsButtons.tsx
aurel-l: Changed to use notCustomisable everywhere needed.
In the same commit, I fixed an issue with the ID mapping download not working for "all"

src/shared/components/results/ResultsData.tsx
aurel-l: Done

xwatkins (approved):

PR #418, opened 2021-07-19
expose HSPs to BLAST user in the form

aurel-l:
## Purpose
Expose HSPs choice to a BLAST user https://www.ebi.ac.uk/panda/jira/browse/TRM-26277

## Approach
Add a parameter to the existing ones

## Testing
Update snapshots

## Checklist


src/tools/blast/components/BlastForm.tsx
```diff
@@ -278,6 +283,7 @@ const BlastForm = () => {
       gapped: gapped.selected as GapAlign,
       // transform string into number
       hits: parseInt(hits.selected as string, 10) as Scores,
+      hsps: hsps.selected as HSPs,
```
dlrice, line 286:
```suggestion
      hsps: parseInt(hsps.selected, 10) || undefined as HSPs,
```
This string "undefined" is passed to the job parameters resulting  in a 400:
```
<?xml version='1.0' encoding='UTF-8'?>
<error>
 <description>For input string: "undefined"</description>
</error>
```

dlrice (commented):
This may have already been discussed and I've forgotten the answer: should we have hits:=1 by default as this what uniprot.org is doing and this what users would expect?

src/tools/blast/components/BlastForm.tsx
aurel-l: Mhhh, that's strange, I'll have a look

dlrice: my previous suggestion had type issues
```suggestion
      hsps: (hsps.selected === 'undefined'
        ? undefined
        : parseInt(hsps.selected as string, 10)) as HSPs,
```

aurel-l:
> This may have already been discussed and I've forgotten the answer: should we have hits:=1 by default as this what uniprot.org is doing and this what users would expect?

Actually we have found that by default this is not set at all in the current website, even though they only display the first one (you'd need to click on the alignment to see the others). So that's why we kept the current default (unset) instead of going for "1".
We found out why the jobs were taking so long, it was a bug in another parameters that was causing this, unrelated to the number of HSPs requested.
With the same default values, the new website is now noticeably quicker than the current one on getting job results

src/tools/blast/components/BlastForm.tsx
dlrice: third times a charm 😅 
```suggestion
      hsps: (parseInt(hsps.selected as string, 10) || undefined) as HSPs,
```

aurel-l: All good now, good catch, I hadn't try going back and forth between options when testing, and thanks for the suggestion

dlrice (approved):

PR #419, opened 2021-07-19
proposal for a sort change

aurel-l:
Sort change


xwatkins (approved):
Had just got to the same conclusion as you regarding the different default return of 1 and SELF of -1. Liking the restriction of `InteractionType.SELF` instead of string. I will add a note to check all this again once backend has fixed the type.

PR #420, opened 2021-07-23
2021 09 help results

aurel-l:
## Purpose
Help result page https://www.ebi.ac.uk/panda/jira/browse/TRM-26244

## Approach
Routed through /help?query=protein (not linked anywhere yet, need to manually go there).
Passing markdown through `marked` and then reusing sanitize-html in order to process the resulting html before injecting it through react's dangerouslySetInnerHTML.
After conversation with Xavier, trying to keep it as independent to the rest of the code as possible (tried to not modify other logic to accommodate help, but try to reuse external logic anyway, just without touching it much).
Visual style of highlight validated by Marija already.

## Testing
Manually go to help page and search at the moment
Small unit test, but will need to add more in the future when add mock (but when the API is stable)

## Checklist
- [X] My PR is scoped properly, and “does one thing only”
- [X] I have checked that linting checks pass and type safety is respected

Anything related to scoring and returned results has to do with backend, planned discussion already

I've added a basic entry page for help articles, just to have somewhere to point the link to, but it's not part of this PR and related jira


src/shared/utils/cleanText.ts
```diff
@@ -13,22 +13,22 @@ import styles from './styles/clean-text.module.scss';
 // List of tags to remove from the library default accepted set
 const excludedTags = new Set(['a']);
 
-export const cleanTextDefaultOptions: IOptions = {
+export const cleanTextDefaultOptions = Object.freeze<IOptions>({
```
dlrice, line 16: We often want to freeze nested objects - shall we just go ahead and

aurel-l: Agree, will tackle that in the follow up PR about entry pages though as I've made further modifications including Object.freeze there too.
Will add `deep-freeze` as a dependency and replaces all places using Object.freeze with it

aurel-l: Handled in https://github.com/ebi-uniprot/uniprot-website/commit/82a1abde5fa648f057a6095d52c9d5c93854001d and https://github.com/ebi-uniprot/uniprot-website/commit/b06bf62e22320c865320bce6f83070007ec80869 (other branch)

src/help/components/results/HelpCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const HelpCard = ({ id, title, titleMatch, contentMatch }: Props) => {
+  const to = generatePath(LocationToPath[Location.HelpEntry], {
+    accession: id,
+  });
+
+  return (
+    <Card
+      header={
+        <h2 className="tiny">
```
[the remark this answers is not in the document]
aurel-l: For this I put a note in the design meeting for tomorrow in order to not forget to discuss it. I think I remembered Marija was keen on having that as a link, but your observation about consistency makes sense. I kept it as a link for now.

src/shared/components/search/SearchContainer.tsx
```diff
@@ -94,6 +94,9 @@ const SearchContainer: FC<
       const { query } = queryString.parse(history.location.search, {
         decode: true,
       });
+      if (history.location.pathname.includes('help')) {
```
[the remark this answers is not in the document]
aurel-l: Yep, done here and lower in the same file

PR #425, opened 2021-08-05
2021 10 organism not sent id mapping

dlrice:
## Purpose
- Taxon ID not sent when submitting ID mapping form 
- [Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26384)

## Approach
- Include `taxId` in `serverParametersToFormParameters` for ID mapping
- When updating tests a warning arised about updating an unmounted component so updated AutocompleteWrapper to be a functional component and to `useSafeState` to resolve
- Removed duplicate mock endpoints `IDMappingResult.spec.tsx`

## Testing
- Unit test

## Checklist


src/query-builder/components/AutocompleteWrapper.tsx
```diff
@@ -32,121 +33,89 @@ type SelectValue = {
...
+const minAPISuggesterChars = 3;
 
-const minCharsToShowDropdown = 3;
+export const prepareData = (suggestions: Suggestion[]) =>
+  suggestions.map(
+    (suggestion: Suggestion): SelectValue => ({
+      pathLabel: `${suggestion.value} [${suggestion.id}]`,
+      itemLabel: suggestion.value,
+      apiId: suggestion.id,
+      id: v1(),
```
aurel-l, line 44: It looks like in Franklin it's used to set a react key. Not asking for that to be modified at this point, but I'm wondering if you know why this couldn't be `suggestion.id`? Could it be the case that 2 suggestions might have the same ID?

src/query-builder/components/AutocompleteWrapper.tsx
```diff
@@ -32,121 +33,89 @@ type SelectValue = {
...
+    (textInputValue: string) => {
+      if (textInputValue.length < minAPISuggesterChars) {
+        setData([]);
+      } else {
+        const suggesterUrl = getSuggesterUrl(url, textInputValue);
+        setLoading(true);
+        fetchData<Suggestions>(suggesterUrl)
+          .then((response) => prepareData(response.data.suggestions))
+          .then((responseData) => {
+            setData(responseData);
```
aurel-l, line 81:
```suggestion
          .then((response) => {
            setData(prepareData(response.data.suggestions));
```

src/query-builder/components/AutocompleteWrapper.tsx
dlrice: According to backend we shouldn't have to worry about duplicate IDs being returned so using this for the ID now.

aurel-l (approved):

PR #427, opened 2021-08-13
2021 09 subcell viz go tab

dlrice:
## Purpose
Add GO tab with corresponding sub-cellular location visualisation to entry page [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-25568)]

## Approach
- Modify `Tab(s)` component in franklin ([PR](https://github.com/ebi-uniprot/franklin-sites/pull/218)) to allow some tabs to be "cached" - in document but not visible - to prevent having to re-create `@swissprot/swissbiopics-visualizer` and reattach the listeners, tooltips, etc.
- Extract GO data from cross-references.
- Create `uniProtLocationIds` & `goLocationIds` as CSV string in `SubcellularLocationWithVizView` rather than in `SubcellViz` and only pass the appropriate one depending on tab.
- Added `GoEvidenceType` for the future

## To do in future work
- **Location categories**: create and consume sub-cellular location to category mapping [[TRM-26463](https://www.ebi.ac.uk/panda/jira/browse/TRM-26463)]
- **Add source tags**: this will form part of the existing jira [[TRM-23931](https://www.ebi.ac.uk/panda/jira/browse/TRM-23931)]

## Testing
Update snapshots and add unit tests

## Checklist


src/uniprotkb/components/protein-data-views/SubCellViz.tsx
```diff
@@ -1,11 +1,13 @@
...
 import { FC, memo, useEffect, useRef } from 'react';
 import tippy from 'tippy.js';
 import '@swissprot/swissbiopics-visualizer';
 
 import 'tippy.js/dist/tippy.css';
 import './styles/sub-cell-viz.scss';
 
-import { SubcellularLocationComment } from '../../types/commentTypes';
+import { RequireExactlyOne } from 'type-fest';
+import { VizTab } from './SubcellularLocationWithVizView';
```
aurel-l, line 10: These need to go higher

aurel-l:
Looking good 👍🏽 
Left a comment about the position of an import, nothing more than that.

Comparing with the current website, I'm wondering if we need the `Complete GO annotation on QuickGO ...` link too? It should be fairly easy to add.

src/uniprotkb/components/protein-data-views/SubCellViz.tsx
dlrice: Fixed now

dlrice:
> Comparing with the current website, I'm wondering if we need the `Complete GO annotation on QuickGO ...` link too? It should be fairly easy to add.
Good idea, I didn't even notice it on the current site. Added it in now with an `<hr/>` to separate it from the list of locations.

aurel-l (approved):
Looking good 👍🏽

PR #431, opened 2021-08-18
useTextFileInput: reset input ref to handle the case that a user loads the same file twice

dlrice:
## Purpose
Bug fix: useTextFileInput unresponsive to loading same file twice [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26469)]

## Approach
`inputRef.current.value = ''` inside the hook

## Testing
none

## Checklist

### DIFF
--- a/src/shared/hooks/useTextFileInput.ts
+++ b/src/shared/hooks/useTextFileInput.ts
@@ -34,39 +34,47 @@ const useTextFileInput = ({
   const onErrorRef = useRef(onError);
   onErrorRef.current = onError;
 
-  const handleNewFile = useCallback((file?: File) => {
-    if (!file) {
-      return;
-    }
-    let didCancel = false;
+  const handleNewFile = useCallback(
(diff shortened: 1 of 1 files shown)


src/shared/hooks/useTextFileInput.ts
```diff
@@ -34,39 +34,47 @@ const useTextFileInput = ({
...
+                `The file "${file.name}" doesn't appear to be in the correct format. It needs to be a plain text file.`
+              )
+            );
+          } else {
+            onFileContentRef.current(text);
+            // Reset input ref to handle the case that a user loads the same file twice
+            if (inputRef?.current) {
+              // eslint-disable-next-line no-param-reassign
+              inputRef.current.value = '';
+            }
```
dlrice, line 62: There's a whole bunch of auto-formatting in this PR - these are the only lines that really changed

aurel-l (approved):
Nice 👍🏽

PR #434, opened 2021-08-20
no link to results page when no results

aurel-l:
## Purpose
Don't let the user go to a results page if there are no results https://www.ebi.ac.uk/panda/jira/browse/TRM-26302 & https://www.ebi.ac.uk/panda/jira/browse/TRM-26345

## Approach
Remove links in the status text, the result card, and the popup

## Testing
Manual testing with P05635 to Allergome (ID Mapping), and P80488 (BLAST)

## Checklist

### DIFF
--- a/src/tools/dashboard/components/Row.tsx
+++ b/src/tools/dashboard/components/Row.tsx
@@ -163,10 +163,9 @@ const NiceStatus = ({ job, jobLink }: NiceStatusProps) => {
     case Status.NOT_FOUND:
       return <>Job not found on the server</>;
     case Status.FINISHED: {
-      const status = jobLink ? (
-        // eslint-disable-next-line uniprot-website/use-config-location
-        <Link to={jobLink}>Completed</Link>
-      ) : null;
+      if (!jobLink) {
+        return null;
+      }
       // either a BLAST or ID Mapping job could have those
       if ('data' in job && job.data && 'hits' in job.data) {
         const actualHits = job.data.hits;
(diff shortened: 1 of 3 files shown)


dlrice (commented):
As far as I have seen in my testing the number of hits is displayed only in the case that there are no hits:
```
Completed (0 hits)
```
and this made me expect jobs that ran with hits/results to also display a number here eg `Completed (2 hits)`. Maybe using
```
Completed: no results found
```
would be a little more consistent or should we always return the number of hits/results?

aurel-l:
> As far as I have seen in my testing the number of hits is displayed only in the case that there are no hits:
> 
> ```
> Completed (0 hits)
> ```
> 
> and this made me expect jobs that ran with hits/results to also display a number here eg `Completed (2 hits)`. Maybe using
> 
> ```
> Completed: no results found
> ```
> 
> would be a little more consistent or should we always return the number of hits/results?

I changed "0 hits" to "No results found".
Regarding not showing the number of hits everytime that was decided in the past as to show only numbers when different compared to expected value:
 - request 250 hits, got 250 hits -> all good, show "Completed"
 - request 1000 hits, got 250 hits -> something might be off, show "Completed (250 hits)" out of the 1000 requested.
This is to show that here if we were to show 250 hits for both results a user might not notice that something might be off in the 2nd one.

src/tools/dashboard/components/Row.tsx
```diff
@@ -183,11 +182,16 @@ const NiceStatus = ({ job, jobLink }: NiceStatusProps) => {
...
+                <span>Completed</span>
+              ) : (
+                // eslint-disable-next-line uniprot-website/use-config-location
+                <Link to={jobLink}>Completed</Link>
+              )}{' '}
               <span
                 title={`${actualHits} ${hitText} results found instead of the requested ${expectedHits}`}
               >
-                ({job.data.hits} {hitText})
+                ({actualHits ? `${actualHits} ${hitText}` : 'No results found'})
```
dlrice, line 194:
```suggestion
                ({actualHits ? `${actualHits} ${hitText}` : 'no results found'})
```
to yield `Completed (no results found)`. I'm being pedantic here but I just find only having the `Completed` capitalized a little easier on the eyes - maybe because i'm reading it like a sentence.

aurel-l: I see, yep, looks better actually

dlrice (approved):

PR #437, opened 2021-08-26
2021 11 entry history

aurel-l:
## Purpose
UniProtKB entry history section https://www.ebi.ac.uk/panda/jira/browse/TRM-25989

## Approach
 - Added as a new subpage for a UniProtKB entry
 - Use UniSave API
   - Added UniSave endpoints to the file with all the URLs, and added UniSave types
 - Use a library to view diff (similar to GitHub diff view), use the whole content of the file instead of the API provided diff, to be able to see the diff in context of the file or to use different diffing algorithms if needed.
 - Fixed the endless messages when redirecting a merged entry (just once)
 - Fixed other tabs when redirecting a merge entry => I'm actually redirecting so that the data is correct in the other tabs too.
 - Reused and modified existing components for obsolete entry messages

## Testing
Updated current snapshots, coverage barely lowered so kept it like this for now to finish up the sprint (had to slightly lower functions coverage)

## Checklist
  - One _big_ thing

Examples I used:
 - Normal entry: P05067
 - Demerged entry: Q00007
 - Merged entry: B2RBU8
 - Deleted entry: I8FBX0
 - Obsolete entry with no history: P50409

### DIFF
--- a/src/uniprotkb/types/uniSave.ts
+++ b/src/uniprotkb/types/uniSave.ts
@@ -27,3 +27,17 @@ export type UniSaveStatus = {
   accession: string;
   events: UniSaveEvent[];
 };
+
+export type DiffEntry = {
+  content: string;
+  entryVersion: number;
+};
+
+export type UniSaveDiff = {
+  accession: string;
+  diffInfo: {
+    diff: string;
+    entry1: DiffEntry;
+    entry2: DiffEntry;
+  };
+};
--- a/src/uniprotkb/types/uniSave.ts
+++ b/src/uniprotkb/types/uniSave.ts
@@ -1,0 +1,29 @@
+export type UniSaveVersion = {
+  accession: string;
+  database: string; // 'Swiss-Prot' | 'TrEMBL';
+  entryVersion: number;
+  firstRelease: string; // release version
+  firstReleaseDate: string; // DD-MMM-YYYY
+  lastRelease: string; // release version
+  lastReleaseDate: string; // DD-MMM-YYYY
+  name: string;
+  sequenceVersion: number;
+  content?: string;
+};
+
+export type UniSaveAccession = {
+  results: UniSaveVersion[];
+};
+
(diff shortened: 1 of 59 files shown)


src/uniprotkb/types/uniSave.ts
```diff
@@ -0,0 +1,43 @@
+export type UniSaveVersion = {
+  accession: string;
+  database: string; // 'Swiss-Prot' | 'TrEMBL';
```
dlrice, line 3:
```suggestion
  database: 'Swiss-Prot' | 'TrEMBL';
```
Any reason to not narrow it down now?

dlrice (changes requested):
Really liking the improved diff view. A few comments:

1. Noticed a [comment](https://www.ebi.ac.uk/panda/jira/browse/TRM-25989?focusedCommentId=363875&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-363875) in the jira:

> As soon as there are two selected, ‘Compare’ button becomes active, when there are more selected, ‘Compare’ button blurs out
> 
> ‘Compare’ can go next to ‘Download’

Which I couldn't see in this PR. In addition on the current site there is also the ability to select and download all.

2. This is probably more of an issue to discuss with the back-end:
When looking at http://localhost:8080/uniprotkb/P50409/entry I see:

```
It can now be found as secondary accession in P63150, Q00007, and P63151. [List]
```

Clicking on the list link takes me to the UniProtKB query results `sec_acc:P50409` which lists:

```
P63151
P63150
```

While I was expecting Q00007 to be included here too?

src/uniprotkb/types/uniSave.ts
aurel-l: Nope, at the beginning I was just worried some old data might be different, but looks like it's always those 2, so I'll take your suggested change

dlrice (approved):

PR #438, opened 2021-09-08
2021 12 help center autocomplete

dlrice:
## Purpose
Create a means to quickly search while not leaving the help center landing page. [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26243)]

## Approach
- Created HelpQuickSearch
- Limited `numberResultsInView` to 5
- Save the user query in history state
- Modified `helpURL.search` to accept parsed (rather than encoding in the `HelpQuickSearch` as it would then be immediately parsed within `helpURL.search`)
- Created `CleanHighlightMarkDown` to be shared in both the full results and the quick search result
- Replaced color codes with `$colour-gainsborough`

## Testing
- Snapshot and unit test
- Added mock help search data with jira to update when API is stable

## Checklist


src/help/components/landing/HelpQuickSearch.tsx
```diff
@@ -0,0 +1,122 @@
...
+      }
+    },
+    [allArticlesLocation, history]
+  );
+
+  const allArticles = dataObject?.data?.results;
+  const infoData = allArticles
+    ?.slice(0, numberResultsInView)
+    .map(({ matches, title, id }) => {
+      const titleMatch = matches?.title?.[0];
```
xwatkins, line 77: Are we sure there will always be at least 1 title, or should we check whether the array is empty? Same question for content.

src/help/components/landing/HelpQuickSearch.tsx
```diff
@@ -0,0 +1,122 @@
...
+      />
+      {!!allArticles?.length && !!infoData?.length && searchValue && (
+        <Card>
+          <InfoList
+            className={styles['help-autocomplete__result-list']}
+            infoData={infoData}
+          />
+          <div className={styles['help-autocomplete__all-link']}>
+            <Link to={allArticlesLocation}>
+              Show all results ({allArticles.length})
```
xwatkins, line 113: Should we not show this if there's only 1 result?

src/help/components/landing/styles/help-quick-search.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  &__result-list {
+    max-height: 30rem;
+    overflow-y: auto;
+  }
+
+  &__all-link {
+    text-align: right;
+  }
+
+  :global(.search-input__suffix) {
```
xwatkins, line 13: I guess they have to be `:global` to style Franklin components, but it looks like `.search-input__suffix` is also defined globally here https://github.com/ebi-uniprot/uniprot-website/pull/438/files?file-filters%5B%5D=.scss&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-3243e9ae200e982d2c3fe06ca36dd4fd9f387b7036bef39d8ba05f4292c9174bR21

src/help/components/results/styles/highlight.scss
```diff
@@ -1,6 +1,6 @@
 @import 'franklin-sites/src/styles/colours';
 
-.help-card :global(.match-highlight) {
+.match-highlight {
```
xwatkins, line 3: Does this not need `:global` ?

aurel-l: This is because it's not a CSS module file anymore, so the class name wouldn't be changed.

The reason I put a CSS module class name in the original rule was to make sure to scope this rule somehow, in case something else somewhere else in the logic would end up having the same class name injected.

src/help/components/landing/HelpQuickSearch.tsx
aurel-l: I think this refers to the last thread on the Slack #ux channel so options have been discussed there already

src/help/components/results/styles/highlight.scss
dlrice: As `highlight.scss` is now imported only within `CleanHighlightMarkDown.tsx` this should be OK right or am I polluting the global namespace?

src/help/components/landing/HelpQuickSearch.tsx
xwatkins: I don't think there's any point in going "full view" for 1 result as the facets don't add any functionality in that case...

src/help/components/results/styles/highlight.scss
xwatkins: No that's ok.

src/help/components/landing/HelpQuickSearch.tsx
dlrice: My thinking was to do it for consistency as there's only two ways to get to the full results:

1. Pushing enter
2. Clicking on the full view link

Route 1 may not always be obvious and it could be frustrating if the user doesn't always see the link because of the number of results.

xwatkins: There's only 1 result though, so in the user workflow nowhere else to go than to that entry?

dlrice: I'm thinking of the instance when the user is returned a single result which isn't what they're looking for and they're thinking "I want the full view with the categories on the side - how do I get there? There used to be a link at the bottom in my other searches."

aurel-l: I wouldn't be too attached to one approach or the other. Once either is implemented, we can easily play with it, have a feel, find edge-case examples, and discuss in a design meeting. This is easily changeable one way or the other if needed

xwatkins: 👌

src/help/components/landing/HelpQuickSearch.tsx
dlrice: In the return of the map fn there is check for `titleMatch` and if it's not there just return the article title:
```
title: (
  <Link to={to}>
    {titleMatch ? <CleanHighlightMarkDown md={titleMatch} /> : title}
  </Link>
),
```
and in our API model we have `title` as required - could place additional checks if you think it's needed?

For the content we also check if there is a match but in this case it just returns undefined and the whole info list row isn't rendered. I can instead do:
```
content: contentMatch ? (
  <CleanHighlightMarkDown md={contentMatch} />
) : (
  ' '
),
```
so it will at least return the title like this (hardcoded to see what it would look like):
Screenshot 2021-09-10 at 12 07 34

xwatkins:
image
Couple of other issues:
-  the blue line to the left of the text body doesn't always go all the way down [minor]
- some html doesn't seem to be cleaned up (see screenshot)

dlrice:
> * some html doesn't seem to be cleaned up (see screenshot)
The code here corresponds to some perl from the [manual](https://github.com/ebi-uniprot/uniprot-manual/wiki/api_downloading) so this shouldn't be filtered out

src/help/components/landing/styles/help-quick-search.module.scss
dlrice: As discussed in slack it's OK to duplicate this in two separate places

dlrice:
All comment should now be addressed

xwatkins (approved):

src/help/components/landing/styles/help-quick-search.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  :global(.info-list) {
+    margin: 0rem;
+  }
+
+  :global(.decorated-list-item) {
+    padding: 0.5rem;
+  }
+
+  :global(.decorated-list-item__title) {
+    flex-basis: 25vw;
```
aurel-l, line 35: Screenshot 2021-09-23 at 08 18 10
This gives strange layout when the screen is wide, the screenshot is for very wide screen.
I'd say use `25%` instead, so that it stays with a basis of 25% of the width of container

dlrice: Yeah, looks nicer. Can you think of a way to get the infolists's separating bar to have the height of `max(height title, height content)`?
Screenshot 2021-09-23 at 08 37 30

aurel-l: I would do `align-self: stretch` on `:global(.decorated-list-item__content)` a bit lower in the same file

dlrice: Nice! I think this is the intended style so we might as well put this in franklin

aurel-l: If you put that in franklin have a look if you can have a more generic rule than this override. using `align-items: stretch` on the flex container instead of the specific child for example. And what out for side effects in different use cases (thinking specifically of the 2 columns approaches)

dlrice: I can't see any unwanted side-effects so pushed this to franklin's main.

PR #441, opened 2021-09-13
2021 12 rule column taxonomic scope

aurel-l:
## Purpose
UniRule/ARBA results, taxonomic scope column https://www.ebi.ac.uk/panda/jira/browse/TRM-26591

## Approach
- Extract taxonomy info from the rule's main rule
- Dedupe taxons in a map
- Use a similar approach as other result columns with list taxon data (ExpandableList + TaxonomyView)
- Display negative taxonomy scope (wasn't in current website, but might be useful), happens only in UniRule as far as I can see

## Testing
Visual test + snapshot render from mock data

## Checklist


src/automatic-annotations/shared/column-renderers/TaxonomicScope.tsx
```diff
@@ -0,0 +1,58 @@
...
+}: Partial<UniRuleAPIModel | ARBAAPIModel>) => {
+  // Using a map in order to remove duplicates
+  const taxonScopeMap = new Map<number, CustomTaxonScope>();
+  for (const conditionSet of mainRule?.conditionSets || []) {
+    for (const condition of conditionSet.conditions || []) {
+      if (condition.type !== 'taxon') {
+        continue; // eslint-disable-line no-continue
+      }
+      for (const conditionValue of condition.conditionValues || []) {
+        // This shouldn't happen
```
dlrice, line 27: Would this imply something is wrong with the data? Would a `console.warn` or `console.assert` be useful in that case?

src/automatic-annotations/shared/column-renderers/TaxonomicScope.tsx
```diff
@@ -0,0 +1,58 @@
...
+        /* istanbul ignore if */
+        if (!conditionValue.cvId) {
+          continue; // eslint-disable-line no-continue
+        }
+        const taxonId = +conditionValue.cvId;
+        taxonScopeMap.set(taxonId, {
```
dlrice, line 33: Just an idea, check if `taxonId` is already present and if so skip it to avoid rewriting (presumably) the same thing? Unless this would mess up the order?

dlrice (approved):

src/automatic-annotations/shared/column-renderers/TaxonomicScope.tsx
aurel-l: Not sure if that ever happens to be honest, but it's to do type guarding for TypeScript to be happy and avoid possible bugs if it does happen. I added the console.warn as suggested

src/automatic-annotations/shared/column-renderers/TaxonomicScope.tsx
aurel-l: I added a check to avoid re-setting a new object if already seen this taxon. Again, not sure if that would ever happen, or even make sense in the context of the rule, but oh well...

dlrice: I've just consoled this and it does happen though admittedly not enough (~3% of rows in table view) to make a difference.

aurel-l: I'll have a look then, to try to understand what it means and if it's just some edge-cases or something we need to tackle more specifically.

PR #444, opened 2021-09-14
2021 12 rule column annotation covered

aurel-l:
## Purpose
Add annotation covered column to UniRule and ARBA https://www.ebi.ac.uk/panda/jira/browse/TRM-26592

## Approach
Common renderer for both namespaces (see TaxonomicScope approach)

## Testing
Update snapshots

## Checklist

### DIFF
--- a/src/automatic-annotations/shared/column-renderers/AnnotationCovered.tsx
+++ b/src/automatic-annotations/shared/column-renderers/AnnotationCovered.tsx
@@ -1,0 +1,35 @@
+import { ExpandableList } from 'franklin-sites';
+
+import { ARBAAPIModel } from '../../arba/adapters/arbaConverter';
+import { UniRuleAPIModel } from '../../unirule/adapters/uniRuleConverter';
+
+const AnnotationCovered = ({
+  mainRule,
+}: Partial<UniRuleAPIModel | ARBAAPIModel>) => {
+  const annotations = new Set<string>();
+
+  for (const annotation of mainRule?.annotations || []) {
+    if ('keyword' in annotation) {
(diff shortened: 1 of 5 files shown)


src/automatic-annotations/shared/column-renderers/AnnotationCovered.tsx
```diff
@@ -0,0 +1,35 @@
...
+      annotations.add('GO (Gene Ontology) term');
+    } else {
+      // in case we're missing a case
+      console.warn(annotation); // eslint-disable-line no-console
+    }
+  }
+
+  return (
+    <ExpandableList descriptionString="annotations" displayNumberOfHiddenItems>
+      {Array.from(annotations)}
```
dlrice, line 30: As far as I can see the order of the annotations is random. Sorting could slightly improve the ability to compare rows.

dlrice (approved):

src/automatic-annotations/shared/column-renderers/AnnotationCovered.tsx
aurel-l: Not sure what would be more useful, right now it's the order of first occurence in the rule (maybe from more to less important for the curator creating the rule?); could order alphabetically, to have consistent ordering; or could order by number of occurence of said annotation?

dlrice: The current site takes the order from the rule so that would be the safe choice

aurel-l: OK, so will keep that then

PR #445, opened 2021-09-15
2021 12 help center page

dlrice:
## Purpose
Create help center landing page [[jira]](https://www.ebi.ac.uk/panda/jira/browse/TRM-26242)

## Approach

- Created `IllustratedListTile` for each tile
- Used grid to layout the `IllustratedListTile`s
- Exported `Contact` from footer (and transformed to using CSS modules)
- Added SVGs into uniprot, if any needs to go into franklin instead I can do this. Some aren't monochromatic so would need to be adapted.
- Added circle reveal effect for the quick search component as per @aurel-l's suggestion

## Testing
- Snapshot

## Checklist


src/app/config/urls.ts
```diff
@@ -184,3 +184,8 @@ export const jobTypeToPath = (type: JobTypes, result?: boolean) => {
   }
   throw new Error(`"${type}"invalid job type`);
 };
+
+export const getHelpEntryPath = (entry: string) =>
```
xwatkins, line 188: Wondering if it would be possible to overload the `getEntryPath` function above instead of defining a new one, either by specifying the type of namespace as `SearchableNamespace|Location` and then checking the type in the function (maybe risky), or by creating another `getEntryPathFromLocation` more generic method.

src/help/components/landing/IllustratedListTile.tsx
```diff
@@ -0,0 +1,57 @@
...
+import cn from 'classnames';
+import { RequireExactlyOne } from 'type-fest';
+import { LocationDescriptor } from 'history';
+
+import styles from './styles/illustrated-list-tile.module.scss';
+
+type LinkOrHref = RequireExactlyOne<
+  {
+    label: string;
+    to: LocationDescriptor<unknown>;
```
xwatkins, line 13:
```suggestion
    to: LocationDescriptor;
```
Don't think the `unknown` generics adds much

src/help/components/landing/IllustratedListTile.tsx
```diff
@@ -0,0 +1,57 @@
...
+    url: string;
+  },
+  'to' | 'url'
+>;
+
+export type IllustratedListTileProps = {
+  title: string;
+  image: ReactElement;
+  linkList: LinkOrHref[];
+  moreTo: LocationDescriptor<unknown>;
```
xwatkins, line 23:
```suggestion
  moreTo: LocationDescriptor;
```

src/shared/components/layouts/UniProtFooter.tsx
```diff
@@ -302,14 +302,14 @@ const FooterShortcuts = () => (
   </ul>
 );
 
-const FooterContactAndElixir = () => (
-  <div className="contact-elixir">
+export const Contact = () => (
```
xwatkins, line 305: Wondering whether this might be better as its own component rather than in footer? Might need to put it in other places later.

dlrice: Done

src/app/config/urls.ts
dlrice: Good idea, went with your last suggestion

dlrice:
Addressed all comments

xwatkins (approved):

PR #450, opened 2021-09-22
Subcell viz membranes

dlrice:
## Purpose
Handle locations which are membranes or "part". [[Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-25569)]

## Approach

- From seen locations extract classnames that start with `mp_` or `part_` and convert these into IDs.
- Apply styles with selector `[class*="mp_"] .coloured, svg [class*="part_"] .coloured`

## Testing
None

## View changes
Accession A1L3X0 is a good example.

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/SubCellViz.tsx
+++ b/src/uniprotkb/components/protein-data-views/SubCellViz.tsx
@@ -12,6 +12,8 @@ import './styles/sub-cell-viz.scss';
 /*
   The logic implemented here to get our data into @swissprot/swissbiopics-visualizer has been lifted
   from the source code at http://sp.sib.swiss/scripts/uniprot_entry-bundle.js with some modifications
+
+  Good membrane example: A1L3X0
 */
 
 const shapes = [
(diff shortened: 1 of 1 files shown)


src/uniprotkb/components/protein-data-views/SubCellViz.tsx
```diff
@@ -62,6 +66,17 @@ const getGoTermClassNames = (locationGroup: Element) =>
...
+const getUniProtTextSelectors = (subcellularPresentSVG: Element) =>
+  [
+    `#${subcellularPresentSVG.id}term`,
+    ...Array.from(subcellularPresentSVG.classList)
+      .map((className: string) => {
+        const id = className.match(reMpPart)?.groups?.id;
+        return id && `#${id}term`;
+      })
+      .filter(Boolean),
+  ] as string[];
```
dlrice, line 78: Needed `as string[]` otherwise it thinks there could be come `undefined` elements which I don't think is possible with `filter(Boolean)`. Open to improvements.

aurel-l: I would suggest going for a function returning a type predicate instead of just the Boolean constructor
```ts
const getUniProtTextSelectors = (subcellularPresentSVG: Element): string[] => [
  `#${subcellularPresentSVG.id}term`,
  ...Array.from(subcellularPresentSVG.classList)
    .map((className: string) => {
      const id = className.match(reMpPart)?.groups?.id;
      return id && `#${id}term`;
    })
    .filter((sel: string | undefined): sel is string => Boolean(sel)),
];
See https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates for details
```

aurel-l (approved):

PR #453, opened 2021-09-23
2021 13 schema updates uniprotkb

xwatkins:
## Purpose
- Check model changes for UniProtKB
- Update new url path without `/accession`
- Handle `locator` field in Citations

## Approach
- There were no changes to the model
- All paths were updated and test cases updated. Checked in browser.
- Managed to dig up the old code from uuw to see how the urls corresponding to `locator` were handled and translated the logic. In the process, realised that ARVO was missing, Leo is looking into that and will create a subsequent Jira.

## Testing
Added a test to check the generation of the url was correct.
Examples to try: O64948, P98060

## Checklist

### DIFF
--- a/src/shared/config/apiUrls.ts
+++ b/src/shared/config/apiUrls.ts
@@ -108,7 +108,7 @@ const apiUrls = {
     joinUrl(apiPrefix, '/genecentric/', accession),
   idMappingFields: joinUrl(apiPrefix, '/configure/idmapping/fields'),
   entry: (id: string | undefined, namespace: Namespace) =>
-    id && joinUrl(apiPrefix, `/${namespace}/`, id),
+    id && joinUrl(apiPrefix, namespace, id),
   sequenceFasta: (accession: string) =>
     `${apiUrls.entry(accession, Namespace.uniprotkb)}.fasta`,
   entryDownload: (
--- a/src/shared/config/apiUrls.ts
+++ b/src/shared/config/apiUrls.ts
@@ -126,7 +126,7 @@ const apiUrls = {
           fileFormatToUrlParameter[format]
         }`,
   entryPublications: (accession: string) =>
-    joinUrl(apiPrefix, '/uniprotkb/', accession, '/publications'),
+    joinUrl(apiPrefix, 'uniprotkb', accession, '/publications'),
   taxonomySuggester: 'suggester?dict=taxonomy&query=?',
   organismSuggester: 'suggester?dict=organism&query=?',
 
--- a/src/shared/config/apiUrls.ts
+++ b/src/shared/config/apiUrls.ts
@@ -108,15 +108,7 @@ const apiUrls = {
     joinUrl(apiPrefix, '/genecentric/', accession),
   idMappingFields: joinUrl(apiPrefix, '/configure/idmapping/fields'),
   entry: (id: string | undefined, namespace: Namespace) =>
-    id &&
-    joinUrl(
-      apiPrefix,
-      // NOTE: The inclusion of /accession/ subpath for uniprotkb is going to be reviewed by backend
-      // and potentially removed to bring it in line with the other namespaces
-      // NOTE: uniparc entry isn't working/deployed yet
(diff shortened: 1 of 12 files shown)


src/shared/config/apiUrls.ts
```diff
@@ -108,15 +108,7 @@ const apiUrls = {
...
-    id &&
-    joinUrl(
-      apiPrefix,
-      // NOTE: The inclusion of /accession/ subpath for uniprotkb is going to be reviewed by backend
-      // and potentially removed to bring it in line with the other namespaces
-      // NOTE: uniparc entry isn't working/deployed yet
-      `/${namespace}/${namespace === Namespace.uniprotkb ? 'accession/' : ''}`,
-      id
-    ),
+    id && joinUrl(apiPrefix, `/${namespace}/`, id),
```
dlrice, line 111:
```suggestion
    id && joinUrl(apiPrefix, namespace, id),
```

src/shared/config/apiUrls.ts
```diff
@@ -134,7 +126,7 @@ const apiUrls = {
           fileFormatToUrlParameter[format]
         }`,
   entryPublications: (accession: string) =>
-    joinUrl(apiPrefix, '/uniprotkb/accession', accession, '/publications'),
+    joinUrl(apiPrefix, '/uniprotkb/', accession, '/publications'),
```
dlrice, line 129:
```suggestion
    joinUrl(apiPrefix, 'uniprotkb', accession, '/publications'),
```

src/supporting-data/citations/components/LiteratureCitation.tsx
```diff
@@ -37,6 +37,36 @@ const getLinkToAuthor = (author: string) => ({
   search: `query=lit_author:"${author}"`,
 });
 
+const WORM_BREEDERS_GAZETTE_URL = 'http://www.wormbook.org/wli/';
+// const ARVO_URL = 'http://iovs.arvojournals.org/article.aspx?articleid=';
+const PGR_URL = 'https://www.ebi.ac.uk/~textman/pgr-htdocs/pgr/';
```
dlrice, line 42: Could these go in `src/shared/config/externalUrls.ts`?

src/supporting-data/citations/components/LiteratureCitation.tsx
```diff
@@ -37,6 +37,36 @@ const getLinkToAuthor = (author: string) => ({
...
+export const getLocatorUrl = (
+  locator?: string,
+  name?: 'Plant Gene Register' | "Worm Breeder's Gazette" | string
+) => {
+  if (!locator || !name) {
+    return null;
+  }
+  switch (name) {
+    case 'Plant Gene Register':
+      return `${PGR_URL}${locator}`;
```
dlrice, line 53:
```suggestion
      return `${PGR_URL}${locator}.html`;
```

src/supporting-data/citations/components/__tests__/LiteratureCitation.spec.tsx
```diff
@@ -52,3 +52,22 @@ describe('Publication component', () => {
     expect(screen.getByText(/Submitted to/)).toBeInTheDocument();
   });
 });
+
+describe('Publication component url', () => {
+  test('Plant Gene Register', () => {
+    const url = getLocatorUrl('PGR98-023', 'Plant Gene Register');
+    expect(url).toEqual(
+      'https://www.ebi.ac.uk/~textman/pgr-htdocs/pgr/PGR98-023'
```
dlrice, line 60:
```suggestion
      'https://www.ebi.ac.uk/~textman/pgr-htdocs/pgr/PGR98-023.html'
```

src/supporting-data/citations/components/LiteratureCitation.tsx
xwatkins: In this case I'd rather have them localised with the component - reason for this is these are legacy, only concern a handful of publications and might be removed in the future.

dlrice (approved):

PR #455, opened 2021-09-24
2021 13 schema updates help

dlrice:
## Purpose
Update help for model changes. [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26660)]

## Approach
* Update mock data which has not had any type changes

## Testing
* Use mock data in more tests

## Checklist


src/help/components/entry/__tests__/Entry.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+describe('Help entry tests', () => {
+  it('should render the Entry component', async () => {
+    const props = {
+      match: {
+        params: {
+          accession: 'canonical_and_isoforms',
+        },
+      },
+    } as RouteChildrenProps<{ accession: string }>;
+    const { asFragment } = customRender(<Entry {...props} />);
```
dlrice, line 25: Could not figure out how to get `customRender` to pass in the `match.params.accession` other than the method seen here. Open to improvements.

xwatkins: You could use `useRouteMatch` to get the accession in `Entry.tsx` instead of passing the match as a prop. Then use the second argument of `customRender` to pass the `route`.

aurel-l: Not sure if changing the implementation in order to make a test nicer is the best thing.
That being said, I skipped using the useRouteMatch when implementing this component because I thought it was a pity to not just reuse what was passed as props. But if we want to explicitely require from within the component what is needed to render it instead of relying on props passed from above then it's fine to change it.

xwatkins: I was suggesting the `useRoutMatch` approach as we're using it in other places such as UniProtKB entry, etc...

aurel-l: Yes, but there too we are unnecessarily (but explicitely, which is also good), asking for something that is already passed to us. So maybe there might be away to remove that bit of logic from the all entry components in order to possibly abstract them away more easily?

PR #470, opened 2021-10-19
redirect obsolete entries without obvious redirect to history tab

aurel-l:
## Purpose
Redirect obsolete entries to history subpage https://www.ebi.ac.uk/panda/jira/browse/TRM-26729

## Approach
Instead of displaying an entry subpage saying that an entry is obsolete and having to switch to the history subpage to see the rest of information related to it being deprecated, jump to the history subpage directly.
As a consequence we don't use the big message about the entry being deprecated and I removed the corresponding code.

## Testing
Removed obsolete tests

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/Entry.tsx
+++ b/src/uniprotkb/components/entry/Entry.tsx
@@ -245,7 +245,7 @@ const Entry: FC = () => {
     // (I hope) I know what I'm doing here, I want to stick with whatever value
     // match?.params.subPage had when the component was mounted.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [dispatch, isObsolete]);
+  }, [isObsolete]);
 
   if (
     loading ||


src/uniprotkb/components/entry/Entry.tsx
```diff
@@ -220,6 +220,33 @@ const Entry: FC = () => {
...
+            match?.params.accession,
+            TabLocation.History
+          )
+        );
+      });
+    }
+    // (I hope) I know what I'm doing here, I want to stick with whatever value
+    // match?.params.subPage had when the component was mounted.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dispatch, isObsolete]);
```
dlrice, line 248:
```suggestion
  }, [isObsolete]);
```
`dispatch` not used by the `useEffect` so my thinking is it can be excluded

dlrice (approved):

PR #472, opened 2021-10-20
2021 14 unipathway

dlrice:
## Purpose
Deal with UniPathway not having a [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26809)] by adjusting types.

## Approach
Adjust types to guard against lack of `uriLink`

## Testing
Added UniPathway back to mock data

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/XRefView.tsx
+++ b/src/uniprotkb/components/protein-data-views/XRefView.tsx
@@ -108,7 +108,7 @@ const EMBLXref: FC<{
   const { properties, additionalIds } = xref;
   if (!databaseInfo?.uriLink || !genBankInfo?.uriLink || !ddbjInfo?.uriLink) {
     logging.error(
-      'EMBL, GenBank or DDBJ database information not found in drlineconiguration'
+      'EMBL, GenBank or DDBJ database information not found in database configuration file'
     );
     return null;
   }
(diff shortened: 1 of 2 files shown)


src/uniprotkb/components/protein-data-views/XRefView.tsx
```diff
@@ -106,10 +106,11 @@ const EMBLXref: FC<{
   const genBankInfo = databaseToDatabaseInfo.GenBank;
   const ddbjInfo = databaseToDatabaseInfo.DDBJ;
   const { properties, additionalIds } = xref;
-  if (!genBankInfo || !ddbjInfo) {
-    logging.warn(
-      'GenBank or DDBJ database information not found in drlineconiguration'
+  if (!databaseInfo?.uriLink || !genBankInfo?.uriLink || !ddbjInfo?.uriLink) {
+    logging.error(
+      'EMBL, GenBank or DDBJ database information not found in drlineconiguration'
```
aurel-l, line 111: Is it meant to say `drlineconfiguration` (with "f")?

aurel-l (approved):

dlrice:
> Now I'm thinking about it more, not a blocker though, but I'm wondering if you had considered changing `processUrlTemplate` to take `urlTemplate: string | null | undefined` in order to avoid all the ternary operators and extra checks that you added in this PR?

I was thinking that at some point too. I decided against it as `processUrlTemplate`  would have had to return null and handled by the consumer anyway. In the case of no url we don't want to render a link with no url, we want to render just the text.

src/uniprotkb/components/protein-data-views/XRefView.tsx
dlrice: Good catch! Also, drlineconfiguration has been renamed to allDatabases ~1year ago so will change this to the following to guard against future file name changes:
```
EMBL, GenBank or DDBJ database information not found in database configuration file
```

PR #473, opened 2021-10-21
2021 14 feature columns

xwatkins:
## Purpose
Some of the definition for feature columns were missing.

## Approach
- Refactored the columns and feature definition for categories so it's only in one place.
- Simplified code in column definition
- Use normal table instead of Nightingale component in the results section. Most of the functionality of the component is not needed here, and we are looking at having a compact display, which the component doesn't support (if only 1 item, there is a huge space below because of the overflow).
- Removed evidence tags from table (this was decided in the past)
- Use compact table modifier
- Changed the table header colour to differentiate from main table header, and update z-index.

## Testing
Updated snapshots.

## Checklist


dlrice (commented):
1. `UniProtKBColumn.ftVariant` is still using a protvista table
2. `ft_topo_dom` is missing a column renderer

PR #479, opened 2021-11-04
2021 15 subcell loc viz review status

dlrice:
## Purpose
Reflect manual/automatic annotation in sub-cellular visualization [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-26907)].

## Approach
- Get review status from associated ECO codes for each location
- Incorporate into style sheet
- Slight difference in uniprot and beta.uniprot in that I am reflecting reviewed status after unreviewed to so that reviewed takes precedence (in the case when a part of a location and a location contradict in review status)
- Updated `getEvidenceCodeData` to be an object to provide `O(1)` lookup time rather than the switch's `O(n)`
- Don't highlight non-active locations.
- Updated highlight colors with Marija

## Viewing changes
- Any accession will do. On my local setup I've had to hard-refresh the page in the browser to see the changes reflected in the subcellviz. No idea why this is happening.

## Testing
- Update unit tests and snapshots

## Checklist


aurel-l (approved):
Pick it up otherwise it would have stuck around forever and we need that closed for this sprint.

Looks good to me.
Had to merge main in to test with the new API at rest.uniprot, and modified some filter functions to use type guards and avoid casting (see latest 2 commits).

When comparing with the current website the colours are now the same, but I do find it strange that automatic annotations could yield a yellow colour like in P29431 for example. Do you mind double checking that this is correct? And maybe @xwatkins have a follow up as I probably won't be able to follow up before the end of the sprint.

dlrice:
> but I do find it strange that automatic annotations could yield a yellow colour like in P29431 for example

Looking at this the following ECO data is used:

```
  [ecoCode.ISM]: { //255
    manual: true,
    label: 'Automatic assertion according to rules',
    description: 'Inferred from sequence model',
    labelRender: rulesCountRenderer,
  },
```
and from https://www.uniprot.org/help/evidences:

> The ECO code ECO:0000255 is also used for information which has been generated by various sequence analysis programs that are used during the manual curation process and which has been verified by a curator.

I agree that the rule label isn't very clear. Maybe it should be as is indicated in the help file eg:
```
  label: 'Automatic assertion according to rules and manually verified',
```

xwatkins:
If you're thinking of changing the text, it needs to be ran past Elisabeth at SIB.

aurel-l:
OK, so good to go for that PR, and we can have a conversation about that text in the evidence at a later point

PR #492, opened 2021-11-29
2021 16 use db endpoint

dlrice:
## Purpose
Use the DB endpoint: https://rest.uniprot.org/beta/configure/uniprotkb/allDatabases

## Approach

## Testing

## Checklist


src/app/components/App.tsx
```diff
@@ -286,174 +288,176 @@ const App = () => {
...
-                  </SingleColumnLayout>
-                )}
-              />
-            </Switch>
-          </Suspense>
-        </BaseLayout>
-        <ErrorBoundary fallback={null}>
-          <GDPR />
-        </ErrorBoundary>
+        <DBMapsProvider>
```
aurel-l, line 291: This file is already quite busy with all the routes, and we might end up adding some other top-level contexts eventually.
I would create a file exporting a GlobalContexts component, containing the ReduxProvider in index.tsx, and FranklinSite, Router, and DBMapsProvider. and make the index.tsx look like
```
ReactDOM.render(
  <GlobalContexts store={store} history={history}>
    <App />
  </GlobalContexts>,
  document.getElementById('root')
);
```
So that any further global context that we might add in the future, we would put with the rest of them.

aurel-l: (Sorry, the GitHub preview for that comment doesn't really look like what I was expecting, it's meant to cover these lines:
```
<FranklinSite>
  <Router history={history}>
    <DBMapsProvider>
```
)

dlrice: That's a nice idea - I'll do that now

src/shared/contexts/DatabaseInfoMaps.tsx
```diff
@@ -0,0 +1,69 @@
...
+  // Sanity check for dynamic database info and static column definition
+  useEffect(() => {
+    if (data) {
+      const definedColumns = new Set<string>();
+      Object.values(UniProtKBColumn).forEach((column) => {
+        if (column.startsWith('xref_')) {
+          definedColumns.add(column.replace('xref_', ''));
+        }
+      });
+      data.forEach((dbInfoMapping) => {
```
dlrice, line 43:
```suggestion
      data.forEach((database) => {
```

src/shared/contexts/DatabaseInfoMaps.tsx
```diff
@@ -0,0 +1,69 @@
...
+    () => (data ? getDatabaseInfoMaps(data) : data),
+    [data]
+  );
+
+  // Sanity check for dynamic database info and static column definition
+  useEffect(() => {
+    if (data) {
+      const definedColumns = new Set<string>();
+      Object.values(UniProtKBColumn).forEach((column) => {
+        if (column.startsWith('xref_')) {
```
dlrice, line 39: Could reuse existing `reXrefPrefix`

src/shared/contexts/DatabaseInfoMaps.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const databaseInfoMaps = useMemo(
+    () => (data ? getDatabaseInfoMaps(data) : data),
+    [data]
+  );
+
+  // Sanity check for dynamic database info and static column definition
+  useEffect(() => {
+    if (data) {
+      const definedColumns = new Set<string>();
+      Object.values(UniProtKBColumn).forEach((column) => {
```
dlrice, line 38: What do you think about extracting the sanity check into its own function and then also testing this?

xwatkins: Yep, will get on it.

PR #494, opened 2021-12-01
Add evidence, don't collapse inline

xwatkins:
## Purpose
Address issues due to the rendering of keywords and their evidences

## Approach
- Added the evidence tag component
- Changed logic to not use ExpandableList when in inline mode
- Remove extraneous dot

## Testing
Update test snapshots

## Checklist


src/uniprotkb/components/results/UniProtKBCard.tsx
```diff
@@ -34,11 +34,12 @@ const UniProtKBCard = ({ data }: { data: UniProtkbAPIModel }) => {
       'Disease',
     ]);
 
-    return categorisedKeywords.map((keywordCategory, index) => (
-      <Fragment key={keywordCategory.category}>
-        {index > 0 && ' · '}
-        <KeywordList keywords={keywordCategory.keywords} inline />
-      </Fragment>
+    return categorisedKeywords.map((keywordCategory) => (
```
dlrice, line 37: Screenshot 2021-12-01 at 14 28 42

Noticed some slight spacing differences between each of the `KeywordList` components - maybe would be easiest to flatten all of the `categorisedKeywords` and use a single `KeywordList`

xwatkins: Good catch - in the end removed the `useMemo` altogether as we're not giving any indication of category, just throwing keywords on the screen.

dlrice: Without using `getKeywordsForCategories` we aren't filtering to only include:
```
      'Molecular function',
      'Biological process',
      'Disease',
```
but will have all:
```
export type KeywordCategory =
  | 'Biological process'
  | 'Cellular component'
  | 'Coding sequence diversity'
  | 'Developmental stage'
  | 'Disease'
  | 'Domain'
  | 'Ligand'
  | 'Molecular function'
  | 'PTM'
  | 'Technical term';
```
Seems we intentionally filtered this in the past - do we still want to do this?

dlrice:
Screenshot 2021-12-01 at 15 06 24

src/uniprotkb/components/results/UniProtKBCard.tsx
xwatkins: Right, good point I thought it was like aspects - I'll revert and flatten.

aurel-l: You could also just filter yourself, something like "keywords.filter(kw => selectedCategories.has(kw.category))" with selectedCategories the set of 3 categories, defined outside of the component, and you should be good to go.

aurel-l:
> Screenshot 2021-12-01 at 15 06 24

See `.keyword-view` class, space needs to be after except on the last element to avoid .1 raised by @dlrice. I think .2 is OK, cards could have variable heights depending on content, and it would avoid having more things scrolling horizontally

dlrice:
> I think .2 is OK, cards could have variable heights depending on content, and it would avoid having more things scrolling horizontally

Was thinking more about keeping a single keyword together on one line eg:

one keyword on one line:
```
#keyword1 #keyword2
#keyword3 #Protein transport
```
one keyword split over two lines:
```
#keyword1 #keyword2
#keyword3  #Protein
transport
```

xwatkins:
Switched to flexbox as it tackles both issues.

src/uniprotkb/components/results/UniProtKBCard.tsx
```diff
@@ -23,23 +23,18 @@ const UniProtKBCard = ({ data }: { data: UniProtkbAPIModel }) => {
...
-
-    return categorisedKeywords.map((keywordCategory, index) => (
-      <Fragment key={keywordCategory.category}>
-        {index > 0 && ' · '}
-        <KeywordList keywords={keywordCategory.keywords} inline />
-      </Fragment>
-    ));
+    ])
+      .map(({ keywords }) => keywords)
+      .flat();
```
dlrice, line 37:
```suggestion
      .flatMap(({ keywords }) => keywords);
```

dlrice (approved):

PR #495, opened 2021-12-01
keywords

aurel-l:
avoid categorising then flattening, and take into account possible empty array

### DIFF
--- a/src/shared/components/results/__tests__/__snapshots__/ResultsData.spec.tsx.snap
+++ b/src/shared/components/results/__tests__/__snapshots__/ResultsData.spec.tsx.snap
@@ -110,9 +110,9 @@ exports[`ResultsData component should render the card view with the correct numb
                   class="keyword-view--inline"
                 >
                   <a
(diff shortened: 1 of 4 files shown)


xwatkins:
While it looks neater and is probably slightly more performant (and with the size of list we're dealing with, it's likely irrelevant), I would be inclined to keep the current approach:
`getKeywordsForCategories` is used in multiple places, so a change in the logic required to do this grouping/filtering only needs to be done in one place.

aurel-l:
Note that, additionally to that, it also returns a list of keywords in an expected alphabetical order (otherwise they're grouped by category, then alphabetically, but the user doesn't know that and it looks like they're mixed), and it covers the case where you end up with an empty list of keywords (see the 2 empty `small > div.keyword-view--inline` in your snapshots)

xwatkins:
https://www.ebi.ac.uk/panda/jira/browse/TRM-22668

Fixing the empty list issue on the branch

aurel-l:
Changes made in #494

PR #496, opened 2021-12-02
2021 16 isoforms

aurel-l:
## Purpose
Add isoform info and links where needed https://www.ebi.ac.uk/panda/jira/browse/TRM-27048 & https://www.ebi.ac.uk/panda/jira/browse/TRM-26821

## Approach
Add isoform text wherever we have the `molecule` information associated with a comment
Make it a link when in the entry page, plain text when in the search results
Make the isoform section in the entry page targetable through a hash navigation

## Testing
Updated current snapshots

## Checklist


src/uniprotkb/components/entry/FunctionSection.tsx
```diff
@@ -139,6 +139,13 @@ export const CofactorView = ({ cofactors, title }: CofactorViewProps) => {
       {cofactors.map((cofactorComment, index) => (
         // eslint-disable-next-line react/no-array-index-key
         <section className="text-block" key={index}>
+          {cofactorComment.molecule && (
+            <h4 className="tiny">
+              <a href={`#${cofactorComment.molecule.replace(' ', '_')}`}>
+                {cofactorComment.molecule}:
```
dlrice, line 145: Do you think the `:` is needed? When rendered it already looks like a title to me so it's clear that follows pertains to the molecule. I know the current site does it but I'm wondering if it's necessary?

src/shared/components/entry/__tests__/__snapshots__/SequenceView.spec.tsx.snap
```diff
@@ -393,9 +392,8 @@ exports[`SequenceView component should render SequenceInfo with provided sequenc
       </div>
     </div>
   </section>
-  <hr />
   <h3
-    id="P05067-2"
+    id="Isoform_APP770-2"
```
dlrice, line 396: Was there a particular reason to move away from the current site's isoform anchor tag format?:

```
https://www.uniprot.org/uniprot/P07954#P07954-1
```

src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
```diff
@@ -3027,6 +3054,15 @@ exports[`Entry view should render for non-human entry 1`] = `
       <div
         class="card__content"
       >
+        <h3
```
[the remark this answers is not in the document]
aurel-l: You really have eagle eyes 🦅 
I did notice it, but figured it wouldn't be important, because it's just noticeable when comparing before and after.
This is because we're generating headings at the component level and I couldn't override this one, so I just visually hid its content but it's margin is still there.

src/uniprotkb/components/entry/FunctionSection.tsx
aurel-l: Not sure, I matched the current website on this, but could remove if you think the new style is good enough and makes it redundant. Don't have any opinion on this

src/shared/components/entry/__tests__/__snapshots__/SequenceView.spec.tsx.snap
aurel-l: Because I don't have this information wherever we might link to this section. The `molecule` field is showing the name of the isoform, and not its accession.

src/uniprotkb/components/entry/FunctionSection.tsx
dlrice: I vote to remove the `:` as my brain is expecting something appear horizontally beside it eg

```
Isoform Mitochondrial: something about isoform mitochondrial
```

dlrice (approved):

PR #500, opened 2021-12-06
Isoform section

aurel-l:
## Purpose
Modifications to the Isoforms sections https://www.ebi.ac.uk/panda/jira/browse/TRM-26879 & https://www.ebi.ac.uk/panda/jira/browse/TRM-26631

## Approach
- When on isoform page, display only the isoform sequence, opened
  - Added a quick link back to the canonical
- Added link to UniParc and UniRef for each isoform

## Testing
Updated snapshots & checked on canonical page and corresponding isoform pages

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
       "global": {
         "lines": 82.16,
         "statements": 81.94,
-        "functions": 82.41,
+        "functions": 82.38,
         "branches": 69.6
       }
     },
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
       "global": {
         "lines": 82.16,
         "statements": 81.94,
-        "functions": 82.38,
+        "functions": 82.37,
         "branches": 69.6
       }
     },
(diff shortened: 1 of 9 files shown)


dlrice (commented):
1. Could we combine the two sentences together into one line to keep the white space
Screenshot 2021-12-07 at 09 56 54

 
2. Not part of this but I noticed that if there isn't any info-list data it still takes up a lot of space - what do you think about checking for this before rendering?
Screenshot 2021-12-07 at 09 57 34

aurel-l:
> 1. Could we combine the two sentences together into one line to keep the white space to a minimum
> 
> 
> Screenshot 2021-12-07 at 09 56 54

Addressed in my most recent commit
> 
>     2. Not part of this but I noticed that if there isn't any info-list data it still takes up a lot of space - what do you think about checking for this before rendering?
> 
> 
> Screenshot 2021-12-07 at 09 57 34

I believe this should probably be handle by the InfoList in Franklin. In this specific case this contains 2 elements whose content is empty. I would check that on Franklin's side and render null if that is the case.

dlrice (approved):

PR #501, opened 2021-12-06
2021 16 subsequence links

xwatkins:
## Purpose

## Approach

## Testing

## Checklist

### DIFF
--- a/src/app/config/urls.ts
+++ b/src/app/config/urls.ts
@@ -195,18 +195,6 @@ export const getLocationEntryPath = (location: Location, accession: string) =>
 export const getLocationEntryPathFor = (location: Location) =>
   partial(getLocationEntryPath, location);
 
-export const getURLToJobWithData = (
-  jobType: JobTypes,
-  primaryAccession: string,
-  options?: {
-    start: number;
-    end: number;
-  }
-) =>
-  `${jobTypeToPath(jobType)}?ids=${primaryAccession}${
-    options ? `[${options.start}-${options.end}]` : ''
-  }`;
-
 // eslint-disable-next-line consistent-return
 export const jobTypeToPath = (type: JobTypes, result?: boolean) => {
   switch (type) {
@@ -227,3 +215,15 @@ export const jobTypeToPath = (type: JobTypes, result?: boolean) => {
   }
   throw new Error(`"${type}"invalid job type`);
 };
+
+export const getURLToJobWithData = (
+  jobType: JobTypes,
+  primaryAccession: string,
+  options?: {
+    start: number;
+    end: number;
+  }
+) =>
+  `${jobTypeToPath(jobType)}?ids=${primaryAccession}${
+    options ? `[${options.start}-${options.end}]` : ''
+  }`;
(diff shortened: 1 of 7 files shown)


src/shared/components/action-buttons/ToolsButton.tsx
```diff
@@ -32,60 +19,25 @@ const ToolsButton: FC<ToolsButtonProps> = ({
...
-  useEffect(() => {
-    // reset loading state every time the selection changes
-    setLoading(false);
-  }, [selectedEntries, setLoading]);
+    history.push(LocationToPath[location], {
+      parameters: {
+        ids: entries,
+      },
+    });
+  }, [history, location]);
```
aurel-l, line 34: No need to do that in a click handler now, you can just use a Link instead of a Button and pass what you're passing to history.push to the `to` prop

xwatkins: A bit confused by the `ref`, was that to handle updates while the fetch hadn't finished?

xwatkins: I think we still need this to be a Button because it can be disabled - also there might be styling divergences if we don't use the `Button` component.

xwatkins: Had forgotten `Buttons` could be "anything" 😆

xwatkins: In the end stayed with a regular `button` - there are styling side issues I didn't want to get into as part of this PR.

aurel-l: Can you detail the styling issues you encountered?

xwatkins: results page: vertical alignment , mouse over when "disabled"...
All are probably all addressable locally, but I imagine just the tip of the iceberg, so would need a more generic solution.

aurel-l: Note that these issues are probably also present in all place where we combine both

xwatkins: Yep ... except I don't think we currently use anchors within `.button-group`. This is why I'm saying this needs a generic solution (to be addressed globally).

aurel-l: Entry feedback and add a publication, on the top of my mind

xwatkins: image
There's an issue there too...

src/app/config/urls.ts
```diff
@@ -195,6 +195,17 @@ export const getLocationEntryPath = (location: Location, accession: string) =>
...
+  primaryAccession: string,
+  options?: {
+    start: number;
+    end: number;
+  }
+) =>
+  `${LocationToPath[Location.Blast]}?ids=${primaryAccession}${
+    options ? `[${options.start}-${options.end}]` : ''
+  }`;
+
```
aurel-l, line 208:
```suggestion
export const getURLToJobWithData = (
  jobType: JobTypes,
  primaryAccession: string,
  options?: {
    start: number;
    end: number;
  }
) =>
  `${jobTypeToPath(jobType)}?ids=${primaryAccession}${
    options ? `[${options.start}-${options.end}]` : ''
  }`;

```

Knowing that this is a common file I wouldn't make it specific. If it becomes too specific it should probably live somewhere in src/tools/blast and depend on `jobTypeToPath` as this is already a function to generate paths to a job form

aurel-l (approved):

PR #502, opened 2021-12-07
2021 16 subsequence links query params

dlrice:
## Purpose
Provide the ability to BLAST and align subsequences

## Approach
- Create `parseIdsFromSearchParams` to parse search params of the form `P05067[99-114],P05068`
- Use location to get search params in `useInitialFormParameters.ts`
- Use `useDataApi` to request sequences from the accessions endpoint and parse with `parseIdsFromSearchParams`
- In BlastForm (so far, as a proof of concept) show loading while accessions are downloading
- In BlastForm, watch for changes to initialFormValues.sequence and update

## View
http://localhost:8080/blast?ids=P05067[99-114],P05068

## Testing
Unit tests

## Checklist


src/tools/hooks/useInitialFormParameters.ts
```diff
@@ -18,41 +25,119 @@ type FormValue = {
...
-    return defaultFormValues;
-  }, [defaultFormValues, history.location?.state]);
 
-  return initialValues;
+    // ids parameter from the url has been passed so handle the fetched accessions once loaded
+    if (!!accessionsData?.results?.length && idsMaybeWithRange) {
+      const idToSequence = keyBy(
+        accessionsData.results,
+        ({ primaryAccession }) => primaryAccession
+      );
```
aurel-l, line 99: Would that be the same as
```
Object.fromEntries(accessionsData.results.map((datum) => [datum.primaryAccession, datum]))
```
?

aurel-l: what about:
```
const idToSequence = new Map(accessionsData.results.map((datum) => [datum.primaryAccession, datum]))
```

then lower we would do
```
idToSequence.get(id)
```

aurel-l (approved):

PR #503, opened 2021-12-09
2021 16 subsequence links other forms

xwatkins:
## Purpose
Update the code in all forms to handle the new default parameters loading.

## Approach
Wrap state management in `useEffect`. This has implications in the defaults for these states - some will be `null`, some `undefined` but it might be worth having a standard approach and default all to `null`. See what you think.

## Testing
Updated tests.

## Checklist

### DIFF
--- a/src/shared/components/action-buttons/ToolsButton.tsx
+++ b/src/shared/components/action-buttons/ToolsButton.tsx
@@ -23,7 +23,7 @@ const ToolsButton: FC<ToolsButtonProps> = ({
   const handleClick = useCallback(async () => {
     history.push({
       pathname: LocationToPath[location],
-      search: selectedEntries.join(),
+      search: `ids=${selectedEntries.join()}`,
     });
   }, [history, location, selectedEntries]);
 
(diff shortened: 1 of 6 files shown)


src/shared/components/action-buttons/ToolsButton.tsx
```diff
@@ -23,7 +23,7 @@ const ToolsButton: FC<ToolsButtonProps> = ({
   const handleClick = useCallback(async () => {
     history.push({
       pathname: LocationToPath[location],
-      search: selectedEntries.join(),
+      search: `ids=${selectedEntries.join()}`,
```
dlrice, line 26:
```suggestion
      search: `ids=${selectedEntries.join(',')}`,
```

src/tools/align/components/AlignForm.tsx
```diff
@@ -54,6 +56,7 @@ import {
 
 import sticky from '../../../shared/styles/sticky.module.scss';
 import '../../styles/ToolsForm.scss';
+import { logger } from '@sentry/utils';
```
dlrice, line 59: I think this wants to be something like `import * as logging from './shared/utils/logging';`

src/shared/components/action-buttons/ToolsButton.tsx
xwatkins: Thanks Dan!

dlrice (approved):

PR #504, opened 2021-12-10
Parse publication cited for to include BLAST links

xwatkins:
## Purpose
Add link to BLAST with subsequence in Publication cited for text. See http://localhost:8080/uniprotkb/P25719/publications

## Approach
Check for matching regex and replace text

## Testing
Adding checks for Regex function

## Checklist

### DIFF
--- a/src/shared/utils/__tests__/__snapshots__/utils.spec.ts.snap
+++ b/src/shared/utils/__tests__/__snapshots__/utils.spec.ts.snap
@@ -1,19 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`deepFindAllByKey addBlastLinksToFreeText should generate parse the text and generate link 1`] = `
-Array [
-  Array [
-    "SOME TEXT ",
-    <Link
-      to="/blast?ids=UP_ACCESSION[10-34]"
-    >
-      10-34
-    </Link>,
-    "",
-  ],
-]
-`;
-
 exports[`deepFindAllByKey addBlastLinksToFreeText should handle no positions 1`] = `
 Array [
   Array [
--- a/src/shared/utils/__tests__/__snapshots__/utils.spec.ts.snap
+++ b/src/shared/utils/__tests__/__snapshots__/utils.spec.ts.snap
@@ -1,0 +1,51 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`deepFindAllByKey addBlastLinksToFreeText should generate parse the text and generate link 1`] = `
+Array [
+  Array [
+    "SOME TEXT ",
+    <Link
+      to="/blast?ids=UP_ACCESSION[10-34]"
+    >
+      10-34
+    </Link>,
+    "",
+  ],
+]
+`;
+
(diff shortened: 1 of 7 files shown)


src/shared/utils/utils.tsx
```diff
@@ -108,3 +110,24 @@ export function* deepFindAllByKey<T = string>(
...
+  texts: string[],
+  primaryAccession: string
+) =>
+  texts.map((text) => {
+    const splitText = text.split(/(\d+-\d+)/);
+    return splitText.map((splitItem, i) => {
+      if (i % 2 === 1) {
+        return (
+          <Link
+            to={getBLASTURL(primaryAccession, splitItem)}
```
aurel-l, line 124: I would do something like

```suggestion
            to={getURLToJobWithData(JobTypes.BLAST, primaryAccession, splitItem)}
```

Or at least modify the existing getURLToJobWithData to directly handle splitItem if you don't want to split it by start and end

src/shared/utils/utils.tsx
```diff
@@ -108,3 +111,28 @@ export function* deepFindAllByKey<T = string>(
...
+    return splitText.map((splitItem, i) => {
+      if (i % 2 === 1) {
+        const range = splitItem.split('-');
+        return (
+          <Link
+            to={getURLToJobWithData(JobTypes.BLAST, primaryAccession, {
+              start: +range[0],
+              end: +range[1],
+            })}
+            key={`${splitItem}`}
```
aurel-l, line 130: If you have decided to go for the content instead of the index, then maybe no need to stringify it?

src/uniprotkb/components/entry/EntryPublications.tsx
```diff
@@ -92,7 +93,17 @@ const PublicationReference: FC<{ reference: Reference; accession: string }> = ({
...
     {
       title: 'Cited for',
-      content: referencePositions?.join(', '),
+      content:
+        referencePositions &&
+        addBlastLinksToFreeText(referencePositions, accession).map(
+          (item, i) => (
+            // eslint-disable-next-line react/no-array-index-key
+            <Fragment key={i}>
+              {i > 0 && ', '}
```
aurel-l, line 102: You could use `listFormat` to handle this kind of conjunction in a proper sentence

xwatkins: I think a comma separated list works better there than separating the blocks by OR 
image

xwatkins: * AND

src/shared/utils/__tests__/__snapshots__/utils.spec.ts.snap
```diff
@@ -0,0 +1,37 @@
...
+  Array [
+    "SOME TEXT",
+  ],
+]
+`;
+
+exports[`deepFindAllByKey addBlastLinksToFreeText should handle positions at the beggining 1`] = `
+Array [
+  Array [
+    "",
```
aurel-l, line 14: Do we want to filter those out?

aurel-l (approved):

PR #506, opened 2021-12-13
Kanban ac accession

dlrice:
## Purpose
[Sequence caution - parse "AC " to create hyperlink](https://www.ebi.ac.uk/panda/jira/browse/TRM-26896)
## Approach
Split caution text by whitespace and iterate to find any instances of an AC preceding an accession and update with link if found.

## Testing
* Snapshots updated
* New unit tests

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/FreeTextView.tsx
+++ b/src/uniprotkb/components/protein-data-views/FreeTextView.tsx
@@ -4,7 +4,7 @@ import { Link } from 'react-router-dom';
 import UniProtKBEvidenceTag from './UniProtKBEvidenceTag';
 
 import { getEntryPath, getEntryPathFor } from '../../../app/config/urls';
-import { uniProtKBAccessionRegEx } from '../../utils';
+import { acRE, uniProtKBAccessionRE } from '../../utils';
 
 import { Namespace } from '../../../shared/types/namespaces';
 import { FreeTextComment, TextWithEvidence } from '../../types/commentTypes';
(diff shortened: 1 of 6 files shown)


src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -20,50 +21,78 @@ const needsNewLineRE = /^\).\s+/;
...
-        })}
-        {!noEvidence && comment.evidences && (
-          <UniProtKBEvidenceTag evidences={comment.evidences} />
-        )}
-      </Fragment>
-    ))}
-  </div>
-);
+export const TextView = ({ comments, noEvidence }: TextViewProps) => {
+  const acRE = new RegExp(`AC ${uniProtKBAccessionRegEx.source}`, 'gi');
```
dlrice, line 49: Needs to go here to access the instantiated `uniProtKBAccessionRegEx`

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -20,50 +21,78 @@ const needsNewLineRE = /^\).\s+/;
...
+  return (
+    <div className="text-block">
+      {comments.map((comment, index) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <Fragment key={index}>
+          {comment.value.split(pubMedRE).map((part, index, { length }) => {
```
xwatkins, line 55: Thinking it might be cleaner to use the exact same strategy used for PubMED ids: split the comment on acRE and looking at exactly matching groups?

dlrice: Will have a go at refactoring

dlrice: Done

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -20,50 +21,82 @@ const needsNewLineRE = /^\).\s+/;
 
 const getEntryPathForCitation = getEntryPathFor(Namespace.citations);
 
+export const ACCommentView = ({ string }: { string: string }) => {
+  // This replaces any occurrences of "AC <accession>" with "AC "<link to the accession>
+  const acRE = new RegExp(`(AC ${uniProtKBAccessionRegEx.source})`, 'i');
```
aurel-l, line 26:
```suggestion
const acRE = new RegExp(`(AC ${uniProtKBAccessionRegEx.source})`, 'i');

export const ACCommentView = ({ string }: { string: string }) => {
  // This replaces any occurrences of "AC <accession>" with "AC "<link to the accession>
```

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -20,50 +21,82 @@ const needsNewLineRE = /^\).\s+/;
...
+            {`AC `}
+            <Link to={getEntryPath(Namespace.uniprotkb, accession)}>
+              {accession}
+            </Link>
+          </Fragment>
+        ) : (
+          // eslint-disable-next-line react/no-array-index-key
+          <Fragment key={index}>
+            {index + 1 === length && !part.endsWith('.') ? `${part}.` : part}
+          </Fragment>
```
aurel-l, line 43: I believe you might not need a Fragment nor a key when you return a string. If that is the case you could remove that here.
That being said, you'll get eslint complaining about ternary operator inside a ternary operator.
I would do multiple returns in different if conditions so that it's clearer to see what is returned under which conditions.

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -20,50 +21,82 @@ const needsNewLineRE = /^\).\s+/;
...
-        })}
-        {!noEvidence && comment.evidences && (
-          <UniProtKBEvidenceTag evidences={comment.evidences} />
-        )}
-      </Fragment>
-    ))}
-  </div>
-);
+export const TextView = ({ comments, noEvidence }: TextViewProps) => {
+  const acRE = new RegExp(`AC ${uniProtKBAccessionRegEx.source}`, 'i');
```
aurel-l, line 53:
```suggestion
const acRE = new RegExp(`AC ${uniProtKBAccessionRegEx.source}`, 'i');

export const TextView = ({ comments, noEvidence }: TextViewProps) => {
```

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
xwatkins: I think this is what @dlrice was referring to here https://github.com/ebi-uniprot/uniprot-website/pull/506#discussion_r767761224
I tried it locally and had the same issue: I think webpack is processing the dependency in an order that means that `uniProtKBAccessionRegEx` isn't initialised at the time... played with it for a bit locally but couldn't find a solution either.

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
xwatkins: Thanks, that's nicer.
The other thing I was referring to (but maybe it's not an issue) is the fact the comments are split on `pubMedRE` (L59) and then evaluated for `aceRE` (L71). I was wondering if it would be better to do another split on the comment values (as another block) afterward? It could be more readable, but maybe less performant?

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
aurel-l: Ah, it's a circular dependency as we've seen before some times. Maybe these regular expressions could be defined directly in the utils module then?

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
dlrice: I see your point and that was also something in the back of my head. After the `pubMedRE` we'll have an array of react elements and strings which I anticipate will complicate it a bit. For each element of the array we could check the type and if it's a string go ahead and split on `acRE`. What do you think?

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
dlrice: Exported from utils module now

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
dlrice: Exported from utils module now

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
dlrice: Done

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
xwatkins: I'm not sure there is an easy way to check for `string|ReactNode`. If it gets too complicated don't bother.

aurel-l (approved):

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
dlrice: I think this is good to go now

xwatkins:
LGTM - wanted to check having both PubMed and AC in the same text so have added a test case.

xwatkins (approved):

PR #508, opened 2021-12-14
Add redirect if missing 'KW-'

xwatkins:
## Purpose
Other resources link to UniProt keywords by using an old format, without the `KW-` prefix.

## Approach
If the submitted accession is ony made of numbers, update the location by adding `KW-` to the number.

## Testing
N/A

## Checklist


aurel-l (approved):
Approved, but changed the implementation 😅 so feel free to re-review.
Will generalise this approach in other places as I've realised some errors in Sentry were caused in the same way (first render with content, but we will directly redirect so there shouldn't be an error anyway). Here it wasn't causing an error, just starting an extra network call, but I guess it's always better to avoid an extra request even if cancelled rightaway.

xwatkins (commented):
Fair enough - I was still in the mindset of what we'd implemented for BLAST/Align (drop query in url)

PR #509, opened 2021-12-14
Add keyword id to display

xwatkins:
## Purpose
Keywords results table was displaying the keyword name instead of the id

## Approach
Update the column renderer to display the id instead. At the same time have added back the keyword id in the title of results cards and heading for the keyword entry page.

## Testing
Update snapshots

## Checklist


src/supporting-data/keywords/components/results/KeywordsCard.tsx
```diff
@@ -33,7 +33,9 @@ const KeywordsCard = ({ data }: { data: KeywordsAPIModel }) => {
       header={
         <>
           <CardCheckboxCell id={id} />
-          <h2 className="tiny">{data.keyword.name}</h2>
+          <h2 className="tiny">
+            {data.keyword.name} ({data.keyword.id})
+          </h2>
```
aurel-l, line 38: Is this needed here though?
I understand that on the entry page it's to match the current website, but here it's not displayed in the result page on the current website.
Also, I'm not sure how useful that is for a user (but I don't know how this page is used), so I wouldn't go further than just matching the current website does.

xwatkins: I had added for consistency as it was requested in the table. But I guess it doesn't really matter, so removed it.

aurel-l (approved):

PR #520, opened 2022-01-10
2022 01 form creation

xwatkins:
## Purpose
Add form generation and POST

## Approach
Create FormData from passed controlled element data, putting the data in the relevant FormData sections.

## Testing
Snapshot test to check expected output. Also a test for entry update Subject.

## Checklist


aurel-l (approved):
Just as a comment, I would probably prefer some kind of custom hook handling all the async logic in the form (fetch token, post form with token, send loading/error state back to form (_à la_ useDataApi), dispatch popups, navigate back).
Either a hook using the `postContactForm` that is in this PR, or replacing `postContactForm` to make it a hook instead of just a plain function. But it doesn't need to be in this PR, and we could refactor later

xwatkins:
Yes, we can improve towards the end of the week  - although as it's unlikely to be used anywhere else I'm not sure it would justify being abstracted into a hook. Let's see later.
I've just pushed some changes to do with form submission, would you have to have another quick look before I merge it in?

aurel-l (approved):

PR #529, opened 2022-01-18
Fetch article and put something on screen

xwatkins:
## Purpose
Fetch an article if a contextual help link is clicked.

## Approach
Add a container to handle fetching of data. Extract `EntryContent` component to display help entry in help center and in ctx help window.

## Testing
None yet

## Checklist


src/help/components/contextual/ContextualHelpContainer.tsx
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  return (
+    <HelpEntryContent
+      data={data}
+      isStale={isStale}
+      handleClick={() => console.log('click!')}
+    />
+  );
+};
```
aurel-l, line 27: You can start using `SlidingPanel` rightaway, size="small". I'll modify the onClose handler in Franklin to make it optional and so that it gives in the callback the reason for which it has been called. So you could have here something like:
```
onClose={(reason) => {
  if (reason === 'button' || reason === 'navigation'){
    setDisplay(false)
  }
}}
```
And ignore the other reason why it's being called (click outside)

aurel-l:
I think the existing component is flexible enough that with little modifications we could use it here. Styling could be done on the uniprot-website side (height: <same height as mockup in em>, margin-top: calc(50vh - <same height as mockup in em>)). And the non closing behaviour could be handle here by choosing for which reasons we allow the contextual help to close

src/help/components/entry/Entry.tsx
```diff
@@ -76,6 +76,59 @@ const cleanTextOptions: IOptions = {
...
+
+  if (!html) {
+    return <ErrorHandler />;
+  }
+
+  return (
+    <>
+      <h1 className={data.categories.includes('faq') ? 'big' : undefined}>
+        {data.title}
+      </h1>
```
aurel-l, line 107: I have concerns about heading hierarchy in a side panel not following the flow of the document, but would need to read more on it though, I don't know what is the good way to do things here

aurel-l (approved):

src/help/components/contextual/styles/contextual-help.module.scss
```diff
@@ -0,0 +1,26 @@
+@import 'franklin-sites/src/styles/colours';
+
+$margin-offset: 1rem;
```
aurel-l, line 3: You could use the margin value from franklin, it happens to be the same

aurel-l: Ah, but I see it's to cancel the normal margin, OK, I don't know then

src/help/components/contextual/ContextualHelpContainer.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { help as helpUrl } from '../../../shared/config/apiUrls';
+import useDataApiWithStale from '../../../shared/hooks/useDataApiWithStale';
+import { HelpEntryResponse } from '../../adapters/helpConverter';
+import { HelpEntryContent } from '../entry/Entry';
+
+import styles from './styles/contextual-help.module.scss';
+
+const ContextualHelpContainer = ({ articleId }: { articleId: string }) => {
+  const { data, loading, error, status, progress } =
+    useDataApiWithStale<HelpEntryResponse>(helpUrl.accession(articleId));
```
aurel-l, line 12: If you don't use the stale flag you could just use useDataApi instead of useDataApiWithStale

aurel-l (approved):

src/help/components/contextual/styles/contextual-help.module.scss
xwatkins: Actually hadn't realised the sliding panel had a `title` prop which does what we want (apart from colour)...

src/help/components/contextual/styles/contextual-help.scss
```diff
@@ -9,3 +11,15 @@
     vertical-align: super;
   }
 }
+
+// Note: this isn't in a module because of the
+// franklin overrides. can be moved to modules
+// if sliding panel heading colour can be specified
+.contextual-help-panel {
```
aurel-l, line 18: You can use `:global` for this. See usage in the rest of the codebase

xwatkins: There's no way allowing having `[data-article-id] { ...}` in the css module is there?

aurel-l: CSS modules should only transform classnames

aurel-l (approved):

PR #530, opened 2022-01-18
Contextual help

aurel-l:
PR for the whole sprint


src/help/components/contextual/styles/contextual-help.scss
```diff
@@ -0,0 +1,10 @@
+// TODO: update style once decided
+[data-article-id]::after {
+  content: 'i';
+  font-size: 0.75rem;
+  font-family: Bookman, Garamond, serif;
+  line-height: 0;
+  position: relative;
+  vertical-align: baseline;
+  top: -1rem;
```
aurel-l, line 9: Probably wanted to play with vertical-align: super;

src/app/components/App.tsx
```diff
@@ -447,6 +448,7 @@ const App = () => {
       <ErrorBoundary fallback={null}>
         <GDPR />
       </ErrorBoundary>
+      <ContextualHelpHandler />
```
aurel-l, line 451: I'd probably wrap that in a similar error boundary as above, we don't want anything problematic in there crashing the whole page

src/app/components/App.tsx
```diff
@@ -23,6 +23,7 @@ import {
 import pkg from '../../../package.json';
 
 import './styles/app.scss';
+import ContextualHelpHandler from '../../help/components/contextual/ContextualHelpHandler';
```
aurel-l, line 26: Should we wrap that in a lazy loader like the rest of the routes? Since it takes a function that returns a promise, we could make it wait a bit in order to prioritise the rest of the page and content. Something like
`lazy(() => sleep(1000).then(() => import(...)))`

src/app/components/App.tsx
aurel-l: I would also probably put in a standalone `Route` matching not /help and not /contact (and anything else that you think about will not need contextual help in order to not display the button and not load the corresponding logic

src/help/components/contextual/ContextualHelpHandler.tsx
```diff
@@ -0,0 +1,36 @@
...
+const eventHandler = (event: MouseEvent) => {
+  const element = event.target as HTMLElement;
+  if (element.dataset.articleId) {
+    console.log(element.dataset.articleId);
+  }
+};
+
+const ContextualHelpHandler = () => {
+  const isHelpResults = useRouteMatch(LocationToPath[Location.HelpResults]);
+  const isHelpEntry = useRouteMatch(LocationToPath[Location.HelpEntry]);
```
aurel-l, line 16: You can skip that one, the previous one will always be truthy (but with the `exact` flag to false) if this one is

src/help/components/entry/Entry.tsx
```diff
@@ -71,9 +72,45 @@ const cleanTextOptions: IOptions = {
...
   transformTags: {
-    ...getTransformTags('h1'),
+    ...getTransformTags(headingLevel),
     a: aTransformer,
   },
+});
+
+export const HelpEntryContent = ({
+  data,
+  handleClick,
```
aurel-l, line 82: Why is this exposed as a prop? I'm unsure what you want to do with it, I would have expected the whole logic to be encapsulated in this component

aurel-l: Still not sure what to do with that so I've ended up overriding this change in a further PR

src/shared/components/entry/SequenceView.tsx
```diff
@@ -94,7 +94,7 @@ export const SequenceInfo = ({
       content: lastUpdateDate,
     },
     {
-      title: 'Checksum',
+      title: <span data-article-id="checksum">Checksum</span>,
```
aurel-l, line 97: Would be good to update the corresponding types in franklin. See TypeScript error.

xwatkins: Updated here https://github.com/ebi-uniprot/franklin-sites/pull/238/commits/312b44112db5ffd1cddddb418d380e97a3010829

PR #531, opened 2022-01-20
contextual help - small screen

aurel-l:
## Purpose
Contextual help logic for small screens

## Approach
When the screen is not tall enough*, change logic:
 - Hide/unmount sliding panel
 - Make in-page help targets go to the corresponding help page directly

Reused the existing logic in `useReducedMotion` but abstracted it so that it can be used with any media query.

*"Not tall enough" = <35em (height of the contextual help panel in the mockup)

## Testing
Manual testing

## Checklist


src/shared/hooks/useMatchMedia.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Specific cases that need to be reused throughout the codebase
+
+// Reduced motion
+
+export const useReducedMotion = () =>
+  useMatchMedia(
+    /** NOTE: This query means that an unsupported browser will default to this
+     * returning a false value of false (as in "please give me animations"). Maybe
+     * we should turn it around and have unsupported default to "reduce"?
```
xwatkins, line 50: I agree we should turn it around. Approving the PR regardless.

xwatkins (approved):

PR #532, opened 2022-01-21
contextual help - side button & content

aurel-l:
## Purpose
Add contextual help side button and start adding some content in it

## Approach
- Move all side buttons (help + feedback) in their own component
- Use router for contextual internal routing logic and reuse components

## Testing
Updated test for existing component
Manual testing for now (behaviour will likely change in the future, so no point at this point)

## Checklist

### DIFF
--- a/src/help/components/entry/Entry.tsx
+++ b/src/help/components/entry/Entry.tsx
@@ -1,5 +1,5 @@
 import { useCallback, MouseEventHandler, useMemo } from 'react';
-import { useHistory, useRouteMatch } from 'react-router-dom';
+import { RouteChildrenProps, useHistory } from 'react-router-dom';
 import { Card, Loader, Message } from 'franklin-sites';
 import { marked } from 'marked';
 import {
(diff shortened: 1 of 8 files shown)


aurel-l:
Realised that the branch name is living one year in the past

src/help/components/entry/Entry.tsx
```diff
@@ -77,15 +76,36 @@ const getCleanTextOptions = (headingLevel: HeadingLevels): IOptions => ({
...
-  upperHeadingLevel?: HeadingLevels;
-}) => {
+}: HelpEntryContentProps) => {
+  const history = useHistory();
+
+  // Hijack clicks on content
+  const handleClick = useCallback<MouseEventHandler<HTMLElement>>(
+    (event) => {
+      // Hijack clicks on content but only interested in link clicks
+      if (event.target instanceof HTMLAnchorElement) {
```
xwatkins, line 94: This might addressed in another PR, but this causes issues with other links in news articles (e.g. to advanced search results, etc...)

aurel-l: This logic needs to be handled in the "catch all" logic in the memory router's switch, in a further PR indeed

xwatkins (approved):

PR #534, opened 2022-01-24
2022_02-contextual_hash

xwatkins:
## Purpose
Add proteomes links. Some contain a link to a subsection of a help article.

## Approach
Split the :accession by '#'. If there is a hash, scroll to it once the article is loaded.

## Checklist

### DIFF
--- a/src/help/components/entry/Entry.tsx
+++ b/src/help/components/entry/Entry.tsx
@@ -79,21 +79,19 @@ const getCleanTextOptions = (headingLevel: HeadingLevels): IOptions => ({
 type HelpEntryContentProps = {
   data: HelpEntryResponse;
   upperHeadingLevel?: HeadingLevels;
-  hash?: string;
 };
 
 export const HelpEntryContent = ({
   data,
   upperHeadingLevel = 'h1',
-  hash,
 }: HelpEntryContentProps) => {
   const history = useHistory();
 
   useEffect(() => {
-    if (hash) {
-      document.getElementById(hash)?.scrollIntoView();
-    }
-  }, [hash]);
+    document
+      .getElementById(history.location.hash.substring(1))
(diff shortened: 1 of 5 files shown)


src/help/components/entry/Entry.tsx
```diff
@@ -139,7 +147,9 @@ const HelpEntry = ({
   match,
   inPanel,
 }: RouteChildrenProps<{ accession: string }> & Props) => {
-  const accession = match?.params.accession;
+  const [accession, hash] = match?.params.accession?.split(
+    encodeURIComponent('#')
+  ) || [match?.params.accession, undefined];
```
aurel-l, line 152: Wait, how does the hash end up in the accession? This is supposed to be extracted from the pathname 🤔 
Is this an issue with the way we're generating the links?
It probably needs to split the hash there and pass it as a hash, otherwise the hash sign is probably getting encoded in order to be part of the pathname

xwatkins: I played a bit with trying to add it in the path definition `/:accession#:hash?` but didn't get anywhere. Do you know if there's a way?

xwatkins: oh hang on, I see what you mean, need to look at `generatePath`

xwatkins: Have another look. Still have to do the split but higher up so I can use hash in path

aurel-l: I would have done the split in the body of `ContextualHelpContainer` where we need it separated.
To be fair if the reference we put in `articleId` might be an article ID, and article ID + a hash, or even an external URL, I'd rename that state.
But it also depends on where you'd want to put the `window.open` to open a new tab when it's an external URL in the data attribute.

aurel-l (approved):

PR #536, opened 2022-01-26
2022 02 colheaders

xwatkins:
## Purpose
Add tooltips for column headers

## Approach
Create config file to host the definition. Wire in to the data-table component.

## Testing
none

## Checklist

### DIFF
--- a/src/shared/components/results/__tests__/__snapshots__/ResultsData.spec.tsx.snap
+++ b/src/shared/components/results/__tests__/__snapshots__/ResultsData.spec.tsx.snap
@@ -3968,7 +3968,11 @@ exports[`ResultsData component should render the table view 1`] = `
           <th
             class=""
           >
-            Entry
+            <span
+              aria-expanded="false"
+            >
+              Entry
+            </span>
           </th>
         </tr>
       </thead>
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "core-js": "3.20.2",
     "d3": "5.16.0",
     "deep-freeze": "0.0.1",
-    "franklin-sites": "0.0.182",
+    "franklin-sites": "0.0.183",
     "history": "4.10.1",
     "idb": "7.0.0",
     "interaction-viewer": "3.4.0",
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "core-js": "3.20.2",
     "d3": "5.16.0",
     "deep-freeze": "0.0.1",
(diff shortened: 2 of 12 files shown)


src/shared/components/results/__tests__/__snapshots__/ResultsData.spec.tsx.snap
```diff
@@ -3968,7 +3968,11 @@ exports[`ResultsData component should render the table view 1`] = `
           <th
             class=""
           >
-            Entry
+            <span
+              aria-expanded="false"
+            >
```
aurel-l, line 3973: Is this an artifact of the tooltip library?

xwatkins: It looks like it - https://github.com/tvkhoa/react-tippy#accessibility

aurel-l (approved):
Looks OK in this PR, but:
 - Any reason why you used fragments over plain strings when there is no "more" link? (e.g. tooltip for `Checksum`)
 - Clicking on the "more" link triggers the sorting, and doesn't follow the link. But I'm guessing this needs to be solved in Franklin rather than in this PR, right?

xwatkins:
> Looks OK in this PR, but:
> 
> * Any reason why you used fragments over plain strings when there is no "more" link? (e.g. tooltip for `Checksum`)
That comes from Dan's script, and I had to copy/paste all of them manually... had a quick check at the beginning but missed that. I guess it doesn't really matter in the big scheme of things?

> * Clicking on the "more" link triggers the sorting, and doesn't follow the link. But I'm guessing this needs to be solved in Franklin rather than in this PR, right?
Yes, that reminds me I need to make a Jira.

aurel-l:
> > Looks OK in this PR, but:
> > 
> > * Any reason why you used fragments over plain strings when there is no "more" link? (e.g. tooltip for `Checksum`)
> >   That comes from Dan's script, and I had to copy/paste all of them manually... had a quick check at the beginning but missed that. I guess it doesn't really matter in the big scheme of things?

No it doesn't, I was just wondering if at some point there was a type issue that got resolved. But I don't think it makes any difference.

> 
> > * Clicking on the "more" link triggers the sorting, and doesn't follow the link. But I'm guessing this needs to be solved in Franklin rather than in this PR, right?
> >   Yes, that reminds me I need to make a Jira.

OK, good that it is something that you had already noticed and that we are keeping track of

PR #542, opened 2022-02-02
2022 02 links to supporting data

aurel-l:
## Purpose
Add links to supporting data pages (instead of current website's contextual help) -> from sprint board
Also, https://www.ebi.ac.uk/panda/jira/browse/TRM-26988 & https://www.ebi.ac.uk/panda/jira/browse/TRM-26761

## Approach
On UniProtKB pages, replace where we refer to databases, subcellular locations, and keywords, with links.
Also, tackled the 2 above jiras as part of this because I touched the subcell location part.
For subcell location, link to ID.
For Databases and remaining Keywords (categories), we don't have IDs in the data so link to a search and redirect if we can (if only 1 result). That's the best I could do with the data we had at this moment.

## Testing
Updated snapshots + manual checks

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/XRefView.tsx
+++ b/src/uniprotkb/components/protein-data-views/XRefView.tsx
@@ -354,7 +354,7 @@ const XRefsGroupedByCategory = ({
         <Link
           to={{
             pathname: LocationToPath[Location.DatabaseResults],
-            search: `query=(name:${databaseInfo.name})&direct`,
+            search: `query=(name:${databaseInfo.displayName})&direct`,
           }}
         >
           {databaseInfo.displayName}
--- a/src/uniprotkb/components/protein-data-views/XRefView.tsx
+++ b/src/uniprotkb/components/protein-data-views/XRefView.tsx
@@ -1,6 +1,8 @@
 import { Fragment } from 'react';
 import { isEqual, sortBy, uniqWith } from 'lodash-es';
 import { InfoList, ExternalLink, ExpandableList } from 'franklin-sites';
+import { Link } from 'react-router-dom';
+import { InfoListItem } from 'franklin-sites/dist/types/components/info-list';
 
 import { pluralise } from '../../../shared/utils/utils';
 import * as logging from '../../../shared/utils/logging';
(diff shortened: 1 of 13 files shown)


src/uniprotkb/components/protein-data-views/XRefView.tsx
```diff
@@ -346,30 +344,32 @@ const XRefsGroupedByCategory = ({
...
-    }
-  );
+  ]).map((database) => {
+    const databaseInfo = databaseToDatabaseInfo[database.database];
+    return {
+      title: (
+        <Link
+          to={{
+            pathname: LocationToPath[Location.DatabaseResults],
+            search: `query=(name:${databaseInfo.name})&direct`,
```
dlrice, line 357:
```suggestion
            search: `query=(name:${databaseInfo.displayName})&direct`,
```
Had a look at when `displayName` is different to `name` for all databases and this occurs three times:

```
name: RCSB-PDB != displayName: RCSB PDB
name: SOURCE_MIM != displayName: SOURCE
name: SOURCE_MGI != displayName: SOURCE
```
Using `displayName` doesn't break `RCSB-PDB` (it's a 3d db so we don't link to it anyway) and fixes the `SOURCE` issue.

dlrice (commented):
1. The GO tab of subcell viz no longer includes a icon for locations which are present in the viz
2. Noticed that for 3d structure databases we don't have the opportunity to link to our internal DB entry. I fear adding more stuff there would make it even more complex.
Screenshot 2022-02-03 at 16 38 03

aurel-l:
> 1. The GO tab of subcell viz no longer includes a icon for locations which are present in the viz

That is now fixed

> 2. Noticed that for 3d structure databases we don't have the opportunity to link to our internal DB entry. I fear adding more stuff there would make it even more complex.

I think we might not need to add it in this case, but it's true that they even though they are external links they don't use the external link icon so it might be misleading if users expect to be taken the the DB page. But this is within protvista-structure so we couldn't fix it in the uniprot-website codebase.

dlrice (approved):

PR #543, opened 2022-02-09
add share button

aurel-l:
## Purpose
Add a share button

## Approach
Add a share button to the search results with the expected behaviour as defined on the sprint

## Testing
Manual test

## Checklist

Note the style is not matching the sprint mockup. And there are plain JavaScript approach to do some of the expected behaviour. Would need changes in Franklin so I would delay fixing that until we have progressed more in the sprint so that we don't do changes that we later realise we would need to revert or that are actually not needed


src/shared/components/action-buttons/ShareDropdown.tsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+const CopyLinkWebsite = () => {
+  const dispatch = useDispatch();
+  const ns = useNS() || Namespace.uniprotkb;
+  const [userColumns] = useLocalStorage<Column[]>(
+    `table columns for ${ns}` as const,
+    nsToDefaultColumns(ns)
+  );
+  const [viewMode] = useLocalStorage<ViewMode>('view-mode', ViewMode.CARD);
```
dlrice, line 40: In four locations in the codebase we specify `ViewMode.CARD` as the default view mode. I am wondering if it would be safer to create a `defaultViewMode = ViewMode.CARD` in one location and use this instead.

dlrice (commented):
Safari gives me a menu which I don't know what to do with if I want to just copy the URL:
Screenshot 2022-02-10 at 11 36 56

src/shared/components/action-buttons/ShareDropdown.tsx
aurel-l: OK, with that.
I would also take the opportunity to change to an enum here too. So it would be replacing the current 0 and 1 by 'card' and 'table' (in localStorage and in the shared URL). Do you think it would be acceptable to do?

aurel-l:
> Safari gives me a menu which I don't know what to do with if I want to just copy the URL: Screenshot 2022-02-10 at 11 36 56

I see, strange. Probably the Share API is supported in a strange way in Safari on desktop. I was expecting it to work as on phones. I will look into detecting that and removing it for Safari on desktop (falling back on plain copy behaviour), or removing it altogether, even for phones.

aurel-l:
Comments addressed

dlrice (approved):

PR #544, opened 2022-02-09
More column tooltips

dlrice:
## Purpose

- Add additional column header tooltips.
- [More should be on a new line aligned to the right](https://www.ebi.ac.uk/panda/jira/browse/TRM-27329)
- [Column headers: "more" link should open the contextual panel](https://www.ebi.ac.uk/panda/jira/browse/TRM-27330)
- [Prevent sort when click within tooltip](https://www.ebi.ac.uk/panda/jira/browse/TRM-27323)

## Approach

- Created `getLabelAndTooltip` to generate labels and tooltips with the `data-article-id`
- Created `getFeatureLabelAndTooltip` for use with all of the uniprotkb features
- Created `SharedColumnConfiguration` for instances when label, tooltip and render are the same
- Removed `abbreviations.ts` as not used anymore

Excluded tooltips from:

- From
- To
- All Xrefs
- Reviewed
- UniRef cluster type

Shortened tooltips:

- BUSCO
- CPD

## Testing
Updated snapshots

## Checklist

## Todo:

- ~[Tooltip "more" link / sortable column issue](https://www.ebi.ac.uk/panda/jira/browse/TRM-27323)~
- [Add toooltips to UniParcXRefsColumnConfiguration](https://www.ebi.ac.uk/panda/jira/browse/TRM-27419)
- ~Remove `i` in column header~


src/shared/config/ColumnConfiguration.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ...getLabelAndTooltip(
+      'Organism',
+      'Scientific name (and synonyms) of the source organism',
+      'organism-name'
+    ),
+    render: (data: Schema) => {
+      const taxonomyData = getter(data);
+      return taxonomyData && <TaxonomyView data={taxonomyData} />;
+    },
+  }),
```
aurel-l, line 35: That `<Schema,>` got me thinking if it was a typo. I understand this is to avoid this being parsed as JSX. Wondering if the below suggestion would be acceptable in order to avoid this, even though it wouldn't use an arrow function then (equivalent, but not consistent)
```suggestion
  organism<Schema>(getter: (data: Schema) => TaxonomyDatum | undefined) {
    return {
      ...getLabelAndTooltip(
        'Organism',
        'Scientific name (and synonyms) of the source organism',
        'organism-name'
      ),
      render: (data: Schema) => {
        const taxonomyData = getter(data);
        return taxonomyData && <TaxonomyView data={taxonomyData} />;
      },
    };
  },
```

aurel-l (approved):
Good to go for me.
Just left a non-blocking suggestion around TypeScript syntax.
I've also added a couple of commits covering 2 of the todos, so feel free to have a look if and comment if you want

PR #549, opened 2022-02-11
2022 03 share url  buttons

xwatkins:
## Purpose
Add generate URL for API logic to download panel

## Approach
- Moved some of the Preview logic out of Download
- Created component for the API URL which provides copy-to-clipboard functionality
- Modified styles so that the download action buttons are set within a white band that expands the whole width of the panel
- Created `copySuccessMessage` and `copyFailureMessage` actions

## TODO
- [ ] I think more of the logic around the url creation could be simplified

## Testing
- Added unit tests

## Checklist


aurel-l (approved):
Looks good to me.

I've added some commits regarding:
 - scrolling the added content into view (expecially useful when there's the whole column selection in the middle)
 - Clarify/comment the reason of some CSS
 - Replace the logic to download the preview content to use useDataApi and all the logic already included with it.

So @dlrice please have a review of these bits if you wish

aurel-l (approved):

PR #552, opened 2022-02-16
2022 03 share url consume v2

dlrice:
## Purpose
Consume view and fields parameters from the url.

## Approach
- Create two new hooks: `useViewMode` and `useColumnNames` which holds most of the logic to determine which of localStorage or the url should be used.
- If invalid param/value send a message to user
- Keep view and fields in URL when switching viewMode and sorting columns

## Testing
- Updated tests

## Checklist

### DIFF
--- a/src/shared/hooks/useColumnNames.ts
+++ b/src/shared/hooks/useColumnNames.ts
@@ -1,0 +1,53 @@
+import { partition } from 'lodash-es';
+import qs from 'query-string';
+import { Dispatch, SetStateAction } from 'react';
+import { useLocation } from 'react-router-dom';
+import { IDMappingColumn } from '../../tools/id-mapping/config/IdMappingColumnConfiguration';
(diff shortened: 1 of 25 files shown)


src/shared/hooks/useColumnNames.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const useColumnNames = (
+  namespaceOverride: Namespace | undefined,
+  displayIdMappingColumns?: boolean
+): [
+  Column[],
+  Dispatch<SetStateAction<Column[]>>,
+  boolean,
+  InvalidParamValue | undefined
+] => {
```
aurel-l, line 24: I think at that point it should probably return an object with sensible field names, because this causes a situation where it's used like
```ts
const [, , , invalidColumns] = useColumnNames(namespaceOverride);
```
in an other file which is a bit strange I think.

If `useViewMode` is similar enough in its approach I'd change it there too.

aurel-l (approved):
looks good to me

PR #556, opened 2022-02-23
2021 03 peptide to column

dlrice:
## Purpose
[Missing sequence matching column](https://www.ebi.ac.uk/panda/jira/browse/TRM-26786)

## Approach

- Within the peptide search results component augment the API data with a `peptideSearchMatches` property which lists the start and end (0-based) indices for each match.
- Destructure args with useColumnNames.
- Caveat: due to current backend limitations this PR won't allow users to view other's jobs.

## Testing
- Snapshots and unit tests.
- Peptide search results now in table view mode

## Checklist


aurel-l (approved):
OK for me,

Although, I've changed the logic in `getMatches` in order to get multiple matches within the same sequence, including when that matching sequence is repeated multiple times within itself (see added tests for behaviour).
Please do check if you're happy with this addendum to this PR

PR #565, opened 2022-03-10
modify component to display input info + add ability to render a link…

aurel-l:
## Purpose
Add Input Parameters tab content + https://www.ebi.ac.uk/panda/jira/browse/TRM-26895

## Approach
Input Parameters component was already used for both BLAST and Align, so added more cases to handle Peptide Search and ID Mapping.
Handle case when viewing someone else's Peptide Search result.
Extra: Change the "From" column in order to add a link whenever we can. Pass the from namespace as a local context wrapping the table in order to avoid doing weird things for the column generation logic.

## Testing

## Checklist

### DIFF
--- a/src/tools/components/InputParameters.tsx
+++ b/src/tools/components/InputParameters.tsx
@@ -56,10 +56,9 @@ const InputParameters = ({
 
   return (
     <>
-      <ResubmitButton
-        inputParamsData={inputParamsData.data}
-        jobType={jobType}
-      />
+      {inputParameters && (
+        <ResubmitButton inputParamsData={inputParameters} jobType={jobType} />
+      )}
       <section>
         <p>
           The job with ID <code>{id}</code> has been submitted with these raw
--- a/src/tools/components/InputParameters.tsx
+++ b/src/tools/components/InputParameters.tsx
@@ -27,10 +27,14 @@ const InputParameters = ({
   inputParamsData,
   jobType,
 }: InputParametersProps) => {
-  if (inputParamsData && !('type' in inputParamsData)) {
-    if (inputParamsData.error || !inputParamsData.data) {
-      return <ErrorHandler status={inputParamsData.status} />;
-    }
+  if (
+    inputParamsData &&
+    // This is for TS to typeguard, after that we're sure it's not a local job
+    !('type' in inputParamsData) &&
+    // We now have a data payload for sure, check for errors normally
+    (inputParamsData.error || !inputParamsData.data)
+  ) {
+    return <ErrorHandler status={inputParamsData.status} />;
   }
 
   if (!inputParamsData) {
--- a/src/tools/components/InputParameters.tsx
+++ b/src/tools/components/InputParameters.tsx
@@ -9,6 +9,8 @@ import { PublicServerParameters } from '../types/toolsServerParameters';
 import { JobTypes } from '../types/toolsJobTypes';
 import { FinishedJob } from '../types/toolsJob';
 
+import styles from './styles/input-parameters.module.css';
+
 type InputParametersProps = {
   id: string;
   // No public endpoint to expose this for peptide search, so for now replace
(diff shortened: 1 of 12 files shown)


src/tools/components/InputParameters.tsx
```diff
@@ -8,43 +7,73 @@ import { UseDataAPIState } from '../../shared/hooks/useDataApi';
...
+      : inputParamsData.data;
+
   return (
     <>
-      <ResubmitButton
-        inputParamsData={inputParamsData.data}
-        jobType={jobType}
-      />
+      {inputParameters && (
+        <ResubmitButton inputParamsData={inputParameters} jobType={jobType} />
```
dlrice, line 60: For consistency with blast/align what do you think about also listing this in the overview tab

src/tools/components/InputParameters.tsx
```diff
@@ -8,43 +7,73 @@ import { UseDataAPIState } from '../../shared/hooks/useDataApi';
...
           <InfoList
-            infoData={Object.entries(data).map(([key, value]) => ({
-              title: key,
-              content: <CodeBlock lightMode>{`${value}`}</CodeBlock>,
-            }))}
+            infoData={Object.entries(inputParameters || {})
+              .filter(([key, value]) => !fieldsToHide.has(key) && value)
+              .map(([key, value]) => ({
+                title: key,
+                content: <CodeBlock lightMode>{`${value}`}</CodeBlock>,
```
dlrice, line 75: Screenshot 2022-03-11 at 10 57 52

Would it be possible to wrap long sequences?

src/tools/id-mapping/components/results/FromColumn.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if (
+    // Not UniProtKB
+    fromNamespace !== Namespace.uniprotkb ||
+    // Or, UniProtKB + "from" is an accession
+    reUniProtKBAccession.test(from)
+  ) {
+    return <AccessionView id={from} namespace={fromNamespace} />;
+  }
+
+  // if it's a UniProtKB ID, return a search with the "direct" flag
```
dlrice, line 32: Not sure when we would get here? When running your branch it seems to link directly to the uniprotkb entry

src/tools/components/InputParameters.tsx
aurel-l: I think it might be good, yes, but we'll need to mess up with the results buttons that is used in a lot of other places, so I'll add it as a sticky and if we can't get to it soon as a jira

src/tools/components/InputParameters.tsx
```diff
@@ -8,43 +7,73 @@ import { UseDataAPIState } from '../../shared/hooks/useDataApi';
...
   inputParamsData,
   jobType,
-}) => {
-  const { loading, data, error, status } = inputParamsData;
+}: InputParametersProps) => {
+  if (inputParamsData && !('type' in inputParamsData)) {
+    if (inputParamsData.error || !inputParamsData.data) {
+      return <ErrorHandler status={inputParamsData.status} />;
+    }
+  }
```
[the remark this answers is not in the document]
aurel-l: Merged everything into a single `if`, but commented the different parts of the condition

src/tools/id-mapping/components/results/FromColumn.tsx
aurel-l: It's when you look for an ID, like "A4_HUMAN", and in theory it should redirect you to the entry page, but quickly passing through the search result logic

src/tools/components/InputParameters.tsx
aurel-l: Quickly touched a bit this area in order to try to solve this but didn't manage to do so quickly. It feels like I'd have to fight against some patterns within the InfoList, DecoredListItem, and CodeBlock components in Franklin.
For now I just made it so that it scrolls the sequence section if it's too big and not the whole page.
Feel free to have another go at it if you want.

dlrice (approved):

PR #570, opened 2022-03-14
2022 04 tools adv search messages

dlrice:
## Purpose
Display messages to users that queries are not possible with align and ID mapping to non-uniprot DBs. Show this message when either align or ID mapping with non-uniprot DB is selected in the following places

1. The main search
2. The query builder

## Approach
Created a context to get id mapping details as this is used in three places:

1. The main search
2. The query builder
3. The actual ID mapping results

## Testing
Updated

## Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -29,8 +29,11 @@ merge_and_test:
     - echo Merging $CI_COMMIT_SHORT_SHA into $CI_EXTERNAL_PULL_REQUEST_TARGET_BRANCH_NAME to test upstream effects
     - git config --global user.email "uniprotci@gmail.com"
     - git config --global user.name "UniProt CI"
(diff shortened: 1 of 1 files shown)


src/tools/id-mapping/components/results/IDMappingResult.tsx
```diff
@@ -84,13 +85,13 @@ const IDMappingResult = () => {
   );
 
   const databaseInfoMaps = useDatabaseInfoMaps();
+  const detailsData = useIDMappingDetails();
 
   const [{ selectedFacets, query }] = getParamsFromURL(location.search);
 
   const detailApiUrl =
```
dlrice, line 92: Note to self: this could possibly be removed

aurel-l (approved):

PR #571, opened 2022-03-15
first attempt at component-level lazy loading

aurel-l:
## Purpose
Try to improve the overall performance of the website

Note that it makes the content of the lazy loaded component not findable through ctrl-F search, so it is best for visual components and not for plain text or table of data

## Approach
Add a lazy loader component approach where some bits of the page are not loaded until the user scrolls to their position.
It aims at:
 - improving memory usage on big pages (where a user might never see all of the page)
 - reduce number of blocking and concurrent network requests to favour the ones needed by the user to view the bit of the page currently in view

See indicative performance table below

## Testing
Manual tests + updated snapshots

## Checklist

Comparison table:

|  | Before |   |   |   | After |   |   |  
|-- | -- | -- | -- | -- | -- | -- | -- | --
|  | Home page |   | P05067 |   | Home page |   | P05067 |  
|  | Initial load | Scrolled to bottom | Initial load | Scrolled to bottom | Initial load | Scrolled to bottom | Initial load | Scrolled to bottom
| # of requests | 41 | 41 | 69 | 69 | 37 | 42 | 60 | 79
| Loaded resources (MB) | 13.2 | 13.2 | 41.6 | 41.6 | 13.1 | 13.2 | 21.4 | 41.8
| “Finish” time (s) | 1.69 | N/A | 10.97 | N/A | 1.57 | N/A | 7.47 | N/A  
| Heap size (MB) | 41.6 | 41.8 | 144 | 144 | 34.8 | 34.9 | 85.4 | 152

Disclaimers regarding this perf evaluation:
- Run on chrome devtool, on my machine, from the same place. Expect variability in data. Also, run on a localhost development build, expect better performance on production build
- Mainly to compare before/after, not for absolute values
- Finish time as reported by the network panel, doesn't reflect visual completeness of the view
- Heap size is after running a manual garbage collection
- Scrolled to bottom is by manually slowly scrolling through the page in order to load all of the lazy loaded components


src/app/components/home-page/__tests__/__snapshots__/NonCritical.spec.tsx.snap
```diff
@@ -878,24 +878,7 @@ exports[`non-critical HomePage component should render 1`] = `
...
-        >
-          <span
-            class="visually-hidden"
-          >
-            Welcome to UniProt || UniProt introduction
-          </span>
-        </button>
-      </lite-youtube>
-    </div>
+    />
```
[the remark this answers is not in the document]
aurel-l: So I checked a bit the coverage file by file on the ones that are wrapped by that LazyComponent. All the part that is not lazy loaded but lazy rendered is still covered/rendered in jest tests as jsdom doesn't support IntersectionObserver so it defaults to not do lazy rendering at all for these. And for the parts that are wrapped in a lazy wrapper from react and separated to there own chunks, it's either directly external libraries, or it's components that are somehow covered through a non-lazy test or another test that would be async and waiting for the component to be loaded and renderer. Not sure where the covereage is coming from to be honest, but it's there.

src/shared/components/entry/SequenceView.tsx
```diff
@@ -41,6 +41,7 @@ import {
 import { Namespace } from '../../types/namespaces';
 
 import styles from './styles/sequence-view.module.css';
+import LazyComponent from '../LazyComponent';
```
dlrice, line 44: Import should go above style imports

src/shared/components/entry/SequenceView.tsx
```diff
@@ -100,21 +101,29 @@ export const SequenceInfo = ({
...
-      infoData={infoData}
-      accession={isoformId}
-      downloadUrl={apiUrls.sequenceFasta(isoformId)}
-      onBlastClick={() =>
-        history.push(LocationToPath[Location.Blast], {
-          parameters: { sequence: dataToDisplay?.value },
-        })
+    <LazyComponent
+      fallback={
+        <div style={{ wordBreak: 'break-all' }}>
```
dlrice, line 106: Is there a benefit to inline styles in this instance?

dlrice (approved):

src/shared/components/entry/SequenceView.tsx
aurel-l: I don't think so. We're already loading a stylesheet with this page so I've added a classname for this instead of an inline style

src/shared/components/entry/SequenceView.tsx
aurel-l: Right, I missed that one, it is fixed now

PR #572, opened 2022-03-16
custom script to generate a sitemap

aurel-l:
## Purpose
Generate a sitemap

## Approach
 - Use public API (so not the best for perf, and needs to run within VPN to not be rate-limited)
 - Generate an index sitemap pointing to x other sitemap files containing each up to 5000 URLs
 - Heavy use of generators and streams in order to avoid any memory issue. Assuming that API will be slower than our speed generating the files, but otherwise (if we hook to a quicker source of data) we might need to start handling backpressure
 - At the moment, only use query in UniProtKB for human OR reviewed

## Testing
Manual checks

## Checklist

### DIFF
--- a/deploy/sitemap/index.js
+++ b/deploy/sitemap/index.js
@@ -1,0 +1,146 @@
+#!/usr/bin/env node
+/* eslint-disable no-console, @typescript-eslint/no-var-requires */
+const fs = require('fs');
+const axios = require('axios');
+const ProgressBar = require('progress');
+
+const accessionCountPerFile = 50_000;
+const publicPath = 'https://www.uniprot.org/';
+// const query = '*';
+const query = '(organism_id:9606) OR (reviewed:true)';
+
+const nextRE = /<([0-9a-zA-Z$\-_.+!*'(),?/:=&%]+)>; rel="next"/;
+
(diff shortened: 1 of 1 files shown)


dlrice:
Could we generate this when building the `main` branch as part of CI/CD?

aurel-l:
> Could we generate this when building the `main` branch as part of CI/CD?

So, we could, but we wouldn't really need that to happen every time we release on the main branch as the generated set of URLs is valid for the whole length of time, and it does take a certain amount of time to run (depending on the query we choose to include in the sitemap logic) and incurs a load on the public API.
This should theoretically be run on data release, and part of the process release we would only copy these sitemap files to somewhere public just like the rest of static assets. That's why it's probably something that needs to be done by backend or in the production process, the closer to the data source as possible.

We should probably have a call between us at some point to see how far we want to go with this approach as I'm not aiming at having this kept in the long term, but we might have to...

An option could also be to have a specific pipeline to be run on demand and not attached to any push on any branch.

dlrice:
During CI/CD we could even check what the version is in the served sitemap and if out of date then regenerate.

aurel-l:
Work carried over in this over independant repo https://github.com/ebi-uniprot/uniprot-sitemap

PR #573, opened 2022-03-16
2022 04 download

dlrice:
## Purpose

- Package upgrades
- Fix downloads for:
  - UniSave
  - UniRef entry
  - UniParc entry
  - ID Mapping


aurel-l (approved):

aurel-l:
Merging now as there is a user that just submitted feedback about something that would be fixed with this PR

PR #574, opened 2022-03-16
2022 04 download uniparc & uniref

dlrice:
## Purpose

- Add TSV format to existing download dropdown in UniParc entry
- Add List format to existing download dropdown in UniRef entry 

## Approach

- Add new `apiUrls` for uniparc with databases endpoint
- Add exception for uniparc && tsv in `EntryDownload`
- Open file format selections in new tab
- Updated uniref members download to accommodate file format
- Blocked by not having tsv and xml for uniref members ([Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-27641))

## Testing
Added unit tests

## Checklist


aurel-l:
Actually, same issue with UniRef members

dlrice:
Good catches. Regarding streaming, I think the only thing we could possibly do is to ask backend to provide this. Regarding the uniparc customise column issue, I'll start a chat in the UX channel.

aurel-l:
I think a quick fix for now would be to set size=500 by default for these cases. Most of the time people won't notice that there's a problem (while with the default of 25 they definitely will). And then we can ask backend to add the streaming endpoint later.

aurel-l (approved):
Approving in theory, but I guess pending discussion with Marija

PR #576, opened 2022-03-17
2022 04 download unisave

aurel-l:
## Purpose
UniSave download. Roughly https://www.ebi.ac.uk/panda/jira/browse/TRM-27035

## Approach
- In table, make (txt) and (fasta) links to open the corresponding files in a new tab (similar to current website)
- Add UniSave namespace
- Add download button
- Modify Download component and underlying logic to cope with the specificities of UniSave:
  - No compressed option
  - No size limit
  - No column selection
  - Dedupe of sequences in FASTA format available through specific query field

Note that comparing with the current website, we now have a TSV option

I also had to de-optimise the logic in useUserSelect that was going against React normal flow of things. It was not really a problem in production, but it made it impossible to select anything in development. The only thing is that a "select all" on a big table might take longer (because each checkbox ticked triggers a state change). In React 18 (coming soon) we will have batching of these changes, so hopefully this issue will just disappear on its own soon.

## Testing
Existing tests still pass + added small specific test

## Checklist


dlrice:
1. Could we check if all were versions were selected and if so, drop the `versions=` param to increase server performance (just tried and selecting all versions then downloading caused a 500).
2. Why don't limit the preview number? I ask as it can take a long time to show lots of selected versions.

aurel-l:
> 1. Could we check if all were versions were selected and if so, drop the `versions=` param to increase server performance (just tried and selecting all versions then downloading caused a 500).

Handled

>     2. Why don't limit the preview number? I ask as it can take a long time to show lots of selected versions.

Since we don't have the "size" parameter I first didn't do any limit in the preview, agreed with the Marija. But this morning I realised I could work around by passing just the first 10 selectedEntries, even though then it's more UniSave-specific code which I'm not a big fan of. This is now implemented, except for the "Download all" case

dlrice (approved):

PR #577, opened 2022-03-18
Add sorting to ID mapping results.

dlrice:
## Purpose
Get column sorting in ID mapping results working
https://www.ebi.ac.uk/panda/jira/browse/TRM-27602

(Peptide search sorting doesn't work because the accessions endpoint doesn't currently support this but I'm about to write a jira for this)

## Approach

- When updating column sort use `history.location.pathname` rather than the corresponding `SearchResultsLocations[namespace]`
- Add `sort` param to ID mapping results API request

## Testing
Manual

## Checklist

### DIFF
--- a/src/shared/components/results/__tests__/Results.spec.tsx
+++ b/src/shared/components/results/__tests__/Results.spec.tsx
@@ -42,21 +42,21 @@ describe('Results component', () => {
     fireEvent.click(columnHeader);
     await waitFor(() => {
       expect(history.location.search).toBe(
-        '?query=blah&sort=accession&dir=ascend'
+        '?dir=ascend&query=blah&sort=accession'
       );
     });
     columnHeader = await screen.findByText('Entry');
     fireEvent.click(columnHeader);
     await waitFor(() => {
       expect(history.location.search).toBe(
-        '?query=blah&sort=accession&dir=descend'
+        '?dir=descend&query=blah&sort=accession'
       );
     });
     columnHeader = await screen.findByText('Entry');
     await waitFor(() => {
       fireEvent.click(columnHeader);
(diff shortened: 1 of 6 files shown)


aurel-l:
I'm getting a `/overview?&sort=id&dir=ascend` URL (notice the `?&`). While this still works, it doesn't look really nice

aurel-l:
> Add sort param to ID mapping results API request
Should that be also in the URL generated for the API?

dlrice:
> > Add sort param to ID mapping results API request
> 
> Should that be also in the URL generated for the API in the panel?

Yes, I'm seeing this happening for tabular files - do you have an example where this isn't happening?

aurel-l:
> > > Add sort param to ID mapping results API request
> > 
> > 
> > Should that be also in the URL generated for the API in the panel?
> 
> Yes, I'm seeing this happening for tabular files - do you have an example where this isn't happening?

Ah I see. So is this on purpose that we don't allow that in non-tabular format? I'm just thinking that it might be useful to order FASTA format by entry name for example, or by any field, even though the format doesn't allow to display that specific column.

This is not blocking then, since I see that this is working with tabular format

aurel-l (approved):

dlrice:
> Ah I see. So is this on purpose that we don't allow that in non-tabular format? I'm just thinking that it might be useful to order FASTA format by entry name for example, or by any field, even though the format doesn't allow to display that specific column.

I hadn't even thought of this - I agree that could be useful. I think this would be a good improvement and the backend does support this. I've had a look and it's a simple code change so will just put it in this PR

aurel-l (approved):

PR #579, opened 2022-03-21
fix selection

aurel-l:
## Purpose
Fix selection in ID mapping results

## Approach
Clean-up the "from" in the row ID wherever the selection is used.

Not a big fan of the approach

## Testing
Added 1 new test set

## Checklist

### DIFF
--- a/src/shared/utils/getIdKeyForNamespace.ts
+++ b/src/shared/utils/getIdKeyForNamespace.ts
@@ -28,7 +28,14 @@ const fromSeparator = '|';
 
 export const fromCleanMapper = (entry: string) => {
   if (entry.includes(fromSeparator)) {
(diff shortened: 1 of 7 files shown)


src/shared/utils/getIdKeyForNamespace.ts
```diff
@@ -24,6 +24,15 @@ import {
 
 import { Namespace } from '../types/namespaces';
 
+const fromSeparator = '|';
```
dlrice, line 27: What do you think about checking for the presence of the `fromSeparator` in the IDs and if it's there console error so we can catch it in sentry? I know it's unlikely for an ID to have this but just as an insurance

dlrice (approved):

PR #580, opened 2022-03-22
generalise the use of useMatchWithRedirect

aurel-l:
## Purpose
generalise the use of useMatchWithRedirect to entry pages too, and redirect any non-valid tab value to a default subPage instead of crashing

## Approach
Change the hook to cover more use cases, use it in more places

## Testing
Go edit the "subpage" section of the URL for any tool or UniProtKB or UniParc entry pages to either remove it completely or invent something invalid.

This is really an edge-case for if a user edits the URL to at least point to something useful AND in the case of UniProtKB entry in the case where the user followed a now invalid URL (like https://www.uniprot.org/uniprot/P05067/protvista)

## Checklist

### DIFF
--- a/src/shared/hooks/useMatchWithRedirect.ts
+++ b/src/shared/hooks/useMatchWithRedirect.ts
@@ -3,16 +3,22 @@ import { useHistory, generatePath, useRouteMatch } from 'react-router-dom';
 
 import { LocationToPath, Location } from '../../app/config/urls';
 
-export const useMatchWithRedirect = <T extends { subPage?: string }>(
+const useMatchWithRedirect = <T extends { subPage?: string }>(
   location: Location,
-  defaultSubPage: string
+  defaultSubPage: string,
+  possibleSubPages: Record<string, string>
 ) => {
   const history = useHistory();
   const match = useRouteMatch<T>(LocationToPath[location]);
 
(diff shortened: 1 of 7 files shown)


aurel-l:
Wondering if the `useMatchWithEntry` should just take the enum and assume the first value is the default instead of taking both a default value AND an enum.

dlrice:
> Wondering if the `useMatchWithEntry` should just take the enum and assume the first value is the default instead of taking both a default value AND an enum.

That does seem to be a convention we have followed so far.

At the same time I am wondering if we want to specifically remap old subpages eg provista --> feature-viewer ?

aurel-l:
> > Wondering if the `useMatchWithEntry` should just take the enum and assume the first value is the default instead of taking both a default value AND an enum.
> 
> That does seem to be a convention we have followed so far.

So I've inverted the parameters in the signature so that we can make that default tab an optional value, defaulting to the first value of the possible tabs.

> At the same time I am wondering if we want to specifically remap old subpages eg provista --> feature-viewer ?

Yes, I didn't want to hardcode this specific redirect in the hook at the time of implementation so I was originally happy with it just redirecting to the entry subpage, but looking at it again I'm proposing to pass a redirect map wherever it's actually needed and so I've added the logic to process this mapping in the hook.

dlrice (approved):

PR #581, opened 2022-03-23
2022 05 similar proteins fetch refactor

aurel-l:
## Purpose

## Approach

## Testing

## Checklist


src/uniprotkb/components/entry/similar-proteins/SimilarProteinsTable.tsx
```diff
@@ -80,12 +92,22 @@ const SimilarProteinsTable: FC<{ members: string[] }> = ({ members }) => {
...
+        getIdKey={(row) => row.primaryAccession}
+        density="compact"
+      />
+      <Link
+        to={{
+          pathname: LocationToPath[Location.UniProtKBResults],
+          search: `query=${query}`,
+        }}
+      >
+        Show all {total} UniProtKB {pluralise('entry', total, 'entries')}
```
dlrice, line 108:
```suggestion
        Show all ({total}) UniProtKB entries in this cluster
```

- Just trying to avoid the case of `Show all 1 UniProtKB entry` --> `Show all (1) UniProtKB entries in this cluster`
- Thought `in this cluster` might make it a little clearer

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -24,70 +26,70 @@ import {
...
+      };
+      for (const [accession, response] of zip(allAccessions, responses)) {
+        /* istanbul ignore if */
+        if (!accession || !response) {
+          break; // Shouldn't happen, used to restric types
+        }
+        for (const cluster of response.data.results) {
+          let association: IsoformsAndCluster | undefined = mapping[
+            cluster.entryType
+          ].find((association) => association.cluster.id === cluster.id);
```
dlrice, line 75: - Wasn't sure what `association` means here at first sight - what do you think about using `foundIsoformsAndCluster` or even just `found`?
- Would be a little easier to read if the `find`'s arg name wasn't `association` as this is also used in the line above.

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -24,70 +26,70 @@ import {
...
+          if (!association) {
+            association = { isoforms: [], cluster };
+            mapping[cluster.entryType].push(association);
+          }
+          association.isoforms.push(accession);
+        }
```
dlrice, line 81: What do you think about extracting this block into a separate function for easier unit testing?

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
```diff
@@ -24,70 +26,70 @@ import {
...
+  useEffect(() => {
+    setMappingLoading(true);
+    const promises = allAccessions.map((accession) =>
+      fetchData<{
+        results: UniRefLiteAPIModel[];
+      }>(`${apiUrls.search(Namespace.uniref)}?query=(uniprot_id:${accession})`)
+    );
+    Promise.all(promises).then((responses) => {
+      const mapping: Mapping = {
+        UniRef100: [],
```
dlrice, line 63: Could do `UniRef100: {}` and use the `cluster.id` as the key to avoid needing a `find` later on

src/uniprotkb/components/entry/similar-proteins/SimilarProteinsTable.tsx
aurel-l: It's actually not "in this cluster" as it's the UniProtKB entries in the cluster minus the one(s) that we are looking at now.
Hence, if I'm looking at "Similar proteins" to P05067 (canonical), I don't want to see P05067 itself. It's not "similar", it is itself.
But probably need UX input on that on expected behaviour, especially knowing that we have people that are not too keen on highlighting the "cluster" concept in this section

src/uniprotkb/components/entry/similar-proteins/SimilarProteins.tsx
aurel-l: I think I tried to optimise for render. Everything in this section should happen once, but the render might happen multiple times, so that's why I went for an array because this is what we'll use in the jsx section.
Not sure it would make a difference though, and so if it improves overall readability of the code I'd change it.
Will look tomorrow

src/uniprotkb/components/entry/similar-proteins/SimilarProteinsSection.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <Card
+      header={<h2 data-article-id="similar_proteins_section">{name}</h2>}
+      id={id}
+      data-entry-section
+    >
+      <LazyComponent>
+        <SimilarProteins
+          isoforms={isoforms}
+          primaryAccession={primaryAccession}
+        />
```
aurel-l, line 26: Since this is lazy-loaded, I guess you can also create a lazy app logic chunk replacing `import SimilarProteins from './SimilarProteins';` so that we can even save loading this logic until the user gets to this section

aurel-l:
Changed the target branch to main as in the end we didn't end up using the sprint branch 😅

src/uniprotkb/components/entry/similar-proteins/SimilarProteinsSection.tsx
aurel-l: I just did the proposed change in a commit

src/uniprotkb/components/entry/similar-proteins/SimilarProteinsTable.tsx
dlrice: Is our wording good now?

aurel-l: I think it's fine, yes. If we get different feedback about this specifically this is something we can change easily anyway

PR #583, opened 2022-03-23
Id mapping stream

dlrice:
## Purpose
ID mapping download wasn't using the stream endpoint

## Approach
In the Download component determine if it's mapping results and if so also check if it's namespaced and create the appropriate url.

## Testing
manual

## Checklist


aurel-l (approved):
Great 👍🏽 
Will merge it now so that it can be deployed asap

PR #584, opened 2022-03-24
Use data api options

dlrice:
## Purpose
There is an issue at:

https://beta.uniprot.org/peptide-search/PM20220323d25f892452724e5db0ac67d571484bc9

where we are asking for json from the peptide search endpoint but text is returned instead and so causes a json parsing error.

## Approach

- Request text in the header
- Remove redundant headers from `fetchData` args
- Use a ref for options in `useDataApi` to prevent the useEffect continually re-firing

## Testing
TODO: tests need fixing because of a lodash-es issue.

## Checklist


aurel-l (approved):
Went around the issue by doing object spread wherever needed, including the headers object, so feel free to merge if you're happy with this approach.

PR #591, opened 2022-03-30
2022 05 did you mean url params

dlrice:
## Purpose
Generic approach to load any form field data from the URL (similar way as we do currently with accession). Works with Peptide Search's server params with the exception of taxonomy. Created [jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-27685) for validation more generally.

## Approach

- Iterate through all default values and look first for url params, then state values, and finally use default if neither of these are there
- Alignment forms:
  - Check if the user is at an alignment location before fetching accessions
  - Dispatch message if user provides both ids and sequence in 

## Testing
Added more unit tests

## Checklist

### DIFF
--- a/src/tools/hooks/__tests__/useInitialFormParameters.spec.tsx
+++ b/src/tools/hooks/__tests__/useInitialFormParameters.spec.tsx
@@ -79,12 +79,13 @@ describe('useInitialFormParameters: Align', () => {
 describe('useInitialFormParameters: Peptide Search', () => {
   it('should return defaultAlignFormValues and peps, lEQi, spOnly from url parameters', () => {
     const peps = 'ABCDEF';
-
+    const lEQi = 'on';
+    const spOnly = 'off';
     const { result } = customRenderHook(
       defaultPeptideSearchFormValues,
       queryString.stringifyUrl({
         url: LocationToPath[Location.PeptideSearch],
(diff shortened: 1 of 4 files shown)


src/tools/hooks/__tests__/useInitialFormParameters.spec.tsx
```diff
@@ -6,53 +6,106 @@ import axios from 'axios';
...
+  state?: unknown
+) {
+  const history = createMemoryHistory();
+  if (path) {
+    history.push(path, state);
+  }
+  return renderHook(() => useInitialFormParameters(defaultFormValues), {
+    wrapper: ({ children }) => <Router history={history}>{children}</Router>,
+  });
+}
```
aurel-l, line 29: Should that be with, or near, the customRender helper? I'm wondering if there are other hooks that are tested in a similar way and that need to be wrapped in react-router's logic

aurel-l (approved):

src/tools/hooks/__tests__/useInitialFormParameters.spec.tsx
dlrice: Good idea, moved this up with the other test helpers. Look OK?

aurel-l (approved):

PR #592, opened 2022-04-06
2022 05 did you mean initial component

dlrice:
## Purpose
Create a `DidYouMean` component

## Approach

- Created `DidYouMean` which is provided to `NoResultsPage` via a new optional `message` prop.
- Left align of content agreed with @MarijaLuu
- Fetch other namespace suggestions with `fetchData`
- When loading of other namespace suggestions takes longer than 3s display "Sorry no results". 
- Created `searchLocations` to go from `Namespace` to the search location.

## Testing
Created unit test

## Checklist

### DIFF
--- a/src/shared/components/results/DidYouMean.tsx
+++ b/src/shared/components/results/DidYouMean.tsx
@@ -12,32 +12,46 @@ import {
 // TODO: delete line and file if not needed
 // import styles from './styles/did-you-mean.module.scss';
 
-type Props = {
+type SuggestionSentenceProps = {
+  suggestions: Suggestion[];
+  namespace: SearchableNamespace;
+};
+
+const SuggestionsSentence = ({
+  suggestions,
+  namespace,
+}: SuggestionSentenceProps) => (
+  <>
+    {suggestions.map(({ query }, i, a) => [
+      (i > 0 && i < a.length - 1 && ', ') || (i === a.length - 1 && ' or '),
+      <a href="#" key={query}>
+        {query}
+      </a>,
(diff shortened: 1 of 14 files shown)


src/shared/components/results/DidYouMean.tsx
```diff
@@ -0,0 +1,177 @@
...
+            namespace: currentNamespace as SearchableNamespace,
+          },
+        ]
+      : []),
+    ...otherNamespaceSuggestions,
+  ];
+
+  return (
+    <Message level="info" className={styles['did-you-mean-message']}>
+      <h4>Sorry, no results were found!</h4>
```
aurel-l, line 156: This is the only heading on this page, so it should be a h1, with a classname of `small` if we want to keep the size of a h4

src/shared/components/results/DidYouMean.tsx
```diff
@@ -0,0 +1,171 @@
...
+                suggestions={suggestions}
+                namespace={namespace}
+                key={namespace}
+              />
+            ))}
+          </ul>
+        </div>
+      )}
+      Can&apos;t find what you are looking for? Please{' '}
+      <Link to={LocationToPath[Location.ContactGeneric]}>contact us</Link>
```
aurel-l, line 166: It would need to pass the referrer to the contact page. Do check how it's done in other places in the codebase.

If you want, have a think about how to avoid having to remember to add the referrer everytime we link to the contact page page.

aurel-l (approved):

PR #595, opened 2022-04-09
process the search string as passed by the user to do common useful cleanup

aurel-l:
## Purpose & approach
https://www.ebi.ac.uk/panda/jira/browse/TRM-27749 -> trim the input search
https://www.ebi.ac.uk/panda/jira/browse/TRM-27722 -> replaceAll special keywords to uppercase

## Testing
Manual checks

## Checklist

### DIFF
--- a/src/shared/components/search/SearchContainer.tsx
+++ b/src/shared/components/search/SearchContainer.tsx
@@ -104,6 +104,14 @@ export const cannotQueryMessages = {
     'Filtering Align results is not possible as all of its sequences constitute the alignment.',
 };
 
+const reSearchSpecialWords = /(^| )(or|and|not) /gi;
+const rawQueryClean = (raw: string) =>
+  raw
+    // "     app " -> "app"
+    .trim()
+    // "not app and human oR whatever" -> "NOT app AND human OR whatever"
+    .replaceAll(reSearchSpecialWords, (match) => match.toUpperCase()) || '*';
+
 type Props = {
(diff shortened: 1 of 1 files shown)


src/shared/components/search/SearchContainer.tsx
```diff
@@ -104,6 +104,14 @@ export const cannotQueryMessages = {
     'Filtering Align results is not possible as all of its sequences constitute the alignment.',
 };
 
+const reSearchSpecialWords = /(^| )(or|and|not) /gi;
+const rawQueryClean = (raw: string) =>
+  raw
+    // "     app " -> "app"
+    .trim()
```
dlrice, line 111:
```suggestion
    .trim()
    .replace(/\s+/g, ' ');
```
I don't think multiple spaces would ever need to be preserved in a query?

src/shared/components/search/SearchContainer.tsx
```diff
@@ -104,6 +104,14 @@ export const cannotQueryMessages = {
...
+    // "not app and human oR whatever" -> "NOT app AND human OR whatever"
+    .replaceAll(reSearchSpecialWords, (match) => match.toUpperCase()) || '*';
```
dlrice, line 113:
```suggestion
    .replace(reSearchSpecialWords, (match) => match.toUpperCase()) || '*';
```
Thinking that as it's a global search `replace` will be sufficient

dlrice (approved):

PR #600, opened 2022-04-19
variant section

aurel-l:
## Purpose
Variant section changes.
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-26326 (filtering)
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-26279 (filtering)
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-27440 (filtering)
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-27565 (content, changes)
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-27566 (content, changes)
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-26989 (content, missing data)
 - https://www.ebi.ac.uk/panda/jira/browse/TRM-27720 (content, missing data)

## Approach
=> dump every piece of data I could find either in the visible or the hidden rows
 - Display more content in the table
 - Display more "less critical" content in the collapsible row
 - Hook the visualisation filter events to the table content
 - Change EvidenceTag logic in order to render arbitrary URLs that are in the variant evidences instead of trying to generate it
 - Make the sequence changes wrap when too many aa

## Testing
 - Manual testing on P05067 (lots of variants from multiple sources), P61812 (long change at position 229-414), and P61812.
 - Update mock with more data to render more content and call more of the corresponding logic

## Checklist

### Notes
 - When there are a lot of variants, the variation section is now one of the heaviest section to render, even more so now that we render more things. Its visible content needs to be findable through Ctrl-F so there is not an obvious way to render less. The filtering is quite heavy in processing, but only happens on click.
   - Filtering by default for UniProt-reviewed might help in some case, but the whole data needs to be downloaded, parsed, then aditionally filtered (but not rendered if filtered out), so not an obvious improvement either
   - The nightingale logic and filtering could be improved
 - Observations about the variation endpoint in the Proteins API:
   - `begin` and `start` always the same
   - `alternativeSequence`, `mutatedType`, and `variant` (almost) always the same
     - except when `variant` is `*` in which case the others are undefined
   - `evidences` `url` fields always have an extra space at the start


src/uniprotkb/components/protein-data-views/VariationView.tsx
```diff
@@ -90,60 +181,89 @@ const VariationView = ({
...
+                      {i !== 0 && <br />}
+                      {id}
+                    </Fragment>
+                  ))}
+                </td>
+                <td>{position}</td>
+                <td className={styles.change}>
+                  {variantFeature.wildType}
+                  {'>'}
+                  {variantFeature.alternativeSequence || <em>missing</em>}
```
dlrice, line 228: Does `missing` imply something biologically or is it just no data? I ask as it's not immediately clear if it means there's a deletion or no data.

src/uniprotkb/components/protein-data-views/VariationView.tsx
```diff
@@ -90,60 +181,89 @@ const VariationView = ({
...
+                    new Set(variantFeature.xrefs.map((xref) => xref.id))
+                  ).map((id, i) => (
+                    <Fragment key={id}>
+                      {i !== 0 && <br />}
+                      {id}
+                    </Fragment>
+                  ))}
+                </td>
+                <td>{position}</td>
+                <td className={styles.change}>
```
dlrice, line 225: I wonder if a tooltip or help document link would be useful to here to explain the various symbols (eg `*`, `-`)

src/uniprotkb/components/protein-data-views/VariationView.tsx
aurel-l: I don't remember with one it is, I think we had a look in a past website meeting, and the examples we say all happened to be marked as `del` in other fields. In any case, Maria asked to display `missing` in these cases.

src/uniprotkb/components/protein-data-views/VariationView.tsx
aurel-l: Do you think it should be added to the https://beta.uniprot.org/help/variant article? We would need a curator to phrase it correctly from a biological point of view

src/uniprotkb/components/protein-data-views/VariationView.tsx
dlrice: Looking at this they have `Consequence: inframe deletion` which makes me think we should use the lowercase `d` to indicate this but it is instead placed on the `*` line which also has things like `Consequence: stop gained`

Screenshot 2022-04-19 at 14 43 05

aurel-l: I'm not sure what to do to be honest, because the visualisation puts it to the "*" line and I'm not sure what the reason behind it was.
Do we have an example where a variant ends up on the "d" line? In which case we would be able to compare the data behind it

dlrice: Seems the `d` line isn't used. I've looked at a dozen entries with loads of variants but the `d` line is never occupied. Maybe this issue should be included in the variants refactor post-release.

dlrice (approved):

PR #601, opened 2022-04-19
add taxonomy navigation through search page

aurel-l:
## Purpose
Taxonomy entry children navigation https://www.ebi.ac.uk/panda/jira/browse/TRM-27018

## Approach
Since we're not going to get all direct children in the initial payload, I opted to use the navigation as a "preview" of the corresponding search. Make it look like the ExpandableList component, but one will expand to the corresponding taxonomy search for either the direct children (through a "parent" search) or all the descendants (through an "ancestor" search)

## Testing
Added a test for the new ChildNavigation component

## Checklist

### DIFF
--- a/src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
+++ b/src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
@@ -1,0 +1,73 @@
+import { Link } from 'react-router-dom';
+import { Button } from 'franklin-sites';
+
+import { pluralise } from '../../../../shared/utils/utils';
+import {
+  getEntryPath,
+  LocationToPath,
+  Location,
+} from '../../../../app/config/urls';
+
+import { TaxonomyAPIModel } from '../../adapters/taxonomyConverter';
+import { Namespace } from '../../../../shared/types/namespaces';
+
+import styles from './styles/child-navigation.module.css';
+
+// Match the "ExpandableList" default count
+export const CHILDREN_TO_DISPLAY = 5;
+
+const ChildNavigation = ({
+  taxonId,
+  childTaxons,
+  total,
+}: {
+  taxonId: number;
+  childTaxons: TaxonomyAPIModel[];
+  total: number;
+}) => {
+  const remainingChildren = total - CHILDREN_TO_DISPLAY;
+  return (
+    <ul className="no-bullet">
+      {childTaxons.map((child) => (
+        <li key={child.taxonId}>
+          <Link to={getEntryPath(Namespace.taxonomy, child.taxonId)}>
+            {child.commonName || child.scientificName || child.taxonId}
+          </Link>
+        </li>
+      ))}
+      <li>
+        <Button
+          to={{
+            pathname: LocationToPath[Location.TaxonomyResults],
+            search: `query=parent:${taxonId}`,
+          }}
+          element={Link}
+          variant="tertiary"
+          className={styles['no-margin']}
+        >
+          {remainingChildren > 0
+            ? `${remainingChildren} more direct ${pluralise(
+                'child',
+                remainingChildren,
+                'children'
+              )}`
+            : 'Browse all direct children'}
+        </Button>
+      </li>
+      <li>
+        <Button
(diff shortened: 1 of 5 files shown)


src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+          to={{
+            pathname: LocationToPath[Location.TaxonomyResults],
+            search: `query=parent:${taxonId}`,
+          }}
+          element={Link}
+          variant="tertiary"
+          className={styles['no-margin']}
+        >
+          {remainingChildren > 0
+            ? `${remainingChildren} more direct ${pluralise(
```
dlrice, line 49: 1. I like the word `browse` you've used in

- "Browse all direct children"
- "Browse all descendants"

because it lets me know that I'm going to be taken to a new page of results when I click the button. What do you think about also having it for the case of `remainingChilren > 0` too? Eg

`22 more direct children`  --> `Browse 22 more direct children`

2. What do you think about just showing `Browse all direct children (27)` (eg not subtracting what is visible from the count) as in a sense we are browsing all of the direct children (and not just those that aren't currently visible).

src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <ul className="no-bullet">
+      {childTaxons.map((child) => (
+        <li key={child.taxonId}>
+          <Link to={getEntryPath(Namespace.taxonomy, child.taxonId)}>
+            {child.commonName || child.scientificName || child.taxonId}
+          </Link>
+        </li>
+      ))}
+      <li>
+        <Button
```
dlrice, line 39: Could this just be a `Link`?

src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
```diff
@@ -0,0 +1,73 @@
...
+          to={{
+            pathname: LocationToPath[Location.TaxonomyResults],
+            search: `query=ancestor:${taxonId}`,
+          }}
+          element={Link}
+          variant="tertiary"
+        >
+          Browse all descendants
```
dlrice, line 66: We've changed the wording from the current site (`All lower taxonomy nodes (547,824)`). Would be worth sanity checking with Marija, if she hasn't already.

src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
aurel-l: I like option 2. Less logic, less cognitive overhead for the user (now it would be "so... 22 + the ones I see, let's count...").
I will do that

src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
aurel-l: It's to get the styles of the button to reproduce the style of the ExpandableList pattern. Somehow a Button (even if it's actually a link styled with the Button's classes) would match the pattern's size whereas just a plain link wouldn't.
I'll retry that now in case I find a cleaner approach today with a fresh mind

src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
aurel-l: She was OK with the wording when I showed her yesterday

src/supporting-data/taxonomy/components/entry/ChildNavigation.tsx
aurel-l: OK, so I had a look and I think I prefer calling the franklin Button component even if it's just in order to inject the classes (`button` and `tertiary`) instead of injecting the classes myself, because the component import is explicit whereas using global classes will make any refactor in the future more complicated.
Also, in the end it does render just an `a` tag so semantically the output is fine.

dlrice (approved):

PR #603, opened 2022-04-26
2022 06 card table selector

aurel-l:
## Purpose
Result view selector https://www.ebi.ac.uk/panda/jira/browse/TRM-27567

## Approach
Present a custom popup with a choice for either cards or table as custom results view

## Testing
Manual check on private browser + removing corresponding entry in localStorage on normal instance of browser

## Checklist

Tried in a previous approach to use `@tippyjs/react` as in franklin but ran into issues with react state and events inside of the tooltip (main blocker) and convoluted logic and props in order to keep the tooltip open until a selection was made


aurel-l:
I realise that I did a change in the error component that ended up in there. It is supposed to (try to) prevent displaying an error message when we will reload the page (avoids flashings, and reporting to Sentry). Sorry about mixing it in, I thought I had pushed it before in a different commit

dlrice (approved):
Looks good 👍. I pushed a few commits with trivial changes.

I noticed the blur doesn't stretch over the whole app (the `main-header` can still be interacted with) as the mock-up has but I don't think that's a problem.

aurel-l:
> I noticed the blur doesn't stretch over the whole app (the main-header can still be interacted with) as the mock-up has but I don't think that's a problem.

Yes, when playing with the z-indices I noticed this happening and thought that it might be a good way to still let users escape if they don't want to make a choice, or if for whatever reason something buggy happens they could still reach the contact form

PR #608, opened 2022-05-04
Deploy warning

dlrice:
Note that if this is merged it will cause beta to have the dev banner.

## Purpose

1. Allow us to deploy only prod or development if needed
2. Show a dismissible dev banner for non-www.uniprot.org deployments

## Approach

1. Create some `if-elif-else` logic in gitlabci.
2. Create a `DeploymentWarning` component.

## Testing
None

## Checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -103,9 +103,14 @@ deploy:
   script:
     - git clone $DEPLOY_REPO_URL
     - >
-      if [ "$CI_COMMIT_BRANCH" == "main" ]; then
-        ./${DEPLOY_REPO}/scripts/deploy.sh HX uniprot-website-fallback
-        ./${DEPLOY_REPO}/scripts/deploy.sh HH uniprot-website-prod
+      DEPLOY=./${DEPLOY_REPO}/scripts/deploy.sh
+      if [ "$DEPLOY_FALLBACK_ONLY" = true ]; then
+        $DEPLOY HX uniprot-website-fallback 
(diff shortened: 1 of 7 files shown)


aurel-l (approved):
Looks good. I've just updated the component in order to fix types.

Wondering if something like this would have been enough https://github.com/ebi-uniprot/uniprot-website/blob/c6b75d0dbe6f6086ce45a65ff8efc491eb14b414/src/uniprotkb/components/entry/Entry.tsx#L189-L213 but I don't think that would saved much code anyway compared to your approach and we wouldn't be able to put it where you put it.

A reminder to **only merge this after the release**

dlrice:
> I've just updated the component in order to fix types.

Didn't see an issue with the types - was this showing up on your machine? Or this a best practices thing?

aurel-l:
> > I've just updated the component in order to fix types.
> 
> Didn't see an issue with the types - was this showing up on your machine? Or this a best practices thing?

I saw this happening on GitLab https://gitlab.ebi.ac.uk/uniprot/front-end/uniprot-website/-/jobs/834965

dlrice:
> I saw this happening on GitLab https://gitlab.ebi.ac.uk/uniprot/front-end/uniprot-website/-/jobs/834965

Ah I see, that's for a previous commit and fixed in a later commit in [this job](https://gitlab.ebi.ac.uk/uniprot/front-end/uniprot-website/-/jobs/834992). We can leave as is - both versions do the same thing. I just wasn't sure if there's something I was missing.

PR #612, opened 2022-05-13
ErrorComponent in results page

aurel-l:
## Purpose
add error and status return values from usePagination in order to show an error when the API returns an error

## Approach
usePagination was not exposing any error or status, so expose them in order to be able to use them to show the appropriate error message.

This will show the error message if there is an error while loading the 1st page.

Note that this will surface the error and status of the current page being loaded, but the added condition of checking if there are results in the array should prevent replacing the already loaded data in the event of one of the subsequence page load failing. At the moment this is not handled I believe, and I'm assuming it would either show the Loader component forever or just stop adding to the list as if all results were shown. @dlrice if you want to have a go at this specific case feel free to do so, I just have to stop now and won't have more time to work on it.

## Testing
Manual test during a API load test => show 503 error message on full page

## Checklist

### DIFF
--- a/src/shared/components/results/Results.tsx
+++ b/src/shared/components/results/Results.tsx
@@ -12,6 +12,7 @@ import ResultsData from './ResultsData';
 import ResultsFacets from './ResultsFacets';
 import SideBarLayout from '../layouts/SideBarLayout';
 import NoResultsPage from '../error-pages/NoResultsPage';
+import ErrorHandler from '../error-pages/ErrorHandler';
 import ErrorBoundary from '../error-component/ErrorBoundary';
 import ResultsDataHeader from './ResultsDataHeader';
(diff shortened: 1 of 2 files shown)


dlrice (approved):
Looks good. I'll try to have a look at the case of displaying a message in the case of subsequent pages not loading

aurel-l:
> Looks good. I'll try to have a look at the case of displaying a message in the case of subsequent pages not loading

Just thought about it now, but one way to reproduce could be blocking in the devtools page number 3 (for example) of a search with a lot of results. Can't remember off the top of my head what error/status you would get though, but at least you'd be able to reproduce a bit more easily and consistently

aurel-l:
Since you approved it I'll merge it now, as multiple users reported seeing the wrong error message these days (no results instead of server error)

PR #615, opened 2022-05-30
Missing data

Swaathik:
## Purpose
- Missing fields in proteome entry such as strain and the replacement proteome info for redundant proteome are added.
- The keywords field in diseases are exposed too.
- The download format for Redundant proteomes is set to only FASTA like in the current website.

## Approach
Added an optional property 'supportedFileFormats' in Download component to allow exceptions.

## Testing
Below tests were changed to reflect the latest changes
- src/proteomes/components/entry/__tests__/Components.spec.tsx
- src/proteomes/components/entry/__tests__/ComponentsButtons.spec.tsx

## Checklist

### DIFF
--- a/src/proteomes/components/entry/Components.tsx
+++ b/src/proteomes/components/entry/Components.tsx
@@ -22,8 +22,8 @@ const genomeAccessionDB = 'GenomeAccession' as const;
 const getIdKey = ({ name }: Component) => name;
 
 export const Components: FC<
-  Pick<ProteomesAPIModel, 'components' | 'id' | 'proteinCount'>
-> = ({ components, id, proteinCount }) => {
+  Pick<ProteomesAPIModel, 'components' | 'id' | 'proteinCount' | 'proteomeType'>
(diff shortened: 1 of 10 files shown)


aurel-l (approved):
All good in this PR, but I just did a few minor changes in the latest commit:
 - Order of imports. We try to group imports to keep all related imports together. You'll see as you go the different grouping. Mainly it's external libraries first, uniprot-website codebase middle, types after that, and non-standard (css, files, etc) last, and within groups we also separate into sub-groups according to if they are react components, custom hooks, helper functions, etc. As much as we can, because sometimes we have multiple types of import from a single file. This is mainly useful for huge files with a lot of imports, but we try to apply it everywhere.
 - Changed the span with `data-article-id` to cover the whole "redundant" word, in order to make the whole word clickable
 - Simplified the logic in the strain renderer

Feel free to merge (and delete the branch after you've merged), remember to resolve the corresponding jiras after the change is live on beta

PR #620, opened 2022-06-16
Disease xrefs

dlrice:
## Purpose
[Cross-ref links for Diseases](https://www.ebi.ac.uk/panda/jira/browse/TRM-25838)
This PR fixes the issue of diseases namespace entries having missing database links.

## Approach

- Downloaded all 6,127 diseases entries to determine there are three dbs:
```
{'MIM', 'MeSH', 'MedGen'}
```
Only `MIM` is in the uniprotkb db config so the others needed to be created.
- Hard coded the db configuration by looking at the legacy links (note that I deviated as it seems the legacy site has broken MeSH links). Created [ticket](https://www.ebi.ac.uk/panda/jira/browse/TRM-28081) for API to create on their end.
- Used db config in the column configuration
- Use `logging.warn` to capture when a db doesn't exist in the config.

I spent some time to trying to use the existing xref functionality in UniProtKB but in the end I decided I didn't want to create any changes there as it's all pretty precarious and UniProtKB-centric.

## Testing
Updated snapshots

## Checklist

### DIFF
--- a/src/supporting-data/diseases/config/DiseasesColumnConfiguration.tsx
+++ b/src/supporting-data/diseases/config/DiseasesColumnConfiguration.tsx
@@ -1,8 +1,12 @@
 import { Link } from 'react-router-dom';
-import { ExpandableList } from 'franklin-sites';
+import { ExpandableList, ExternalLink } from 'franklin-sites';
 
 import { getEntryPathFor } from '../../../app/config/urls';
 import { mapToLinks } from '../../../shared/components/MapTo';
+import { processUrlTemplate } from '../../../uniprotkb/components/protein-data-views/XRefView';
+import * as logging from '../../../shared/utils/logging';
+
+import databaseToDatabaseInfo from './databaseInfoMaps';
 
 import { DiseasesAPIModel } from '../adapters/diseasesConverter';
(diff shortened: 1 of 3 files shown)


src/supporting-data/diseases/config/DiseasesColumnConfiguration.tsx
```diff
@@ -58,24 +62,41 @@ DiseasesColumnConfiguration.set(DiseasesColumn.alternativeNames, {
     ) : null,
 });
 
-// NOTE: should probably be links
 DiseasesColumnConfiguration.set(DiseasesColumn.crossReferences, {
   label: 'Cross references',
-  // TODO: https://www.ebi.ac.uk/panda/jira/browse/TRM-25838
   render: ({ crossReferences }) =>
-    crossReferences?.length ? (
+    Boolean(crossReferences?.length) && (
```
aurel-l, line 68: I believe ExpandableList now doesn't render anything if it is passed either undefined, or an empty list, so I would say that this check was not even needed in the first place, and you'd be able to just remove it altogether

aurel-l (approved):

PR #621, opened 2022-06-17
2022 08 blast align

aurel-l:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-27768 & https://www.ebi.ac.uk/panda/jira/browse/TRM-27732 Add ability to align from BLAST results with or without including the initial query sequence

## Approach
- Replace tools button with links
- In the case of BLAST results, use a dropdown for the Align buttons
- Adjust a bit styling/alignment in Franklin (as much as I could without messing up everything, it's a pain)
- Adjust the way the buttons were enabled/disabled depending on the [2, 100] sequences limits taking into account the extra sequence in the new button
- Allow having both "sequence" and "ids" passed to tool forms and concatenate both

## Testing
Updated snapshots + manual tests

## Checklist


aurel-l:
- Make sure to update Franklin to test (otherwise the buttons/links will jump around when enabled/disabled).
- BLAST can be incredibly slow at the moment, so feel free to use ncbiblast-R20220613-081722-0494-44230251-p1m for testing

src/shared/components/action-buttons/Align.tsx
```diff
@@ -1,39 +1,95 @@
...
-    }
+  if (sequence) {
+    return (
+      <Dropdown
+        visibleElement={
+          <Button
+            // If both buttons within were to be disabled
+            disabled={isDisabled(n) && isDisabled(n + 1)}
+            variant="tertiary"
+            title={`Select from 2 to ${ALIGN_LIMIT} entries to run an Align job`}
```
dlrice, line 55:
```suggestion
            title={`Select from 1 to ${ALIGN_LIMIT} entries to run an Align job`}
```
As we can align against the query

dlrice (approved):

PR #627, opened 2022-07-01
Kinetics

Swaathik:
## Purpose
A table view is implemented for Kinetics section in the Entry page. 

## Approach
Protvista-datatable has been used. 
The following are the few use cases considered for RegEx match
- For KM, the text inside the paranthesis should go to Notes column, but sometimes there could be abbreviations which are supposed to be together in the Substrate column.
- There will be notes that should be shown below the tables, they are added to the additional Notes variable and shown as <TextView> just like before.
- For Kcat, values that are not parsed are added to the above mentioned Notes section.

## Testing
https://github.com/ebi-uniprot/uniprot-website/blob/kinetics/src/uniprotkb/components/entry/__tests__/KineticsTableView.spec.tsx
Mock data: https://github.com/ebi-uniprot/uniprot-website/blob/kinetics/src/uniprotkb/components/entry/__tests__/__mocks__/kineticsData.ts

More examples in [here](https://www.uniprot.org/uniprotkb?query=(cc_bpcp_kinetics:NaCl))
Prototype: https://www.figma.com/file/haXmHfuHVWIVlPREZ27Cr3/Kinetics?node-id=0%3A1 

## Checklist


aurel-l (approved):
I did a few changes, mainly regarding reusing the code from other parts of the codebase, and putting the snapshots in their own files. Apart from that, all good to go for me.

PR #630, opened 2022-07-11
Idmapping warnings errors

dlrice:
## Purpose
Show [ID mapping errors and warnings](https://github.com/ebi-uniprot/uniprot-manual/blob/main/help/id_mapping.md#warnings-and-errors-for-results) returned from the API [[jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-27261)]

## Approach
- Combined all errors to get a unique array. Likewise with the warnings.
- Log unrecognized errors & warnings.
- For errors show a full page error with the previously unused `JobErrorPage`.
- For warnings, show a banner at the top.
- Tried to use fallback for the server but the redirect link (and also pagination) was broken in the job details. Leaving the yarn script there as other things worked.
- Disable customisation when results aren't enriched.

## To view

20 (too many IDs for facets):
http://localhost:8080/id-mapping/uniprotkb/7ea4029de730379bc7689e503b78588b518a5687/overview

21 (too many IDs for enrichment):
http://localhost:8080/id-mapping/uniprotkb/fac1cd9456b80a46e05508e3893ffd04d36e1b83/overview

40 (too many IDs for ID mapping):
http://localhost:8080/id-mapping/uniprotkb/a61c3610ce47e067cfdacda2f39e2544ff89e02b/overview

Note a61c3610ce47e067cfdacda2f39e2544ff89e02b doesn't render to column. [Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-28193) made.

## Testing
- Updated tests
- Extracted a fn `renderIDMappingResult`
- Updated out of date urls in the other id mapping sections

## Checklist


src/tools/id-mapping/components/results/__mocks__/tooManyIDsForMappingDetails.ts
```diff
@@ -0,0 +1,12 @@
+export default {
+  from: 'UniProtKB_AC-ID',
+  to: 'GI_number',
+  ids: 'A0A024B7W1,A0A024RBG1,and many more',
+  errors: [
+    {
+      code: 40,
+      message:
+        'Id Mapping API is not supported for mapping results with "mapped to" IDs more than 500000',
```
aurel-l, line 9:
```suggestion
        'ID Mapping API is not supported for mapping results with "mapped to" IDs more than 500 000',
```
I would maybe suggest a phrasing that might make more sense to users like "ID Mapping API is not supported when there are more than 500 000 results". Additionally we could suggest to reduce the input size

aurel-l: That makes me think, we could also provide an automatic batching of queries in the website. We have an input limit, but what we could provide is a batch option in the form, automatically going higher if the user goes over the limit, and from this error message we could generate a link to the input form pre-filled with a higher number in order to split the search even more

aurel-l:
Regarding
> 21 (too many IDs for enrichment)
I would expect the table to default to a simple from/to with links to the entry or external links if possible. At the moment I see an empty table

aurel-l:
Regarding
> 20 (too many IDs for facets)

Would there be a way to put this message where the facets would have been? Because this would make it more obvious what is unavailble if we were to reuse the same space.

src/tools/id-mapping/components/results/__mocks__/tooManyIDsForMappingDetails.ts
dlrice: That's a nice idea. I'll create a few jiras:

1. Ask backend to update their returned error message as you've indicated
2. Create a batch id mapping submission workflow in the website

dlrice:
> Regarding
> 
> > 21 (too many IDs for enrichment)
> 
> I would expect the table to default to a simple from/to with links to the entry or external links if possible. At the moment I see an empty table

Is this for the example I linked above? I wonder if this is related to https://www.ebi.ac.uk/panda/jira/browse/TRM-28193 ? What do you see in a private browsing window where there won't be column preferences

dlrice:
> Regarding
> 
> > 20 (too many IDs for facets)
> 
> Would there be a way to put this message where the facets would have been? Because this would make it more obvious what is unavailble if we were to reuse the same space.

Considered this but Marija preferred not too as it may appear squashed and we couldn't recall an instance of where we places messages in the side panel

src/tools/id-mapping/components/results/__mocks__/tooManyIDsForMappingDetails.ts
aurel-l: 1. Ah, I didn't realise that this message was coming directly from the backend message. Then that shouldn't block this PR I guess
2. Yes, that could be for both. Straight from the input length limit, we could submit multiple jobs, and in the case of an issue of too many results, then do the iterative approach of suggesting to resubmit through a link with a higher number of jobs. If additionally to the "too many results" we could actually know how many results there are we could actually know how many jobs to submit

aurel-l:
> > Regarding
> > > 21 (too many IDs for enrichment)
> > 
> > 
> > I would expect the table to default to a simple from/to with links to the entry or external links if possible. At the moment I see an empty table
> 
> Is this for the example I linked above? I wonder if this is related to https://www.ebi.ac.uk/panda/jira/browse/TRM-28193 ? What do you see in a private browsing window where there won't be column preferences

Yep, that is with the job you linked, this is what I see even with a private window

Screenshot 2022-07-14 at 17 18 53

dlrice:
> Yep, that is with the job you linked, this is what I see even with a private window. And I'm not sure if it is related to the jira you linked

I think this comes down to a mismatch between data expectations between front end and API. Shall we leave this task as part of the new jira I created or would you prefer to resolve in this PR?

aurel-l (approved):

PR #632, opened 2022-07-13
2022 09 lineage

aurel-l:
## Purpose
Add lineage IDs column renderer https://www.ebi.ac.uk/panda/jira/browse/TRM-28178 + refactor lineage loading in UniProtKB entry page in order to render links

## Approach
- Added renderer for lineage IDs
- Reuse as much as possible existing taxonomy components
- Create a new one that automatically loads lineage from taxonomy endpoint, falls back to list of strings before it has data
- Lazy-"activate" this new component to only load data when in view

## Testing
Updated snapshots

## Checklist


src/uniprotkb/components/protein-data-views/EMBLView.tsx
```diff
@@ -11,7 +11,22 @@ import * as logging from '../../../shared/utils/logging';
 import { Xref } from '../../../shared/types/apiModel';
 import { PropertyKey } from '../../types/modelTypes';
 
-import EMBLXrefProperties from '../../config/emblXrefPropertiesData.json';
+const EMBLXrefProperties: Record<string, string> = {
```
Swaathik, line 14: Everything looks fine. I couldn't connect this one. Is this related to lineage? (out of my ignorance)

aurel-l: Ah, sorry no, not related. It was to remove a small imported file because that way it also gave more information to TS about its content (see removed `as` below).
But that got into this change and I forgot to put it somewhere else.

PR #634, opened 2022-07-18
hide all references to tools and basket in mobile view

aurel-l:
## Purpose
disable all tools and basket references in mobile view https://www.ebi.ac.uk/panda/jira/browse/TRM-28221

## Approach
Use of media query, as defined in Franklin, in 2 ways:
- Use directly in CSS through mixin
- Use in JavaScript
Create abstraction for both ways and try to reused wherever possible

## Testing
Manual testing on narrow screen + updated snapshots

## Checklist

### DIFF
--- a/src/app/components/home-page/AnalysisTools.tsx
+++ b/src/app/components/home-page/AnalysisTools.tsx
@@ -22,7 +22,8 @@ const AnalysisTools = () => (
       'uniprot-grid',
       'uniprot-grid--centered',
       'uniprot-grid--with-bleed',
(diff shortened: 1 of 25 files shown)


aurel-l:
> It is not related to Tools. Did we decide to allow entry selection in History for comparison?

I wasn't really sure what we had decided either, even though I do remember we had a look at it. I think it's still usable on mobile so I'd leave it like this, but I'll make sure to let Marija know in the Jira so that she can decide if that still needs to be removed even if it is removed later.

PR #636, opened 2022-07-19
03 release PR checklist

aurel-l:
## Checklist
- [x] Remove some legacy messages:
  - Removed in contextual help
- [x] Update score annotation handling after backend change https://www.ebi.ac.uk/panda/jira/browse/TRM-28219
- [x] Update/create release note article and link in home page
- [x] Schema and UI update regarding binding sites after backend changes https://www.ebi.ac.uk/panda/jira/browse/TRM-28199
- [x] Schema (and maybe UI) update regarding UniRule and ARBA after backend changes https://www.ebi.ac.uk/panda/jira/browse/TRM-28239 -> Might not be breaking, just missing data, so if this is the case, not a blocker then


aurel-l:
@dlrice @Swaathik SIB told me that the article for the new release has been created already (https://github.com/ebi-uniprot/uniprot-manual/blob/main/release-notes/2022-08-03-release.md) so point 3 of this draft PR can be prepared

PR #637, opened 2022-07-19
GO slimsets selection 

Swaathik:
## Purpose
The default slim set for the GO ribbon needs to be set depending on the taxonomy/lineage of the protein.

## Approach
The following grouping has been added to check if the taxonomy lineage matches any of them. 
```
Bacteria (2) -> prokaryotes
Archaea (2157) -> prokaryotes
Viridiplantae (33090) -> plants
Bangiophyceae (2797) -> plants
Florideophyceae (2806) -> plants
Stylonematophyceae (446134) -> plants
Rhodellophyceae (661027) -> plants
Compsopogonophyceae (930023) -> plants
anything else -> generic
```
If there is a match, the slimset is selected by default.
## Testing
Manually tested for entries that are associated with plants

## Checklist


src/uniprotkb/components/entry/GoRibbon.tsx
```diff
@@ -75,6 +75,38 @@ const GoRibbon = ({
...
+      Object.entries(slimSetByTaxon).forEach(([key, value]) => {
+        const presentTaxon = taxonomyInfo?.filter(
+          (t) => value.includes(String(t)) // Lineage is Array of strings here
+        );
+        if (presentTaxon?.length) {
+          setSelectedSet(key);
+        }
+      });
+    }
+  }, [organismData]);
```
aurel-l, line 108: This function is not really an effect, it's more of an initial state, so I would pass this function as a state initialiser to the useState on line 68, returning the found key, or if none is found `'goslim_generic'`

Swaathik: :woman-facepalming: 
I did the change

src/uniprotkb/components/entry/GoRibbon.tsx
```diff
@@ -65,7 +65,39 @@ const GoRibbon = ({
...
+      const slimSetByTaxon = {
+        goslim_plant: [
+          'Viridiplantae',
+          'Bangiophyceae',
+          'Florideophyceae',
+          'Stylonematophyceae',
+          'Rhodellophyceae',
+          'Compsopogonophyceae',
+        ],
+        prokaryotes: ['Bacteria', 'Archaea'],
```
aurel-l, line 86: So, one thing that came out of the conversation with SIB was to use a prokaryote slimming set, but the issue is that we don't have such slimming set available 🤷🏽 
It's good that this logic is here, but it should probably be commented out for now because it renders "agr" in the select element and doesn't render any visualisation.
We'll have further discussions regarding this to see what we can and cannot do, the point for now is to have this PR's logic in place
```suggestion
        // prokaryotes: ['Bacteria', 'Archaea'],
```

Swaathik: Ah, yes it takes 'agr', I didn't notice that. I will comment it for now

PR #639, opened 2022-07-20
2022 09 add basket to subsets

Swaathik:
## Purpose
Add basket button for features spanning >1 amino acid along with the existing BLAST button

## Approach
The modifications subset is stores as a string in the Basket.
A Map of subset to accession is passed to `ResultsButton` to deal with exceptions like ID mapping and Download functions.

## Testing
Additional test added to AccessionView to see how the ids with ranges are shown
Other snapshots updated with the Add Basket button

## Checklist

### DIFF
--- a/src/basket/BasketMiniView.tsx
+++ b/src/basket/BasketMiniView.tsx
@@ -62,16 +62,15 @@ export const updateResultsWithAccessionSubsets = (
   results: APIModel[],
   namespace: Namespace,
   accessions: string[]
-) => {
-  return results.map((r, index) => {
+) =>
+  results.map((r, index) => {
     if (namespace === Namespace.uniprotkb) {
-      const entry: UniProtkbAPIModel = { ...r };
+      const entry = { ...r } as UniProtkbAPIModel;
       entry.primaryAccession = accessions[index];
       return entry;
     }
     return r;
   });
-};
 
 const BasketMiniViewTab = ({
   accessions,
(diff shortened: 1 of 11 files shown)


src/basket/BasketMiniView.tsx
```diff
@@ -54,13 +58,34 @@ type BasketMiniViewTabProps = {
...
 };
 
+export const updateResultsWithAccessionSubsets = (
+  results: APIModel[],
+  namespace: Namespace,
+  accessions: string[]
+) => {
+  return results.map((r, index) => {
+    if (namespace === Namespace.uniprotkb) {
+      const entry: UniProtkbAPIModel = { ...r };
```
Swaathik, line 68: There is a Type error I am unable to solve. Please could you look at it?

aurel-l: In this case you need to cast the object to the right model in order to narrow down from the too generic APIModel to the correct UniProtkbAPIModel.
You could have narrowed down the type by checking if the object contained fields specific to UniProtKB, or here since you know and check that the namespace is uniprotkb then I'd say it's OK to cast the object.

Swaathik: I get it, thank you.

Swaathik:
There was an undefined error thrown and the 'Remove' looks okay now

aurel-l (approved):

aurel-l:
I've noticed that submitting an Align or BLAST from the basket mini view triggers an warning about "ids" not being an accepted column. This appears to happen even before this PR though, so nothing in here caused it, but if you feel like having a look at it please go ahead.

aurel-l:
Identified through user reports that there is an issue with the basket showing wrongly associated accessions and entry data (probably a mix-up of the result of the accessions endpoint). Will revert this PR for now and have a look next week.

PR #645, opened 2022-08-05
Ligand description view

dlrice:
## Purpose
Same ligand view was appearing in both uniprotkb and unirule which can be unified.

## Approach
Create single view

## Testing
Updated

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/LigandDescriptionView.tsx
+++ b/src/uniprotkb/components/protein-data-views/LigandDescriptionView.tsx
@@ -1,0 +1,86 @@
+import { ReactNode } from 'react';
+import { Link } from 'react-router-dom';
+
+import { RichText } from './FreeTextView';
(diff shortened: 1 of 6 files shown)


src/uniprotkb/components/protein-data-views/LigandDescriptionView.tsx
```diff
@@ -0,0 +1,86 @@
...
+            }}
+          >
+            UniProtKB
+          </Link>
+          {' | '}
+          <ExternalLink url={externalUrls.ChEBI(id)}>ChEBI</ExternalLink>
+          {') '}
+        </>
+      )}
+      <RichText>{ligand.note}</RichText>
```
Swaathik, line 53: Should the note here be preceded with a semicolon? as the note is similar to the description. Example: UniRule UR000177404

Swaathik (approved):

PR #646, opened 2022-08-08
xrefs: gencc, clingen, refseq, ncbi nucleotide

dlrice:
## Purpose
Fix a few xref issues:

1. GenCC & ClinGen must have as many links as there are HGNC IDs
2. GenCC & ClinGen must have HGNC IDs for URL text
3. GenCC & ClinGen must be located in Disease & Variants entry page section (as well as all xrefs tab)
Jiras for above three points:
- https://www.ebi.ac.uk/panda/jira/browse/TRM-27783
- https://www.ebi.ac.uk/panda/jira/browse/TRM-28048

4. Fix RefSeq link (https://www.ebi.ac.uk/panda/jira/browse/TRM-28297)
5. Update proteomes link to NCBI nucleotides (https://www.ebi.ac.uk/panda/jira/browse/TRM-28298)

## Approach

1. Add exception to creation of implicit xref to explicitly check if GenCC & ClinGen and if so add an xref for of the HGNC IDs
2. Add exception to xref view render to handle GenCC & ClinGen case
3. Add GenCC & ClinGen to Disease & Variants section in xref config
4. Add NucleotideSequenceId to allowed propertyKeySet
5. Remove ENABrowser and add NCBI nucleotides

## Testing
Updated tests. Specifically note that GenAtlas no longer has `properties.id` but this has actually never been needed (only gene name).

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
+++ b/src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
@@ -7659,7 +7659,7 @@ exports[`Entry view should render for non-human entry 1`] = `
                     </a>
                     <a
                       class="external-link"
-                      href="https://www.ncbi.nlm.nih.gov/nuccore/%necleotideId"
+                      href="https://www.ncbi.nlm.nih.gov/nuccore/NZ_WKPS01000046.1"
                       rel="noopener"
                       target="_blank"
                     >
--- a/src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
+++ b/src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
@@ -7657,7 +7657,18 @@ exports[`Entry view should render for non-human entry 1`] = `
                         width="12.5"
                       />
                     </a>
-                    NZ_WKPS01000046.1
+                    <a
+                      class="external-link"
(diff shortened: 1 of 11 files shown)


src/uniprotkb/components/entry/__tests__/__snapshots__/EntryMain.spec.tsx.snap
```diff
@@ -7657,7 +7657,18 @@ exports[`Entry view should render for non-human entry 1`] = `
                         width="12.5"
                       />
                     </a>
-                    NZ_WKPS01000046.1
+                    <a
+                      class="external-link"
+                      href="https://www.ncbi.nlm.nih.gov/nuccore/%necleotideId"
```
aurel-l, line 7662: Looks like this pattern hasn't been replaced by a valid URL

dlrice: Good spot! Needed the database config file to be updated. Would be nice to devise a means of updating all mocks at once rather than copy and paste. Not sure how to do this given we also want TS applied to the mocks.

aurel-l: Indeed, we don't want to rely on network everytime we do a test to pull the latest value.
Instead, eventually, we might want to have a script that we can call regularly (whenever backend tells us there was a change) that would generate those files that could be commited, like snapshots of the backend data, but used in our side a mocks for our own snapshots.

aurel-l (approved):

PR #647, opened 2022-08-10
fix: ID mapping results table not rendering lineage column

dlrice:
## Purpose
Fix this bug: ID mapping results table not rendering lineage column (https://www.ebi.ac.uk/panda/jira/browse/TRM-28307)

## Approach
Use useColumnNames in ID mapping results to be used in creation of results url

## Testing
None

## Checklist


Swaathik (approved):

aurel-l (approved):
Works fine, and that has also a great side effect of not just fixing the issue at hand, but it also reduces the network load by actually just fetching the data needed for the columns displayed, which can be a great improvement for UniProtKB data!

PR #652, opened 2022-08-19
2022 10 mobile view click targets and homepage

aurel-l:
## Purpose
Mobile view for homepage https://www.ebi.ac.uk/panda/jira/browse/TRM-28225
+ fix too small click targets reported by Google Search Console

## Approach
- Use new classes in franklin to define different grid layout depending on breakpoints
  - Mainly: when 4 tiles, switch to 2 tiles per row on 2 rows
  - Add "in-between" breakpoint for when the middle size is too tight but we don't want to trigger mobile view yet
- Refactor home page content to have a better reflow when going in mobile view
- Create new specific rules for mobile view
- Check links that were reported by Google Search Console as too small or touch target overlapping with others and adjust both size (padding, for bigger click target) and spacing between them (to avoid clicking on the wrong one with touch)
  - Played with values until validated by local lighthouse audit (audit: Snapshot / Mobile / SEO)

## Testing

## Checklist

### DIFF
--- a/src/app/components/home-page/LatestNews.tsx
+++ b/src/app/components/home-page/LatestNews.tsx
@@ -66,7 +66,9 @@ const LatestNews = () => {
     >
       <div
         className={cn(
-          'uniprot-grid-cell--span-4',
+          'uniprot-grid-cell--small-span-12',
+          'uniprot-grid-cell--medium-span-4',
+          'uniprot-grid-cell--medium-offset-9',
           styles['latest-news__news-roll']
         )}
       >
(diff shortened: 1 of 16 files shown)


src/app/components/home-page/LatestNews.tsx
```diff
@@ -215,35 +217,14 @@ const LatestNews = () => {
               </p>
             </article>
           </li>
-          <li>
```
dlrice, line 218: Rather than adding/removing html to update what do you think about making a json object to hold the data and this populates a component?

dlrice (approved):

src/app/components/home-page/LatestNews.tsx
aurel-l: We'll eventually (some time before next release hopefully) do https://www.ebi.ac.uk/panda/jira/browse/TRM-28342 so I guess this will be tackled then and will be coming from an object coming from the network

PR #653, opened 2022-08-25
Ptm table features

dlrice:
## Purpose
Display PTMeXchange data in feature table [jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-28341)

## Approach

- Request data from proteins proteomics-ptm endpoint within the PTM/Processing section
- Convert this data into features to be combined with the uniprot features which are finally provided to `UniProtKBFeaturesView`
- Conditionally add Source column to features table
- Pass any PTMeXchange data to `PtmExchangeEvidenceTag`

Other
- Placed sources in an `ExpandlableList` which will show 10 initial items.
- Render pubmed/europepmc link when uniprot/citations doesn't return anything.
- Sorted a few dictionaries/types. Not really necessary but thought it could be helpful for humans to search over it in the future.
- Changed `FeatureData` to `FeatureDatum[]`. Didn't realise I could have just done `FeatureData[0]` until afterwards. If you prefer the previous way I will revert.

## Testing
- Added snapshot tests

## Checklist


aurel-l:
> * Changed `FeatureData` to `FeatureDatum[]`. Didn't realise I could have just done `FeatureData[0]` until afterwards. If you prefer the previous way I will revert.

I do prefer to define single types and use them with the `Datum[]` approach even though I might not have done that all the time. So all good for me with your change. Note that I've recently discovered you could use `Data[number]` instead of `Data[0]`, I feel like it makes it nicer if we do have to extract a single type from an array.

src/uniprotkb/components/entry/ProteinProcessingSection.tsx
```diff
@@ -24,7 +29,18 @@ const ProteinProcessingSection = ({
...
+    return <Loader />;
+  }
+
+  const convertedPtmExchangeFeatures = proteomicsPtmData
+    ? convertPtmExchangeFeatures(proteomicsPtmData.features)
+    : [];
+
+  if (!hasContent(data) && !convertedPtmExchangeFeatures.length) {
     return null;
   }
```
aurel-l, line 45: Is this what triggers the side-panel to decide to render the shortcut to the section? If this is it it might be good to keep that in mind to have the same approach for the structure and the variation sections (not part of this PR though)

aurel-l: Note that the variations endpoint can be heavy, and this one might also be heavy too in the future. But seeing that these endpoints return 404 when no data, a quick and cheap way to check might be to do a "HEAD" query and see the status code, and do the actual data loading within a lazy component (like for the variations endpoint)

src/uniprotkb/components/protein-data-views/PtmExchangeEvidenceTag.tsx
```diff
@@ -0,0 +1,130 @@
...
+        )
+      )}
+      {publicationReferences && (
+        // Remove this className - using as a stopgap until data is imported into citations
+        <div className={style.publications}>
+          <UniProtKBEntryPublications
+            pubmedIds={
+              publicationReferences
+                .map((reference: Evidence) => reference.id)
+                .filter((id?: string) => id) as string[]
```
aurel-l, line 60:
```suggestion
                .filter((id?: string): id is string => Boolean(id))
```

aurel-l (approved):

src/help/config/featureTypeHelpMappings.ts
```diff
@@ -29,6 +29,7 @@ const FeatureTypeHelpMappings: Omit<Record<FeatureType, string>, 'Other'> = {
   Chain: 'chain',
   Peptide: 'peptide',
   'Modified residue': 'mod_res',
+  'Modified residue (large scale)': 'mod_res_large_scale', // TODO: confirm help page
```
aurel-l, line 32:
```suggestion
  'Modified residue (large scale)': 'mod_res_large_scale',
```

aurel-l (approved):

PR #656, opened 2022-09-07
Ptmexchange confidence score

dlrice:
# To be merged when the PTMeXchange confidence score help is final and published.

## Purpose
Expose the confidence score now made available in the PTMeXchange data from the proteomics-ptm proteins endpoint.

## Approach

- Collect confidence scores for each PTM (these should all be the same)
- If they're not the same or there aren't any confidence scores log, this to be picked up by Sentry and render an `Unknown` score.
- A few other type changes as the API changed their data shape but we don't rely on these

## Testing
Update snapshots

## Checklist


src/uniprotkb/adapters/ptmExchangeFeaturesConverter.ts
```diff
@@ -32,6 +35,24 @@ const convertPtmExchangePtms = (
...
+    new Set(
+      ptms.flatMap(({ dbReferences }) =>
+        dbReferences.map(({ properties }) => properties['Confidence score'])
+      )
+    )
+  );
+  let confidenceScore: ConfidenceScore | undefined;
+  if (!confidenceScores.length) {
+    logging.error('PTMeXchange PTM has no confidence score');
+  } else if (confidenceScores.length > 1) {
```
aurel-l, line 48:
```suggestion
  const confidenceScores = new Set(
    ptms.flatMap(({ dbReferences }) =>
      dbReferences.map(({ properties }) => properties['Confidence score'])
    )
  );
  let confidenceScore: ConfidenceScore | undefined;
  if (!confidenceScores.size) {
    logging.error('PTMeXchange PTM has no confidence score');
  } else if (confidenceScores.size > 1) {
```
Probably micro-optimisation, but I guess the extra array is not really needed

aurel-l: Oh, that would maybe cause an issue with the stringification in the log.error for multiple values. I would say do an Array.from() there as it's not the branch of code that is expected to run the most

dlrice: I see what you mean, I did go back and forth on this a few times when I implemented it. We need to get the only element from `confidenceScores` (a few lines below) so for a set I would need to do `confidenceScores.values().next().value` which felt a bit strange (but valid). Overall, not sure which is preferable with regards to performance and readability?

dlrice: Actually didn't realise I can use destructuring assignment. I'll adapt to use a set then.

aurel-l (approved):

PR #658, opened 2022-09-14
2022 11 datatable expansion

Swaathik:
## Purpose
- Expand/Collapse button added to Protvista-datatable.
- Add 'Full screen view' button to features table to take the user to Feature viewer tab. 

## Approach
Using the expandTable property from Protvista-datatable.

## Testing

## Checklist


src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -1,15 +1,20 @@
...
 
+import { gtagFn } from '../../utils/logging';
+
 import './styles/datatable-toggle.scss';
 
-const DatatableWithToggle = ({ table }: { table: JSX.Element }) => {
-  const [tableState, setTableState] = useState('Expand');
-  const tableRef = useRef(null);
-  const tableContainerRef = useRef<HTMLDivElement>(null);
+const DatatableWithToggle: FC = ({ children }) => {
```
aurel-l, line 11: A slight API change to this component, putting the table as a child of this component, because it is effectively just wrapping it

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -1,15 +1,20 @@
...
+  const [expandTable, setExpandTable] = useState(false);
```
aurel-l, line 12: Changed to match the naming of the attribute of the datatable element

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -20,29 +25,37 @@ const DatatableWithToggle = ({ table }: { table: JSX.Element }) => {
...
-      if (tableState === 'Expand') {
-        tableEl.expandTable = true;
-        setTableState('Collapse');
-      } else {
-        tableEl.expandTable = false;
-        setTableState('Expand');
-        await frame();
-        tableContainerRef?.current?.scrollIntoView({
+  // On expand/collapse change
+  useEffect(() => {
```
aurel-l, line 29: Also, since the collapsed state is a react state, we can put it as a dependency of a useEffect, meaning that it will be called after the change has been applied to the DOM by React, removing the hack I had previously suggested involving waiting

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -20,29 +25,37 @@ const DatatableWithToggle = ({ table }: { table: JSX.Element }) => {
...
+    // except on first render
+    if (!firstRenderRef.current) {
+      // Scroll table back into view when collapsing
+      if (!expandTable) {
+        tableRef.current?.parentElement?.scrollIntoView({
           behavior: 'smooth',
           block: 'center',
         });
       }
+      gtagFn('event', 'feature viewer', {
```
aurel-l, line 39: Took the opportunity to add some tracking as it is a behaviour we might be interested to see in the analytics numbers

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -20,29 +25,37 @@ const DatatableWithToggle = ({ table }: { table: JSX.Element }) => {
...
+        event_category: `datatable ${expandTable ? 'expanded' : 'collapsed'}`,
+        event_label: params.accession,
+      });
     }
-  };
+    // If first render was previous render, then it's not the first anymore...
+    if (firstRenderRef.current) {
+      firstRenderRef.current = false;
+    }
```
aurel-l, line 47: Just a flag needed to avoid triggering the scroll logic on the first render

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -20,29 +25,37 @@ const DatatableWithToggle = ({ table }: { table: JSX.Element }) => {
...
   if (datatableElement.defined) {
     return (
-      <div ref={tableContainerRef} className="datatable-with-toggle">
-        <datatableElement.name ref={tableRef} filter-scroll>
-          {table}
+      <div className="datatable-with-toggle">
+        <datatableElement.name
+          ref={tableRef}
+          filter-scroll
+          expandtable={expandTable ? '' : undefined}
```
aurel-l, line 56: Since React doesn't support web components too well, if I were to use the boolean it would be rendered as a string as `expandtable="true"` or `expandtable="false"` (truthy anyway), but in our custom element logic we tend to follow the usual pattern of "attribute present means true" vs "attribute absent means false". So here I used "undefined" to tell React to remove the attribute

src/shared/components/views/VisualFeaturesView.tsx
```diff
@@ -5,13 +5,16 @@ import { Link, useParams } from 'react-router-dom';
...
 import useCustomElement from '../../hooks/useCustomElement';
+
 import { getEntryPath } from '../../../app/config/urls';
+import { gtagFn } from '../../utils/logging';
+
+import { TabLocation } from '../../../uniprotkb/components/entry/Entry';
 import { Namespace } from '../../types/namespaces';
 
-import './styles/visual-features-view.scss';
+import styles from './styles/visual-features-view.module.scss';
```
aurel-l, line 17: We tend to use CSS modules wherever we can (especially in new styles or when touching old ones) to avoid involuntary global styles / leaks to other components

src/shared/components/views/VisualFeaturesView.tsx
```diff
@@ -70,9 +73,16 @@ function VisualFeaturesView<T>({
           params.accession,
           TabLocation.FeatureViewer
         )}
-        className="full-view"
+        title="View in the Feature Viewer"
```
aurel-l, line 76: Accessibility issue when using an icon for a link, the browser doesn't know what it means and can't expose the information to screen readers. Here putting a title, it will be exposed to those tools.

aurel-l (approved):

src/shared/components/views/DatatableWithToggle.tsx
Swaathik: Great, it makes sense

PR #662, opened 2022-10-21
Implement new ProteinOverview for UniProtKB entries

aurel-l:
## Purpose
Implement new ProteinOverview for UniProtKB entries. https://www.ebi.ac.uk/panda/jira/browse/TRM-28536

## Approach
Modify the existing component (also used in cards) to add a variant that will render in UniProtKB entry pages according to the mockup
Also, add links to help in labels (making https://www.ebi.ac.uk/panda/jira/browse/TRM-27821 obsolete) and change contextual help styling to avoid layout shifts

## Testing

## Checklist


src/app/components/styles/app.scss
```diff
@@ -23,6 +23,27 @@
   }
 }
 
+/* contextual help */
+/* global style now to avoid layout shift when loading logic later */
+[data-article-id] {
+  &::after {
+    content: 'i';
+    cursor: wait;
```
Swaathik, line 31: I wasn't able to see this cursor (nor the layout shift)

Swaathik (approved):

src/app/components/styles/app.scss
aurel-l: The layout shift would have appeared without this change, only on first load of the page (the time it takes for the contextual help logic to load). It would push the text in the info list to the left when the "i" appears.
To see the opacity: 0.5 / cursor: wait in action, you'd probably need to simulate 3g network in the devtools and reload an entry page. It's still quite quick to load though in machines like ours.

aurel-l:
Thanks for the key warning fix!

PR #663, opened 2022-10-21
2022 13 url link update

Swaathik:
## Purpose
Update rule and signature links for HAMAP and Prosite in UniRule pages. (https://www.ebi.ac.uk/panda/jira/browse/TRM-28320)
Also, Pfam links are updated.

## Approach

## Testing

## Checklist

### DIFF
--- a/src/shared/config/externalUrls.ts
+++ b/src/shared/config/externalUrls.ts
@@ -11,7 +11,8 @@ const externalUrls: Record<string, (id: string | number) => string> = {
   ENA: (id) => `//www.ebi.ac.uk/ena/data/view/${id}`,
   // protein centric
   InterPro: (id) => `https://www.ebi.ac.uk/interpro/protein/${id}`,
-  Pfam: (id) => `https://www.ebi.ac.uk/interpro/protein/uniprot/${id}`,
+  Pfam: (id) =>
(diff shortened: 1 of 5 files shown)


src/shared/config/externalUrls.ts
```diff
@@ -11,10 +11,13 @@ const externalUrls: Record<string, (id: string | number) => string> = {
   ENA: (id) => `//www.ebi.ac.uk/ena/data/view/${id}`,
   // protein centric
   InterPro: (id) => `https://www.ebi.ac.uk/interpro/protein/${id}`,
-  Pfam: (id) => `http://pfam.xfam.org/protein/${id}`,
+  Pfam: (id) => `https://www.ebi.ac.uk/interpro/protein/uniprot/${id}`,
```
aurel-l, line 14: Thinking that maybe, at least for now, it should point to `https://www.ebi.ac.uk/interpro/protein/UniProt/${id}/entry/pfam/` maybe? Otherwise it might not be clear to people why they end up on the InterPro page for the protein if they don't immediately see Pfam data in the case it might be drowned within the other members databases.
What do you think?

aurel-l (approved):

src/shared/config/externalUrls.ts
Swaathik: I agree, it is more of an InterPro landing. I will update it

aurel-l (approved):

PR #669, opened 2022-11-18
2022 14 db snp column

Swaathik:
## Purpose
Display dbSNP cross references data(from FeatureCrossReferences unlike other xrefs). [JIRA](https://www.ebi.ac.uk/panda/jira/browse/TRM-28089)

## Approach
Information displayed are as follows:

- Variant change with link to UniProt variant page
- rsID
- Links to dbSNP and Ensembl

Description is there in the data but left out as it started to look cluttered. I will attach screenshot for reference.
Example in Legacy for one such with extra description - https://legacy.uniprot.org/uniprot/?query=database%3A%28type%3Adbsnp%29+AND+A0A1B0GTW7&sort=score

Links to dbSNP and UniProt variant pages are added in Disease and variants section following Elisabeth's comment in the JIRA.(http://localhost:8080/uniprotkb/P78363/entry#disease_variants)

## Testing
Updated the snapshots

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/DiseaseInvolvementView.tsx
+++ b/src/uniprotkb/components/protein-data-views/DiseaseInvolvementView.tsx
@@ -17,6 +17,21 @@ import { FeatureDatum } from './UniProtKBFeaturesView';
 import styles from './styles/variation-view.module.scss';
 import externalUrls from '../../../shared/config/externalUrls';
 
+export const uniprotVariantLink = (variant: FeatureDatum) => {
+  return variant.alternativeSequence?.originalSequence ||
+    variant.alternativeSequence?.alternativeSequences?.[0] ? (
+    <ExternalLink url={externalUrls.UniProt(variant.featureId || '')} noIcon>
+      {variant.alternativeSequence?.originalSequence || <em>missing</em>}
+      {'>'}
+      {variant.alternativeSequence?.alternativeSequences?.[0] || (
+        <em>missing</em>
+      )}
+    </ExternalLink>
+  ) : (
+    <em>missing</em>
+  );
+};
+
 export const DiseaseVariants = ({ variants }: { variants: FeatureDatum[] }) => {
   const table = (
     <table>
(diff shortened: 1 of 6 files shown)


src/uniprotkb/components/protein-data-views/DiseaseInvolvementView.tsx
```diff
@@ -15,6 +15,21 @@ import { Namespace } from '../../../shared/types/namespaces';
 import { FeatureDatum } from './UniProtKBFeaturesView';
 
 import styles from './styles/variation-view.module.scss';
+import externalUrls from '../../../shared/config/externalUrls';
```
aurel-l, line 18: Just a small note on the order of imports, this should be higher

src/uniprotkb/components/protein-data-views/DiseaseInvolvementView.tsx
```diff
@@ -15,6 +15,21 @@ import { Namespace } from '../../../shared/types/namespaces';
...
+    <ExternalLink url={externalUrls.UniProt(variant.featureId || '')} noIcon>
+      {variant.alternativeSequence?.originalSequence || <em>missing</em>}
+      {'>'}
+      {variant.alternativeSequence?.alternativeSequences?.[0] || (
+        <em>missing</em>
+      )}
+    </ExternalLink>
+  ) : (
+    <em>missing</em>
+  );
```
aurel-l, line 32: I think all of them should be wrapped with the external link component, even if they are "missing". In the page example you gave, the first displayed variant had "missing" written but still had a featureId in the data, so it should still link the the corresponding page.

Swaathik: But it takes to the expasy home page. I assume that doesn't exist. 
this has a change - -https://web.expasy.org/variant_pages/VAR_084833.html
while this has 'missing' - https://web.expasy.org/variant_pages/VAR_012494.html

Swaathik: I checked how it is done in legacy. It is not linked there either

aurel-l: Ah, right. I just saw that the ID is there in the data, but didn't realise the corresponding link in the expasy site was leading nowhere.
In which case everything is fine 👌

aurel-l (approved):

PR #671, opened 2022-12-08
Refactor evidence codes 

Swaathik:
## Purpose
Some ECO codes are interpreted differently. 
https://www.ebi.ac.uk/panda/jira/browse/TRM-27017
(https://www.ebi.ac.uk/panda/jira/browse/TRM-26978?focusedCommentId=376847&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-376847
Our code has to be updated to be in line with the above conditions

## Approach
Evidence's label is changed to a function from string, as it allows for conditional checks.

## Testing

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/PtmExchangeEvidenceTag.tsx
+++ b/src/uniprotkb/components/protein-data-views/PtmExchangeEvidenceTag.tsx
@@ -93,7 +93,7 @@ const PtmExchangeEvidenceTag = ({
     <EvidenceTag
       label={
         originalEvidenceData.labelRender?.(originalEvidences) ||
-        originalEvidenceData.label
+        originalEvidenceData.label(originalEvidences)
       }
       className={
         originalEvidenceData.manual
@@ -103,7 +103,7 @@ const PtmExchangeEvidenceTag = ({
     >
       <section className={style['evidence-tag-content']}>
         <h5 data-article-id={`evidences#${originalEvidenceCode}`}>
-          {originalEvidenceData.label}{' '}
+          {originalEvidenceData.label(originalEvidences)}{' '}
           <small>({originalEvidenceData.description})</small>
         </h5>
         <section>
--- a/src/uniprotkb/components/protein-data-views/PtmExchangeEvidenceTag.tsx
+++ b/src/uniprotkb/components/protein-data-views/PtmExchangeEvidenceTag.tsx
@@ -92,8 +92,8 @@ const PtmExchangeEvidenceTag = ({
   return (
     <EvidenceTag
       label={
-        originalEvidenceData.labelRender?.(originalEvidences) ||
-        originalEvidenceData.label(originalEvidences)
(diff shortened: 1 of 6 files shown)


src/uniprotkb/config/evidenceCodes.ts
```diff
@@ -91,126 +93,129 @@ const manualLabelRenderer = (evidences: Evidence[]) => {
 export const ecoCodeToData = {
   [ecoCode.EXP]: {
     manual: true,
-    label: 'Manual assertion based on experiment',
+    label: () => 'Manual assertion based on experiment',
     description: 'Inferred from experiment',
     labelRender: publicationCountRenderer,
```
aurel-l, line 98: Feel like there's a naming issue here between these 3 fields that we'll stumble on in the future.
Should we find a way to rename this throughout the codebase to something clearer? I understand that we use some of them as fallback for the other in some places, so it might still not make it obvious.
I can see the `labelRender` is is used for the evidence tag label, and `label` for the evidence tag content heading. Should we call them like that? (`evidenceTagLabel` & `evidenceTagContentHeading`)
And `description` is used for the GO section, right? should that be `evidenceTagContentHeadingForGO`

src/uniprotkb/config/evidenceCodes.ts
```diff
@@ -219,55 +224,58 @@ export const ecoCodeToData = {
...
     manual: false,
-    label: 'Automatic assertion inferred from database entries',
+    label: () => 'Automatic assertion inferred from database entries',
     description: 'Automatically imported',
     labelRender: () => labels.IMPORTED,
   },
   [ecoCode.AA]: {
     manual: false,
-    label: 'Automatic assertion according to rules',
+    label: (evidences: Evidence[]) =>
```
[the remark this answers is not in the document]
aurel-l: I guess it is because of the exception that was mentioned yesterday to reuse the same ECO code for something different. Acknowledging this expection, is there still an issue with this section of the code?
Maybe leaving a comment about the exception and why would be good, for us in the future to know that this is something done on purpose

src/uniprotkb/components/protein-data-views/PtmExchangeEvidenceTag.tsx
```diff
@@ -103,7 +103,7 @@ const PtmExchangeEvidenceTag = ({
     >
       <section className={style['evidence-tag-content']}>
         <h5 data-article-id={`evidences#${originalEvidenceCode}`}>
-          {originalEvidenceData.label}{' '}
+          {originalEvidenceData.label(originalEvidences)}{' '}
           <small>({originalEvidenceData.description})</small>
```
aurel-l, line 107: Is there still a case where we render both? I thought we now never had the case of a heading saying something with small text saying something similar

Swaathik: I missed to remove this somehow. Right we don't show both of them anywhere

aurel-l (approved):

PR #674, opened 2022-12-20
2022 16 protein names

Swaathik:
## Purpose
Visual improvement on protein name section. JIRA - https://www.ebi.ac.uk/panda/jira/browse/TRM-28855

## Approach
Link the name to the corresponding row in PTM table and text formatting for 'Alternative names' from bold to italic to make it subtle.

## Testing

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/ProteinNamesView.tsx
+++ b/src/uniprotkb/components/protein-data-views/ProteinNamesView.tsx
@@ -1,5 +1,5 @@
 import { Fragment } from 'react';
-import { Link, useLocation } from 'react-router-dom';
+import { Link } from 'react-router-dom';
 import { InfoList, ExpandableList } from 'franklin-sites';
 
 import ExternalLink from '../../../shared/components/ExternalLink';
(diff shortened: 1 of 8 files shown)


src/uniprotkb/components/protein-data-views/ProteinNamesView.tsx
```diff
@@ -16,9 +16,34 @@ import {
...
+  withLink,
+}: {
+  value: string;
+  withLink: boolean;
+}) => {
+  const location = useLocation();
+  return (
+    <>
+      {withLink ? (
+        <Link to={`${location.pathname}#${stringToID(value)}`}>{value}</Link>
```
Swaathik, line 30: There is a custom lint warning about not building the link path from scratch. I searched but couldn't find the necessary util.

aurel-l: It's fine to ignore this if you're reusing an existing location object. The point was to avoid crafting URLs from scratch when I created this custom rule.
That being said, I did change a bit the code in order to reuse the location that the Link's `to` prop provides already instead of pulling it from the library's custom hook

aurel-l (approved):

PR #675, opened 2023-01-12
add map to feature to UniRef entries + change ID mapping form

aurel-l:
## Purpose
Add the UniRef map to feature using ID Mapping [TRM-28956](https://www.ebi.ac.uk/panda/jira/browse/TRM-28956)

## Approach
Refactor and reuse the "MapTo" dropdown
Prepopulate an ID mapping form with the corresponding data and the correct from/to fields
Fix and prepopulate the job name field
Refactor tools forms for consistency and for the above name prepopulation fix

## Testing
Manual testing + added unit test for snapshots of links being generated

## Checklist

### DIFF
--- a/src/shared/components/MapTo.tsx
+++ b/src/shared/components/MapTo.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactNode } from 'react';
+import { ReactNode } from 'react';
 import { Link, LinkProps, useRouteMatch } from 'react-router-dom';
 import { Button, LongNumber, Dropdown } from 'franklin-sites';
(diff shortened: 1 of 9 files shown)


dlrice (approved):

dlrice:
Not really part of this PR but had the thought during the review. I looked at the case when mapping more than 100,000 in the ID mapping form and noticed our notification is at the info level:
Screenshot 2023-01-13 at 08 14 23
What do you think about changing this to the warning level? Or some other means of informing the user more explicitly that they're over the limit?

PR #676, opened 2023-01-25
layout refactoring

aurel-l:
## Purpose
Refactor the different layouts used in the website in order to solve multiple visual issues when scrolling or when content is too big. https://www.ebi.ac.uk/panda/jira/browse/TRM-28232 & https://www.ebi.ac.uk/panda/jira/browse/TRM-25824

## Approach
- Used CSS grids more, fiddled with CSS rules until everything worked fine...
- Removed unneeded "title" section from layout as it was just complicating it
- Added wide mode for tabs that do not have a sidebar in order to better use the space in UniProtKB entry feature viewer & History, and UniParc feature viewer. Haven't used it in Tools results (input/raw output/api) as the content is usually quite narrow and doesn't stretch well
- Added a variant to the SidebarLayout to allow or disallow horizontal scrolling
- Tried to use/convert as much as possible CSS modules, so that it's more explicit which styles are used where
- Wasn't in the backlog, but added an extra message at the bottom of search results (if you reach it)
  - Uses existing DidYouMean component
  - Nicer visual separation between end of results and footer
  - If the list of results is short backend provides suggestions, so also shows that now

## Testing
- Updating snapshots & visual checks throughout all the different types of pages in the website
  - Not that the comparison of big snapshots might be complicated because the layout changed affected everything within it...

## Checklist


src/help/components/entry/__tests__/Entry.spec.tsx
```diff
@@ -15,6 +15,11 @@ const mock = new MockAdapter(axios);
 
 mock.onGet(/api\/help\/canonical_and_isoforms/).reply(200, mockData);
 
+jest.mock('../../../../shared/components/layouts/UniProtFooter', () => ({
+  __esModule: true,
+  default: () => '{{ Footer }}',
+}));
+
```
aurel-l, line 22: Mock footer to avoid filling up the snapshot with the footer's content which is already tested somewhere else so would mainly be noise for the test here.

webpack.config.js
```diff
@@ -168,9 +168,9 @@ module.exports = (env, argv) => {
               options: {
                 modules: {
                   auto: true, // only for files containing ".module." in name
-                  // class name to hash, but also keep name in development
+                  // class name to hash, but also keep name for debug in dev
                   localIdentName: `${
-                    isDev ? '[local]#' : ''
+                    isDev ? '[local]🐛' : ''
```
aurel-l, line 173: Using the exported name in `src/shared/components/BackToTheTop.tsx`, then used within a querySelector, so having a `#` messed it up. It's only for dev so I used a character that would not be clashing with querySelector and would still be quite unique, so a bug, for debugging...

src/shared/components/entry/styles/entry-page.scss
```diff
@@ -3,11 +3,28 @@
 
 .entry-page {
   *[id] {
-    scroll-margin-top: calc(var(--top-header-margin) + #{$global-margin * 0.5});
+    scroll-margin-top: 4.2em;
```
aurel-l, line 6: Manually crafted 🤷🏽

src/shared/components/layouts/styles/sidebar-layout.module.scss
```diff
@@ -0,0 +1,63 @@
...
+  }
+  @include fs-breakpoints('large') {
+    --sidebar-size: 30ch;
+  }
+
+  max-width: 100%;
+  min-height: calc(100vh - var(--top-header-margin));
+  display: grid;
+  align-items: start;
+  grid-template-columns: 3rem calc(var(--sidebar-size) - 3rem) minmax(0, 1fr);
```
aurel-l, line 20: Uses 3 columns (2 for sidebar, 1 for main content), even though visually there are 2, in order to possibly have in the future the main content cover the sidebar partially if we want to have a triggered "expanded" view. Not so much of a problem for results pages because we have the scroll, but for entries where we force no scroll then it might get useful

src/shared/components/results/__tests__/Results.spec.tsx
```diff
@@ -4,9 +4,20 @@ import Results from '../Results';
...
+jest.mock('../SearchSuggestions', () => ({
+  __esModule: true,
+  default: () => '{{ SearchSuggestions }}',
+}));
+
+jest.mock('../DidYouMean', () => ({
+  __esModule: true,
+  default: () => '{{ DidYouMean }}',
+}));
+
```
aurel-l, line 20: Same thing as before when mocking components, but here we don't have snapshots at the moment, but at least we don't trigger extra related logic that is already tested somewhere else => quicker test here, less things to mock, and less things that might break

src/shared/components/views/VisualFeaturesView.tsx
```diff
@@ -22,10 +22,12 @@ function VisualFeaturesView<T>({
   features,
   sequence,
   trackHeight,
+  noLinkToFullView,
 }: {
   features: T[];
   sequence: string;
   trackHeight?: number;
+  noLinkToFullView?: boolean;
```
aurel-l, line 30: Added this because we had links in UniParc pages that were taking nowhere

src/shared/styles/sticky.module.scss
```diff
@@ -2,10 +2,16 @@
 
 .sticky-tabs-container {
   & :global(.tabs__header) {
-    z-index: 100;
+    z-index: 5000;
     position: sticky;
     top: 0;
-    background: $colour-sky-white;
+    background: rgba($colour-sky-white, 0.75);
+    backdrop-filter: blur(2px);
```
aurel-l, line 9: Thought it looked cool, but can definitely be removed if we don't like it. It gives the glossy effect on the tabs within the entry pages

aurel-l:
@dlrice 's comments addressed

PR #679, opened 2023-02-06
Column select refactor

dlrice:
## Depends on https://github.com/ebi-uniprot/franklin-sites/pull/253

Will require linking or waiting for franklin's npm package to be updated.

## Purpose
- [Modify evidence behaviour for advanced search](https://www.ebi.ac.uk/panda/jira/browse/TRM-28359)
- [Improve accordion search performance](https://www.ebi.ac.uk/panda/jira/browse/TRM-23184)

## Approach
- Adapted data for franklin's refactor accordion search
- No longer immediately update local storage upon user single column selection to improve performance.
- Modified panel closing behavior after discussion with Minjoon:
```
click x → just cancel
click outside of the window → save and display new columns
"save" (current close) → save and display new columns
click reset to default → reset but not close the window
```
- Log how user exits customise table with google events
- Incorporated a suggestion by Minjoon: as a new column is added the drag and drop div is scrolled to the right to make the addition clear to the user.

## Testing
- Updated

## Checklist


src/shared/components/action-buttons/CustomiseButton.tsx
```diff
@@ -16,8 +23,64 @@ const CustomiseTable = lazy(
...
 
+/*
+| Reason  | User clicked               |
+|---------|----------------------------|
+| outside | outside the panel          |
+| button  | top right close (x) button |
+| cancel  | cancel                     |
+| submit  | save                       |
+*/
+type Reason = 'outside' | 'button' | 'cancel' | 'submit';
```
aurel-l, line 34:
```suggestion
type Reason = Parameters<Exclude<ComponentProps<typeof SlidingPanel>['onClose'], undefined>>[0] | 'cancel';
```

I wish we had a better way to type this (like exporting more types from franklin) but for now I guess it allows us to make sure types don't diverge and we are not expecting a string that might be changed in the future in franklin

src/shared/components/action-buttons/CustomiseButton.tsx
```diff
@@ -1,9 +1,16 @@
-import { Suspense, useState } from 'react';
+import { FormEvent, Suspense, useState } from 'react';
```
aurel-l, line 1:
```suggestion
import { FormEvent, Suspense, useState, ComponentProps } from 'react';
```
Goes alongside the suggestion below

src/shared/components/column-select/ColumnSelectDragDrop.tsx
```diff
@@ -20,55 +20,69 @@ const ColumnSelectDragDrop: FC<ColumnSelectDragDropProps> = ({
...
-                >
-                  <Chip
-                    disabled={false}
-                    onRemove={() => onRemove(itemId)}
-                    className={cn({
-                      'chip--dragging': snapshot.isDragging,
-                    })}
+    }
+    previousColumns.current = columns;
+  }, [columns]);
```
aurel-l, line 34: Do you mind just commenting what this is accomplishing and the cases when this is important?

aurel-l (approved):
All good, it would just be good to have a bit of a comment in your own words in a section that might not be obvious at first what it does.

PR #688, opened 2023-03-09
app.scss update for 'i' button fix on h3 headers

minjoonkim:
make h1-6 headers all inline-blocks

## Purpose
contextual help `i` buttons on `h3` tags are currently clickable across the width of the website

## Approach
This fix makes sure that the `i` buttons are only clickable on itself

## Testing

## Checklist

### DIFF
--- a/src/app/components/styles/app.scss
+++ b/src/app/components/styles/app.scss
@@ -81,6 +81,12 @@ protvista-manager {
   width: 100%;
 }
 
+@for $i from 1 through 6 {
+  h#{$i} {
+    display: inline-block;
+  }
+}
+
 // Debug utility to show heading levels
 // @for $i from 1 through 6 {
 //   h#{$i} {


aurel-l:
> 



> I just noticed when browsing around with this change applied that some headers on the Home page for example where having issues (following text appearing on the same line as the header).
> 
> I'd suggest restricting this change either to headers within entry pages (all entry pages have the class `.entry-page` so that could be one way to target them) or only the headers that have contextual help annotations (`[data-article-id]` is a selector that could target them).

Note that I see the same issue in the sequence section of the entry
Screenshot 2023-03-09 at 16 46 50
So I guess the 2n suggestion might be best for the CSS selector using the data attribute might be best

aurel-l (approved):

aurel-l:
By the way, now that this has been accepted and that the tests are still passing feel free to merge the pull request yourself and delete the branch.

PR #692, opened 2023-03-29
Search suggestions

Swaathik:
## Purpose
Add search suggestions for search terms that has exact macthes like Gene and E.C. number.

The above as part of this JIRA - https://www.ebi.ac.uk/panda/jira/browse/TRM-29051

## Approach
Parse and modify the clause and pass the query int he same original order

## Testing

## Checklist

### DIFF
--- a/src/shared/components/results/SearchSuggestions.tsx
+++ b/src/shared/components/results/SearchSuggestions.tsx
@@ -1,11 +1,31 @@
+import { Link } from 'react-router-dom';
+
 import AdvancedSearchSuggestion from './AdvancedSearchSuggestion';
 import ExactFieldSuggestion from './ExactFieldSuggestion';
+import TaxonomyLevelsSuggestion from './TaxonomyLevelsSuggestion';
 
 import { Namespace } from '../../types/namespaces';
 
 const simpleQuery = /^[a-zA-Z0-9]+$/;
 
 export const exactMatchSearchTerms = ['gene', 'ec'];
+export const taxonHierarchySearchTerms = ['taxonomy_id', 'organism_id'];
+
+export const SearchTextLink = ({
+  query,
+  text,
+}: {
+  query: string;
+  text: string;
+}) => (
+  <Link
+    to={() => ({
+      search: `query=${query}`,
+    })}
+  >
+    {text}
+  </Link>
+);
 
 const SearchSuggestions = ({
   query,
@@ -31,6 +51,9 @@ const SearchSuggestions = ({
         {simpleQuery.test(query) && <AdvancedSearchSuggestion query={query} />}
         {exactMatchSearchTerms.some((term) => query.includes(term)) && (
           <ExactFieldSuggestion query={query} />
+        )}
+        {taxonHierarchySearchTerms.some((term) => query.includes(term)) && (
+          <TaxonomyLevelsSuggestion query={query} />
         )}
       </>
     );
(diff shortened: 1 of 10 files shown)


aurel-l (commented):
This look good and does what it's supposed to do 👍🏽 , although I think it might become more and more complicated the more we add logic in there.
Knowing that there is no overlap between the 3 possible things we'd want to display in this section, I'd have a generic component that would basically be a switch depending on what is in the query field and render 3 different components depending on the case. Something like:
```
SearchSuggestion:
  if (simple search) return <AdvancedSearchSuggestion /> // current logic
  if (gene or ec in search) return <ExactFieldSuggestion /> // new logic added in this PR
  if (some kind of taxonomy in search) return <TaxonomyLevelsSuggestion /> // logic to jump through different level
```
It means that you could move all the effects and other hooks related to specific bits inside their own component (e.g. the useLayoutEffect to render more or less advanced search suggestions would go in `AdvancedSearchSuggestion`, just like the useDataApi related to that.

src/shared/config/apiUrls.ts
```diff
@@ -230,7 +231,31 @@ export const getAPIQueryParams = ({
...
+        !query.includes('organism_id') &&
+        facetField.includes(proteomeFacet)
+      ) {
+        const index = facetField.findIndex(
+          (facet) => facet === FacetsEnum.Proteome
+        );
+        facetField.splice(index, 1);
+      }
+    }
   }
```
aurel-l, line 257: Was this done to match the behaviour on legacy?
If I understood that bit of code well it adds the proteomes facet on the side if one searches for an organism, right?
Now that I'm playing with it I don't think we need to match that, because the issue is that it will show multiple proteomes if there are more than 1, and it won't be clear which one is the best one to choose.
With your work on the suggestion sentence at the top I think it's enough

Swaathik:
Yes, to match with legacy. I will remove it then

src/shared/components/results/SearchSuggestions.tsx
```diff
@@ -33,132 +74,29 @@ const SearchSuggestions = ({
...
-  );
-
-  const searchTerms = useMemo(() => {
-    if (!searchTermsData || !data?.matchedFields?.length) {
-      return null;
-    }
-    const flattenedTerms = flatten(searchTermsData);
+    !!query?.length;
 
+  if (validQueryWithContent) {
```
aurel-l, line 86: I know there's currently no overlap in these conditions, so there's no case where 2 of them might be displayed, but thinking in the future where we'd want to suggest more things there might be cases where we want to have an order of priority of suggestions.
I'm suggesting this change:
```ts
if (validQueryContent) {
  if (simpleQuery.test(query)) {
    return <AdvancedSearchSuggestion query={query} />;
  }
  if (exactMatchSearchTerms.some((term) => query.includes(term))) {
    return <ExactFieldSuggestion query={query} />;
  }
  if (taxonHierarchySearchTerms.some((term) => query.includes(term)) {
    return <TaxonomyLevelsSuggestion query={query} />;
  }
  // Add more suggestions in the future here
}
return null;
```

The specific case in my mind (for another jira and PR, not to add here), is for example a query like `(mass:[* TO 50])` we might want to suggest something like "do you want to display the "Mass" column?" with a shortcut to add that column if we see that the search field and the column name match. In the case of `taxonomy` we'd had an overlap and we might not want to add both suggestions as it takes too much space

aurel-l (approved):

aurel-l (approved):

PR #693, opened 2023-03-31
2023 03 variants tab

aurel-l:
## Purpose
Move the variants section to a new tab https://www.ebi.ac.uk/panda/jira/browse/TRM-29227

## Approach
- Create new tab for Variants
- Move current variation viewer to the new tab
- Add an explanation where the viewer was before, with a link
- Conditionally only renders the tab and the link to the tab if there are variants on the Proteins API
- In the results table, use protvista-datatable to have the "click to expand feature" and make sure to have the disease and natural variants work using UniProt data (and not Proteins API data)
- Use the feature viewer in the disease & variants section to display all variants (additionally to mutagenesis data)

## Testing
- Updated snapshots & added a test for no data
- Added date in mocks for disease view to make sure only variants related to the disease are rendered with the disease

## Checklist
  - Removed tested code that was not useful anymore, that lowered the coverage


src/uniprotkb/components/entry/DiseaseAndDrugsSection.tsx
```diff
@@ -94,9 +96,35 @@ const DiseaseAndDrugsSection = ({
         features={data.featuresData}
         sequence={sequence}
       />
-      <LazyComponent fallback="Variants" rootMargin="50px">
-        <VariationView primaryAccession={primaryAccession} title="Variants" />
-      </LazyComponent>
+      {hasImportedVariants ? (
```
dlrice, line 99: Just curious why you opt for `condition ? <foo> : null`  rather than `condition && <foo>`?. I personally find the former a it bit a trickier to comprehend as I need to look for the else on another line.

src/uniprotkb/components/entry/DiseaseAndDrugsSection.tsx
```diff
@@ -94,9 +96,35 @@ const DiseaseAndDrugsSection = ({
...
+                    TabLocation.VariantsViewer
+                  )}
+                >
+                  Go to variants tabs <FullViewIcon width="0.75em" />
+                </Link>
+              </p>
+            </div>
+          </div>
+        </section>
+      ) : null}
```
dlrice, line 127: This results in an empty section:

Screenshot 2023-04-11 at 17 07 55

While we used to display:
```
No variation information available for A0A024QZA8
```

Could we redisplay this or maybe even not render the section at as we have will know based on the previous `HEAD` request?

src/uniprotkb/components/protein-data-views/VariationView.tsx
```diff
@@ -212,12 +204,15 @@ const VariationView = ({
     !filteredVariants
   ) {
     return (
-      <div>
+      <section
+        style={{ marginInlineEnd: '1.2rem' }}
```
dlrice, line 208: This style also listed at line 487 so what do think about defining in a css file?

src/uniprotkb/components/protein-data-views/__tests__/__mocks__/diseaseInvolvementUIData.ts
```diff
@@ -1,85 +1,192 @@
...
-      acronym: 'AD',
-      description: 'Some description',
-      diseaseCrossReference: {
-        database: 'EMBL',
-        id: 'some ref id',
-        properties: {
-          prop1: 'value1',
-        },
-        isoformId: 'some isoform',
+const mock: { comments: DiseaseComment[]; features: FeatureDatum[] } = {
```
dlrice, line 4: Is this from an endpoint? Would it be possible to list the meta info like we do in other mocks eg:

```
// Source: /api/uniprotkb/search?facets=reviewed%2Cmodel_organism%2Cproteins_with%2Cexistence%2Cannotation_score%2Clength&query=glucose&size=25
// Retrieved: 2022-08-02
```

src/uniprotkb/components/protein-data-views/__tests__/__snapshots__/DiseaseInvolvementView.spec.tsx.snap
```diff
@@ -141,19 +141,7 @@ exports[`DiseaseInvolvement should render DiseaseInvolvement 1`] = `
...
-                rel="noopener"
-                target="_blank"
-              >
-                dbSNP:rs13447459
-                <test-file-stub
-                  data-testid="external-link-icon"
-                  width="12.5"
-                />
-              </a>
+              in AD; supposed to be rendered with the disease
```
dlrice, line 144: Just double checking if the link is supposed to be removed here?

src/uniprotkb/components/protein-data-views/__tests__/__snapshots__/DiseaseInvolvementView.spec.tsx.snap
```diff
@@ -176,19 +164,7 @@ exports[`DiseaseInvolvement should render DiseaseInvolvement 1`] = `
...
-                rel="noopener"
-                target="_blank"
-              >
-                dbSNP:rs13447492
-                <test-file-stub
-                  data-testid="external-link-icon"
-                  width="12.5"
-                />
-              </a>
+              in AD
```
dlrice, line 167: Same question here.

src/uniref/components/entry/EntryMain.tsx
```diff
@@ -23,4 +21,4 @@ const EntryMain = ({ transformedData }: EntryMainProps) => (
   </>
 );
 
-export default memo(EntryMain, isSameEntry);
+export default memo(EntryMain);
```
dlrice, line 24: I seem to remember this was important incase one entry linked to another, then memo wouldn't know what has changed. Is this no longer an issue?

src/uniprotkb/components/entry/DiseaseAndDrugsSection.tsx
aurel-l: Probably I initially had a variable with a number instead of a boolean, and wanted to avoid rendering "0". I've changed as suggested

aurel-l:
Somehow I realised that some entries do have the Proteins API variation endpoint return 200 even though they don't have variations, e.g. https://www.ebi.ac.uk/proteins/api/variation/A2JGV3
Not sure why, but not willing to handle this case now. Will wait for the endpoint to return the number of variants through the headers at some point in the future to handle things properly

src/uniprotkb/components/protein-data-views/__tests__/__mocks__/diseaseInvolvementUIData.ts
aurel-l: That's not from an endpoint, I just modified the pre-PR mock (that looks like it also wasn't from an endpoint) in order to cover more edge-cases.

src/uniprotkb/components/protein-data-views/VariationView.tsx
aurel-l: I've now created a class for that

src/uniref/components/entry/EntryMain.tsx
aurel-l: Regarding this if that was an issue in the past I believe it is not anymore, I've tested multiple links through entries and I haven't been able to reproduce the issue.
One thing I could think it would have been useful is if the object passed to the EntryMain component is new, but because it's been created through the data transformers after a re-render, and we would want to detect that the data it refers to is the same by checking if it's the same accession. It feels a bit hacky, and I've checked rerenders without this and it has been nothing crazy.
In this case this bit of code was blocking the needed rerender when we know that there are variants or not, because that's done asynchronously after the first render, so I think it's worth removing it.

src/uniprotkb/components/protein-data-views/__tests__/__snapshots__/DiseaseInvolvementView.spec.tsx.snap
aurel-l: Re-added example data in the mock that triggers rendering a link

PR #696, opened 2023-04-12
UniProtKB landing page

aurel-l:
## Purpose
UniProtKB landing page, using the new statistics endpoint. [TRM-28200](https://www.ebi.ac.uk/panda/jira/browse/TRM-28200)

## Approach

## Testing

## Checklist


Swaathik:
example used for pie chart: https://observablehq.com/@mast4461/d3-donut-chart-labels

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -0,0 +1,241 @@
...
+    .append('text')
+    .attr('dy', '.35em')
+    .style('pointer-events', 'auto')
+    .style('cursor', 'pointer')
+    .style('fill', '#014371')
+    .style('font-weight', 'bold')
+    .on('click', function (d) {
+      history.push({
+        pathname: LocationToPath[Location.UniProtKBResults],
+        search: `query=(taxonomy_name:${d.data.name})`,
```
aurel-l, line 116: Might be good to use Dan's new URLSearchParameters wrapper that would automatically escape whatever needs escaping (I think either the parenthesis or the colon, or both)

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  const unreviewedStats = useDataApi<StatisticsPayload>(
+    releaseNumber &&
+      `${API_PREFIX}/statistics/releases/${releaseNumber}/unreviewed`
+  );
+
+  useEffect(() => {
+    if (
+      svgRef.current &&
+      !reviewedStats.loading &&
```
aurel-l, line 194: Any specific reason you're checking the loading flag only for reviewedStats but not for unreviewedStats?

src/messages/components/__tests__/MessageManagerContainer.spec.tsx
```diff
@@ -58,7 +58,7 @@ describe('Message Manager component', () => {
         format: MessageFormat.IN_PAGE,
         locations: [Location.UniProtKBResults],
       }),
-      route: '/uniprotkb?query=blah',
+      route: '/uniprotkb?=query=blah',
```
aurel-l, line 61: This `?=` feels wrong, could you double check why this is needed, or if it's just a typo that just slipped in without any effect on the test

src/uniprotkb/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+          className={cn(
+            'uniprot-grid-cell--small-span-12',
+            'uniprot-grid-cell--medium-span-9',
+            styles.statistics
+          )}
+        >
+          <h4>Statistics</h4>
+          <div className={styles.statistics__content}>
+            <div className={styles.chart}>
+              <Link
```
aurel-l, line 131: This link is inline, so if the page is too wide it will go on the left of the graph. This should either be placed within a div (that will be block) or be added a class to assign `display: block` to the link itself

aurel-l: Making it a heading could also work, and would be OK for both this and the "Number of entries" text in the next grid cell

src/uniprotkb/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+      (value) => value.value === 'true'
+    )?.count;
+    numberUnreviewed = data?.facets?.[0].values.find(
+      (value) => value.value === 'false'
+    )?.count;
+  }
+
+  return (
+    <div className={styles['landing-page']}>
+      <h2 className={styles['landing-page__title']}>UniProtKB</h2>
```
aurel-l, line 81: This should be a h1 (it's the title of the whole page), and the other headings after it should be h2, and if any heading within these subsections then it should follow and be h3.
If you need to play with the sizes because their default ones are either too big or too small check Franklin's typography file for the classes you can use (I think there are `.huge`, `.big`, etc)

Swaathik: h3 looked very big for the subsections in statistics. I have used 'tiny'. Please have a look

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -0,0 +1,230 @@
...
+    d.startAngle + (d.endAngle - d.startAngle) / 2;
+
+  text
+    .enter()
+    .append('text')
+    .attr('dy', '.35em')
+    .style('pointer-events', 'auto')
+    .style('cursor', 'pointer')
+    .style('fill', '#014371')
+    .style('font-weight', 'bold')
```
aurel-l, line 114: What do you think about setting as much as possible the static things in CSS instead of programatically in javascript? It removes a bit of logic, not sure if that would make such a difference perfomance-wise, but at least there's less stuff in the javascript part so it might be easier to follow and maintain the important logic.
See src/tools/align/components/results/NewickTree.tsx & src/tools/align/components/results/styles/NewickTree.scss as an example

Swaathik (commented):
The taxon link in the pie chart doesn't work in landing page while it is okay in statistics page

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -1,230 +1,101 @@
...
+    if (reviewedStats.loading || unreviewedStats.loading) {
+      return undefined;
+    }
+    const taxonSummed = Object.fromEntries(
+      Array.from(nameToQueryKingdoms.keys(), (name) => [
+        name,
+        {
+          name,
+          entryCount: 0,
+          to: {
```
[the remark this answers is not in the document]
aurel-l: Good catch, there was a typo in the Location object I was generating. Somehow I'm surprised that TS wasn't complaining about it. It should be fixed now

aurel-l:
> The taxon link in the pie chart doesn't work in landing page while it is okay in statistics page

Good catch, there was a typo in the Location object I was generating. Somehow I'm surprised that TS wasn't complaining about it. It should be fixed now

src/uniprotkb/components/landing-page/StatisticsChart.tsx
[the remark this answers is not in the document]
aurel-l: Do you think each slice of the pie should point to the group by page for that taxon?
I was thinking that maybe with the title of the section pointing to the common ancestor of all taxons in the graph that would be good enough.
Here we don't have any number associated with the pie slices, so we can make these links point to anywhere, even simply to the taxonomy page for it. Although I was also thinking that this being the UniProtKB landing page it would be important that most links point to searches within that namespace, and we coud have links to specific taxons whenever we work on a landing page for the taxonomy namespace.

src/uniprotkb/components/statistics/StatisticsPage.tsx
```diff
@@ -0,0 +1,915 @@
...
+                    {percent === '0.00' ? '<0.01' : percent}%
+                  </td>
+                )}
+                {/* Per-entry average */}
+                {/* WRONG: This needs to be divided by number of entries in data release */}
+                {/* {!hasOnlyEntryCounts && (
+                <td className={styles.end}>
+                  {(row.count / row.entryCount).toFixed(2)}
+                </td>
+              )} */}
```
dlrice, line 237: Is this to be fixed as part of this PR?

dlrice (commented):
1. At http://localhost:8080/uniprotkb/statistics#taxonomic-origin > Taxonomic distribution of the sequences across kingdoms and Taxonomic distribution of the sequences within eukaryota it's not immediately obvious to me which column of the table is being represented in the pie chart. 
2. At http://localhost:8080/uniprotkb/statistics#sequence-size > Sequence Count (UniProtKB reviewed) there is a table with the column "Name" which I'm not sure it refers to? Would it be possible to just have a histogram like how the old stats has "Length distribution of the sequences"?
3. When I see a table I want to sort by any of the columns that I see eg Count vs Entry Count
4. At http://localhost:8080/uniprotkb/statistics#journal-citations > Top Journal (UniProtKB reviewed) I'm guessing Count is Citation count? If so could be a little easier to understand if it was explicitly labeled as this. And is the Percent column for Count or Entry count?
5. In the Table of the frequency of occurrence of species for consistency do we also want a column for all of UniProtKB?

Screenshot 2023-08-29 at 16 24 27

6. Referring to the image above, on the landing page my eye is expecting the Explore... texts to line up but they differ slightly vertically.
7. Referring to the same image above, on the landing page could the Number of entries and the corresponding FTP download links be a bit closer? Maybe a table layout would be useful here as it's all related and it could be a bit easier parse?
8. Referring to the same image above, the Statistics header seems to be detached from the numbers - perhaps this should sit above Number of Entries instead?

src/shared/config/ftpUrls.ts
```diff
@@ -8,6 +8,14 @@ const ftpUniProt = 'https://ftp.uniprot.org/pub/databases/uniprot/';
 const ftpUrls = {
   uniprot: ftpUniProt,
   uniprotkb: joinUrl(ftpUniProt, 'knowledgebase/complete'),
+  uniprotkb_reviewed: joinUrl(
```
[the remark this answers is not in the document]
aurel-l: changed

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import useDataApi from '../../../shared/hooks/useDataApi';
+
+import { nameToQueryKingdoms } from '../statistics/taxonomyQueries';
+import { stringifyQuery, stringifyUrl } from '../../../shared/utils/url';
+
+import { LocationToPath, Location } from '../../../app/config/urls';
+import { StatisticsPayload } from '../statistics/StatisticsPage';
+
+import PieChart, { StatisticsGraphItem } from '../graphs/PieChart';
```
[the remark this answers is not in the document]
aurel-l: changed

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const StatisticsChart = ({
+  releaseNumber,
+  reviewed,
+  unreviewed,
+}: StatisticsChartProps) => {
+  const reviewedStats = useDataApi<StatisticsPayload>(
+    releaseNumber &&
+      stringifyUrl(
+        `${API_PREFIX}/statistics/releases/${releaseNumber}/reviewed`,
```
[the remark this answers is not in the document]
aurel-l: changed, I haven't put the statistics category selection in there though, if we start using it more then we could add it in the future

src/uniprotkb/components/statistics/StatisticsPage.tsx
aurel-l: We can address this in the statistics page PR specifically

src/uniprotkb/components/statistics/__tests__/__mocks__/reviewedStats.tsx
```diff
@@ -0,0 +1,5248 @@
+import { StatisticsPayload } from '../../StatisticsPage';
+
+// Retrieved: 2023-04-06
+const reviewedStats: StatisticsPayload = {
```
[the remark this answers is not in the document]
aurel-l: added

src/uniprotkb/components/statistics/__tests__/__mocks__/unreviewedStats.tsx
```diff
@@ -0,0 +1,4756 @@
+import { StatisticsPayload } from '../../StatisticsPage';
+
+// Retrieved: 2023-04-06
```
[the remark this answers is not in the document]
aurel-l: added

src/uniprotkb/components/landing-page/StatisticsChart.tsx
```diff
@@ -0,0 +1,104 @@
...
+    );
+
+    const unreviewedData = unreviewedStats.data?.results.find(
+      (category) => category.categoryName === 'SUPERKINGDOM'
+    );
+
+    for (const [name, entry] of Object.entries(taxonSummed)) {
+      if (reviewed) {
+        entry.entryCount +=
+          reviewedData?.items.find((item) => item.name === name)?.entryCount ||
```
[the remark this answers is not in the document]
aurel-l: changed

aurel-l:
> 1. At http://localhost:8080/uniprotkb/statistics#taxonomic-origin > Taxonomic distribution of the sequences across kingdoms and Taxonomic distribution of the sequences within eukaryota it's not immediately obvious to me which column of the table is being represented in the pie chart.
> 
>     2. At http://localhost:8080/uniprotkb/statistics#sequence-size > Sequence Count (UniProtKB reviewed) there is a table with the column "Name" which I'm not sure it refers to? Would it be possible to just have a histogram like how the old stats has "Length distribution of the sequences"?
> 
>     3. When I see a table I want to sort by any of the columns that I see eg Count vs Entry Count
> 
>     4. At http://localhost:8080/uniprotkb/statistics#journal-citations > Top Journal (UniProtKB reviewed) I'm guessing Count is Citation count? If so could be a little easier to understand if it was explicitly labeled as this. And is the Percent column for Count or Entry count?
> 
>     5. In the Table of the frequency of occurrence of species for consistency do we also want a column for all of UniProtKB?

Should address all of the above in a follow-up PR.
For point 5 frequency I'm not sure we can calculate that because we don't know which species is in which bucket. Anywhere on this page where we have whole UniProtKB data is because we have deduced it from reviewed and unreviewed data, but here it won't be possible.

> Screenshot 2023-08-29 at 16 24 27
> 
>     6. Referring to the image above, on the landing page my eye is expecting the Explore... texts to line up but they differ slightly vertically.

Should be fixed now

>     7. Referring to the same image above, on the landing page could the Number of entries and the corresponding FTP download links be a bit closer? Maybe a table layout would be useful here as it's all related and it could be a bit easier parse?
> 
>     8. Referring to the same image above, the Statistics header seems to be detached from the numbers - perhaps this should sit above Number of Entries instead?

For these we followed the mockup + made it work with the grid layout. Also consider what should happen on narrower screens. Not sure how to fix these points without changing the layout substantially compared to the mockup

dlrice (approved):

PR #697, opened 2023-04-13
Remove sidebar on mobile view

Swaathik:
## Purpose
Remove sidebar for entry pages for small screens

## Approach
Added media query and checked the wider-tab-content to be active only when it is *not* small. 
## Testing

## Checklist

### DIFF
--- a/src/shared/components/layouts/styles/sidebar-layout.module.scss
+++ b/src/shared/components/layouts/styles/sidebar-layout.module.scss
@@ -35,10 +35,6 @@
     &::-webkit-scrollbar {
       display: none;
     }
-    // Hides the facets as they appear until the page content loads
-    @media only screen and (max-width: 640px) {
-      display: none;
-    }
   }
 
   & .empty-sidebar {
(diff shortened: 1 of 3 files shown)


src/shared/components/layouts/styles/sidebar-layout.module.scss
```diff
@@ -35,6 +35,10 @@
     &::-webkit-scrollbar {
       display: none;
     }
+    // Hides the facets as they appear until the page content loads
+    @media only screen and (max-width: 640px) {
+      display: none;
+    }
```
aurel-l, line 41: Similarly, I'd make it use the franklin mixin, although it's defined as min-width so you'd need to have it `display: none` by default, and in the media query it would be `display: initial` or something like this

src/shared/components/layouts/styles/sidebar-layout.module.scss
```diff
@@ -35,6 +35,10 @@
     &::-webkit-scrollbar {
       display: none;
     }
+    // Hides the facets as they appear until the page content loads
+    @media only screen and (max-width: 640px) {
+      display: none;
```
aurel-l, line 40: It's good to use display none in CSS because it means the browser should avoid doing the work of rendering the content, but we could also avoid the rendering in react by conditionally rendering the whole sidebar in SidebarLayout.tsx. You have the `useSmallScreen` custom hook available already and you could use that as a condition to render or not the `<aside>` tag in SidebarLayout

aurel-l (changes requested):

aurel-l (approved):

PR #701, opened 2023-04-28
Refactor state management in tools forms 

Swaathik:
## Purpose
Handle state management through useReducer in all the tools forms following BLAST approach.

## Approach

## Testing

## Checklist


dlrice (commented):
Looks good, just noticed a few bugs (a few of which were from my original blast form reducer) so created a small https://github.com/ebi-uniprot/uniprot-website/pull/703

PR #703, opened 2023-05-03
2023 05 tools refactoring pr

dlrice:
- Noticed a bug with the job name setting that was in my original BLAST form reducer so fixed this along with the other reducers.
- Swapped `parsedSequences`, `peptideSequence` and `formValues[PeptideSearchFields.peps]` to reflect the original logic.

### DIFF
--- a/src/tools/align/state/alignFormReducer.ts
+++ b/src/tools/align/state/alignFormReducer.ts
@@ -22,6 +22,24 @@ const isInvalid = (parsedSequences: SequenceObject[]) =>
   parsedSequences.some((parsedSequence) => !parsedSequence.valid) ||
   parsedSequences.length <= 1;
 
+const getJobName = (parsedSequences: SequenceObject[]) => {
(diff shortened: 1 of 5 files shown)


src/tools/align/state/alignFormReducer.ts
```diff
@@ -70,30 +88,14 @@ export const alignFormParsedSequencesReducer = (
...
-  }
-
   const name =
     formValues[AlignFields.name].userSelected &&
     formValues[AlignFields.name].selected
       ? formValues[AlignFields.name]
       : {
           ...formValues[AlignFields.name],
-          selected: potentialJobName,
+          userSelected: false,
```
Swaathik, line 97: This is necessary in the following scenario:
1. User selected a name
2. Then reset it
3. then provide sequence info
4. Name is not populated as userSelected is 'true' at that point and name is empty.
Right?

Swaathik (approved):

PR #705, opened 2023-05-05
Side button relocation 2

minjoonkim:
## Purpose
Feedback and help buttons are covering the scroll bar, which users have complained. 

## Approach
If the scroll bar exists, it moves the feedback and help buttons 17px from the right.
If the scroll bar is like a "mac" style where it only appears on scroll, the buttons are unmoved

## Testing
Help and fallback feedback button (google forms) were checked. 
HotJar feedback button code is there, but needs check in live because it is not able to test in localhost

## Checklist

### DIFF
--- a/src/help/components/contextual/SideButtons.tsx
+++ b/src/help/components/contextual/SideButtons.tsx
@@ -5,7 +5,8 @@ import { sleep } from 'timing-functions';
 
 import { LocationToPath, Location } from '../../../app/config/urls';
 
-import styles from './styles/side-buttons.module.scss';
+import sideButtonStyles from './styles/side-buttons.module.scss';
+import baseStyles from '../../../shared/components/layouts/styles/base-layout.module.scss';
 
 type Props = {
   displayHelp: boolean;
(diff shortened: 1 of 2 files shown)


src/help/components/contextual/SideButtons.tsx
```diff
@@ -16,6 +17,36 @@ const SideButtons = ({ displayHelp, onClick }: Props) => {
...
+    const mainContent = document?.querySelector<HTMLElement>(
+      `.${baseStyles['main-content']}`
+    );
+    const scrollBarWidth =
+      mainContent && mainContent.offsetWidth - mainContent.clientWidth;
+
+    if (scrollBarWidth) {
+      const hjButton = document.querySelector<HTMLElement>(
+        '#_hj_feedback_container div button'
+      );
```
aurel-l, line 30: this should probably go lower, within the function line 50, because hotjar is added later and doing this too early you might not get a handle on the button yet

src/help/components/contextual/SideButtons.tsx
```diff
@@ -16,6 +17,36 @@ const SideButtons = ({ displayHelp, onClick }: Props) => {
...
+
+      const sideButton = document?.querySelector<HTMLElement>(
+        `.${sideButtonStyles['side-button']}`
+      );
+      const sideButtonHelp = document?.querySelector<HTMLElement>(
+        `.${sideButtonStyles.help}`
+      );
+
+      if (hjButton) {
+        hjButton.style.right = '17px';
```
aurel-l, line 40: Shouldn't this, and the other "right" values below, be "= `${scrollBarWidth}px`"?

minjoonkim: On macs, the scroll bar width calculated is 15px, but that still has some overlap with the scroll bar for some reason. 15px looks like the right-most edge of the button on actual scrolling part of the scroll bar, while with 17px the edge is on the scroll area (if that makes sense). 

Windows has 17px regardless

src/help/components/contextual/SideButtons.tsx
minjoonkim: I was under the impression that hotjar was added first, and if there's no hotjar even after `sleep(3000)` we then display the fallback feedback button? Anyways I did move it so it's underneath the `sleep` in the next commit

aurel-l: The hotjar logic is running immediately, although on some browsers (probably slow ones) we noticed that there was a delay before adding the button. It would cause both or fallback feedback and the hotjar one to be loaded on top of each other, hence the extra 3000ms delay before checking

aurel-l (approved):

PR #709, opened 2023-05-18
add embedding option to async download

aurel-l:
## Purpose
Add embeddings to download options [TRM-29410](https://www.ebi.ac.uk/panda/jira/browse/TRM-29410)

## Approach
Use the Async download logic for any download of the h5/embedding format
Link to FTP file when any of the pre-computed embedded files could be used instead (see https://www.uniprot.org/help/downloads)

## Testing
Manual testing + updated tests for the detection of query that can be linked to FTP instead

## Checklist


aurel-l:
Pending at the moment: identify existing files on FTP and point to them for specific queries

src/shared/components/download/DownloadPreview.tsx
```diff
@@ -40,9 +42,13 @@ const DownloadPreview = ({
...
       <h4>Preview</h4>
-      <CodeBlock lightMode data-testid="download-preview">
-        {typeof data === 'string' ? data : JSON.stringify(data, null, 2)}
-      </CodeBlock>
+      {data ? (
+        <CodeBlock lightMode data-testid="download-preview">
+          {typeof data === 'string' ? data : JSON.stringify(data, null, 2)}
+        </CodeBlock>
+      ) : (
+        'No preview available for this format'
```
dlrice, line 50: Similar approach will be needed for the API URLs.

aurel-l: Good catch, I'd forgotten to handle this case.
I've now handled it by disabling the whole API section in this case

dlrice (approved):

PR #710, opened 2023-05-18
Group by taxonomy

dlrice:
**Needs UX as well as code review**

## Purpose
Group by for taxonomy [TRM-29493](https://www.ebi.ac.uk/panda/jira/browse/TRM-29493)

## Approach
- Create group by facet in similar fashion to legacy
- Url changes to `groupBy=taxonomy`
- Display the top-level node and an active parent which corresponds to url param eg `parent=9606`
- Display group by within results with access retained to other facets. On facet selection the group by will reload, losing any open nodes. However the active parent will be retained.
- Use small up and down arrows to change active parent or return to return to the top level
- Provide autocomplete search box to set active parent

## Testing
Created snapshot

## Checklist


src/shared/config/apiUrls.ts
```diff
@@ -109,8 +109,19 @@ const apiUrls = {
...
+    if (!id) {
+      return undefined;
+    }
+    const url = joinUrl(apiPrefix, namespace, id);
+    if (columns?.length) {
+      return queryString.stringifyUrl({
+        url,
+        query: { fields: columns.join(',') },
+      });
+    }
```
aurel-l, line 122: Is this for future-proofing for download customisation on entry pages?
I couldn't find the columns parameter used anywhere in the PR or in the rest of the codebase

aurel-l:
It would be good that, just like when clicking on facets while in the groupby view we stay in the group by view, we have the same behaviour when:
- Modifying the search from the top search bar
- Clicking on the suggestions next to the result count (`or search "human" as a`)
- Submitting a new advanced search
This last one I'm not entirely sure about, as the intent from the user might be to restart a brand new search and to rest the view, but in the end it can be the same when submitting a new search from the top bar. It's complicated to decide when a new search is a refinement of the previous one, or a brand new one.
What's your opinion on that.

Changes needed would be to go in the different components and to reimplement the links to have their `to` prop use a function of the previous location to only modify the `search` and return the exact same other query parameters (like the `groupby` and `parent` ones) without having to explicitely pass them

dlrice:
> It would be good that, just like when clicking on facets while in the groupby view we stay in the group by view

Yes, I agree with this. I subconsciously thought this when developing and testing it.

> * Submitting a new advanced search
>   This last one I'm not entirely sure about, as the intent from the user might be to restart a brand new search and to rest the view, but in the end it can be the same when submitting a new search from the top bar. It's complicated to decide when a new search is a refinement of the previous one, or a brand new one.
>   What's your opinion on that.

As we have the group by as a facet (albeit it's a weird facet!), I think for consistency creating a search from the query builder should remove previous facets.

src/uniprotkb/components/results/styles/group-by.module.scss
```diff
@@ -0,0 +1,136 @@
...
+        height: 1ch;
+
+        button {
+          padding: 0;
+          border: 0;
+          margin-top: -0.1rem;
+          margin-bottom: 0rem;
+          margin-left: 0.725rem;
+          transition: transform 0.25s ease-out;
+          outline: 0;
```
aurel-l, line 101:
```suggestion
```

Would suggest removing this, as when tabbing around we can't see where the focus is

src/shared/config/apiUrls.ts
aurel-l: Ah, I see, makes sense.
I guess leave it here because it might be useful in the future anyway

aurel-l (approved):
All good for me.
I did address some small things in my latest commits so if you want have a look there and if you disagree we can discuss or revert, it was perfectly functional before so the latest changes are not bugfixes or anything critical

aurel-l:
> > It would be good that, just like when clicking on facets while in the groupby view we stay in the group by view
> 
> Yes, I agree with this. I subconsciously thought this when developing and testing it.
> 
> > * Submitting a new advanced search
> >   This last one I'm not entirely sure about, as the intent from the user might be to restart a brand new search and to rest the view, but in the end it can be the same when submitting a new search from the top bar. It's complicated to decide when a new search is a refinement of the previous one, or a brand new one.
> >   What's your opinion on that.
> 
> As we have the group by as a facet (albeit it's a weird facet!), I think for consistency creating a search from the query builder should remove previous facets.

I think we can leave it like this for now and see if users provide feedback about unexpected behaviour (in their opinion) when they use it

PR #711, opened 2023-05-22
More ga4 events

dlrice:
## Purpose
Add remaining ga4 events as described in [confluence document](https://www.ebi.ac.uk/seqdb/confluence/pages/viewpage.action?spaceKey=UniProt&title=Google+Analytics+4) [Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-29187)

## Testing
None

## Checklist

### DIFF
--- a/src/proteomes/components/entry/ComponentsButtons.tsx
+++ b/src/proteomes/components/entry/ComponentsButtons.tsx
@@ -46,7 +46,7 @@ const ComponentsButtons = ({
 }: Props) => {
   const [displayDownloadPanel, setDisplayDownloadPanel] = useState(false);
 
-  const handleDownloadToggle = useCallback(
+  const handleToggleDownload = useCallback(
     (reason: DownloadPanelFormCloseReason, downloadMethod?: DownloadMethod) => {
       if (displayDownloadPanel) {
         sendGtagEventPanelResultsDownloadClose(reason, downloadMethod);
@@ -102,7 +102,7 @@ const ComponentsButtons = ({
           <SlidingPanel
             title="Download"
             position="left"
-            onClose={handleDownloadToggle}
+            onClose={handleToggleDownload}
           >
             <ErrorBoundary>
               <DownloadComponent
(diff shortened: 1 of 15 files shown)


src/proteomes/components/entry/ComponentsButtons.tsx
```diff
@@ -40,6 +46,19 @@ const ComponentsButtons = ({
...
 
+  const handleToggleDownload = useCallback(
+    (reason: DownloadPanelFormCloseReason, downloadMethod?: DownloadMethod) => {
+      if (displayDownloadPanel) {
+        sendGtagEventPanelResultsDownloadClose(reason, downloadMethod);
+        setDisplayDownloadPanel(false);
+      } else {
+        sendGtagEventPanelOpen('results_download');
+        setDisplayDownloadPanel(true);
+      }
```
aurel-l, line 57:
```suggestion
      setDisplayDownloadPanel(displayDownloadPanel => {
        if (displayDownloadPanel) {
          sendGtagEventPanelResultsDownloadClose(reason, downloadMethod);
        } else {
          sendGtagEventPanelOpen('results_download');
        }
        setDisplayDownloadPanel(!displayDownloadPanel)
      });
```

I would do something like this, to avoid having `displayDownloadPanel` as a dependency of the effect.
This has been edited in GitHub, so if you go for something like this make sure that lint and type checks still pass

src/shared/components/results/ResultsButtons.tsx
```diff
@@ -154,6 +161,19 @@ const ResultsButtons: FC<ResultsButtonsProps> = ({
     invalidUrlViewMode,
   ]);
 
+  const handleToggleDownload = useCallback(
```
aurel-l, line 164: Same comment as https://github.com/ebi-uniprot/uniprot-website/pull/711/files#r1200059944

dlrice: Good point, done.

aurel-l (approved):

PR #714, opened 2023-05-30
Redirect to star search when query is nonexistent or is empty string.

dlrice:
## Purpose
Deal with situation where the results route doesn't have a `query=` url parameter or this is just blank.

https://www.ebi.ac.uk/panda/jira/browse/TRM-26268
https://www.ebi.ac.uk/panda/jira/browse/TRM-29073

## Approach
Add logic within `ResultsOrLanding` to detect absence of `query=` and add this in as `query=*`.

## Testing
None

## Checklist

### DIFF
--- a/src/app/components/App.tsx
+++ b/src/app/components/App.tsx
@@ -16,6 +16,7 @@ import {
   Replay,
 } from '@sentry/react';
 import { Integrations as SentryIntegrations } from '@sentry/tracing';
+import queryString from 'query-string';
 
 import BaseLayout from '../../shared/components/layouts/BaseLayout';
 import { SingleColumnLayout } from '../../shared/components/layouts/SingleColumnLayout';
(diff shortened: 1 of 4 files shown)


src/app/components/App.tsx
```diff
@@ -284,16 +285,31 @@ const BackToTheTop = lazy(() =>
...
+    if (props.location.search) {
+      const params = Object.fromEntries(
+        new URLSearchParams(props.location.search)
+      );
+      if (!params.query) {
+        return (
+          <Redirect
+            to={queryString.stringifyUrl({
+              url: props.location.pathname,
+              query: { ...params, query: '*' },
```
Swaathik, line 304: I didn't know this is coming. I added a specific query=* targeting one thing here - https://github.com/ebi-uniprot/uniprot-website/commit/b9ab0e073118e8bc9b5b84508fce31c97071cc90

I tested by reverting it and the results page load with query=*. I will commit (undoing my previous fix) in this PR if you don't mind

PR #716, opened 2023-06-07
Add important job parameters in tools dashboard

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-28866

## Approach
Added a component to deal with specific cases

## Testing

## Checklist


src/tools/dashboard/components/styles/Dashboard.scss
```diff
@@ -237,7 +245,8 @@
         grid-template-areas:
           'type name   time'
           '.    status time'
-          'id   id     actions';
+          'id   id     actions'
+          'parameters parameters .';
         grid-template-columns: 10ch minmax(27ch, auto) 12ch;
         grid-gap: 0;
         grid-template-rows: 1fr 1fr 1fr;
```
aurel-l, line 252: Haven't had a chance to test yet, but here since you've added a row it might be good to be explicit about the sizing of that row by adding a fourth value. I'm assuming this should be `auto` and not `1fr`, the point being that it shouldn't have any height if no parameter is displayed

Swaathik: You are right. I set it to auto and tested. It takes space only when there is content

src/tools/dashboard/components/Row.tsx
```diff
@@ -258,6 +265,94 @@ const NiceStatus = ({ job, jobLink, jobUrl }: NiceStatusProps) => {
   }
 };
 
+interface JobSpecificParametersProps {
+  job: Job;
+}
+
+const taxonsWithEllipsisReveal = (taxIDs: SelectedTaxon[]) => {
+  const visibleID = taxIDs[0];
+  const hiddenIDs = taxIDs.slice(1);
```
aurel-l, line 274:
```suggestion
  const [visibleID, ...hiddenIDs] = taxIDs;
```

src/tools/dashboard/components/Row.tsx
```diff
@@ -258,6 +265,94 @@ const NiceStatus = ({ job, jobLink, jobUrl }: NiceStatusProps) => {
...
+        ? Object.fromEntries(
+            idMappingFields.groups.flatMap(({ items }) =>
+              items.map((item) => [item.name, item.displayName])
+            )
+          )
+        : {};
+      return dbNameToDbInfo;
+    }, [idMappingFields]);
+
+  if (job.status === Status.FINISHED) {
```
aurel-l, line 306: Any specific reason to only display this when finished? I think it might be useful also when running, especially to understand why one might take long to finish

src/tools/dashboard/components/Row.tsx
```diff
@@ -258,6 +265,94 @@ const NiceStatus = ({ job, jobLink, jobUrl }: NiceStatusProps) => {
...
+          {hiddenIDs.map((taxon) => taxon.label).join(', ')}
+        </EllipsisReveal>
+      ) : null}
+    </span>
+  );
+};
+
+const JobSpecificParamaters = ({ job }: JobSpecificParametersProps) => {
+  const { data: idMappingFields } = useDataApi<IDMappingFormConfig>(
+    apiUrls.idMappingFields
```
aurel-l, line 291:
```suggestion
    job.type === JobTypes.ID_MAPPING ? apiUrls.idMappingFields : undefined
```
to avoid doing a network request when it's not needed

Swaathik: I had a condition outside but using hooks inside posed a problem. it is a clever way to do it inside

src/tools/dashboard/components/Row.tsx
aurel-l: Yeah, I think it might be useful for any job in any status

aurel-l (approved):

PR #718, opened 2023-06-14
Did you mean component changes

Swaathik:
## Purpose
Update sequence validator in Franklin for peptide search suggestions
Group the suggestion by namespace for better readability

## Approach

## Testing

## Checklist

### DIFF
--- a/src/shared/components/results/DidYouMean.tsx
+++ b/src/shared/components/results/DidYouMean.tsx
@@ -20,6 +20,7 @@ import {
   Location,
 } from '../../../app/config/urls';
 import apiUrls from '../../config/apiUrls';
+import { PEPTIDE_SEARCH_MINIMUM_LIMIT } from '../../../tools/peptide-search/components/PeptideSearchForm';
 
 import {
   Namespace,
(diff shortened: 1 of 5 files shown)


src/tools/peptide-search/components/PeptideSearchForm.tsx
```diff
@@ -50,7 +50,8 @@ import sticky from '../../../shared/styles/sticky.module.scss';
 import '../../styles/ToolsForm.scss';
 
 // just because, no actual known limit
-export const PEPTIDE_SEARCH_LIMIT = 100;
+export const PEPTIDE_SEARCH_MINIMUM_LIMIT = 7;
+export const PEPTIDE_SEARCH_MAXIMUM_LIMIT = 100;
```
aurel-l, line 54: Those are a bit tricky to understand side-by-side as the 7 is the number of aa in 1 sequence, and the 100 is the number of sequences supported at the same time in one search (that's what the "just because, no actual known limit" comment refers to).
I'd try to find different names compared to the ones here to make things clearer

src/shared/components/results/DidYouMean.tsx
```diff
@@ -214,7 +221,8 @@ const DidYouMean = ({
       content = (
         <div className={styles.suggestions}>
           Did you mean to search for:
-          <ul className={styles['suggestions-list']}>{suggestionNodes}</ul>
+          {suggestionNodes}
+          {/* <ul className={styles['suggestions-list']}>{suggestionNodes}</ul> */}
```
aurel-l, line 225: The commented out line can probably be removed now

src/tools/peptide-search/components/PeptideSearchForm.tsx
Swaathik: Right. PEPTIDE_SEARCH_AA_MINIMUM_LENGTH or PEPTIDE_SEARCH_SEQ_MINIMUM_LENGTH for the former?
PEPTIDE_SEARCH_SEQUENCES_COUNT for the latter?

aurel-l: Either sound good to me 👍🏽

aurel-l (approved):

PR #724, opened 2023-07-14
Advanced search suggestions

Swaathik:
## Purpose
There should be no suggestions of queries which
- return 0 hits
- return the same hits as the original question.

As the search results of a suggestion are subsets of the whole search, comparing the number of hits is enough.

 e.g. https://www.uniprot.org/uniprotkb?query=BSU06210 don't
suggest to restrict to gene name.

https://www.uniprot.org/uniprotkb?query=gene%3Acyk3%20AND%20organism_id%3A284812
don't suggest exact match as it is the same result

## Approach
Exact match, taxonomy and matching fields suggestion takes into account the number of results to decide. if it is different to the total number of hits that are currently shown in the results page, the suggestion is shown

## Testing

## Checklist

### DIFF
--- a/src/shared/components/results/AdvancedSearchSuggestion.tsx
+++ b/src/shared/components/results/AdvancedSearchSuggestion.tsx
@@ -113,7 +113,7 @@ const AdvancedSearchSuggestion = ({
   }
 
   /*  If the single matched field's number of hits is equal to the existing total that is displayed, do not suggest.
-      Ideally we need to fetch results of the relevant suggestion to compare with the current result set
+      As the results will be a subset of the overall search anyway, comparing by number of hits is enough in this use case.
   */
   if (termsToDisplay === 1 && searchTerms[0].hits === total) {
     return null;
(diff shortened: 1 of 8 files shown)


dlrice (approved):

src/shared/components/results/TaxonomyLevelsSuggestion.tsx
```diff
@@ -6,29 +8,68 @@ import {
...
+
+  const { data } = useDataApi<SearchResults<UniProtkbAPIModel>>(
+    `${apiUrls.search(Namespace.uniprotkb)}?${searchParams}`
+  );
+
+  useEffect(() => {
+    if (data?.results.length && data?.results.length !== total) {
+      setShowTaxonSuggestion(true);
+    }
+  }, [data, total]);
```
aurel-l, line 45: Any reason why this is in a `useEffect`?
This is derived from data and from a prop (`total`) so it could probably just be:
```ts
const showTaxonSuggestion = data?.results.length && data?.results.length !== total;
```

src/shared/components/results/AdvancedSearchSuggestion.tsx
```diff
@@ -23,17 +22,25 @@ type MatchedFieldsResponse = {
...
+  query: string;
+  total: number;
+}) => {
   const [linesToDisplay, setLinesToDisplay] = useState(1);
   const [termsToDisplay, setTermsToDisplay] = useState(+Infinity);
   const ref = useRef<HTMLElement>(null);
 
+  const searchParams = new URLSearchParams({
+    size: '0',
+    query: `${query}`,
```
aurel-l, line 38:
```suggestion
    query,
```

PR #725, opened 2023-07-14
Rich text dbsnp

dlrice:
## Purpose
[Improve RichText component to parse variant description texts (parse dbsnp links)](https://www.ebi.ac.uk/panda/jira/browse/TRM-29431)

## Approach

- Add dbsnp regexp
- Extract `getTextProcessingParts` create tests as I was messing things up but couldn't figure out why
- Made explicit capture and non-capture groups to avoid multiple regex matches

## Testing
Updated and added

## Checklist

### DIFF
--- a/src/uniprotkb/components/protein-data-views/FreeTextView.tsx
+++ b/src/uniprotkb/components/protein-data-views/FreeTextView.tsx
@@ -1,6 +1,7 @@
 import { Fragment, FC, ReactNode } from 'react';
 import { Link, useRouteMatch } from 'react-router-dom';
 
+import { ExternalLink } from 'franklin-sites';
 import UniProtKBEvidenceTag from './UniProtKBEvidenceTag';
 import SimilarityView from './SimilarityView';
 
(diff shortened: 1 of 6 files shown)


src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -93,6 +93,15 @@ export const RichText = ({ children, addPeriod, noLink }: RichTextProps) => (
             </Fragment>
           );
         }
+        const dbSnpMatch = part.match(reDbSnpCapture);
+        if (dbSnpMatch?.groups) {
+          const { rsid } = dbSnpMatch.groups;
+          return (
+            <ExternalLink url={externalUrls.dbSNP(rsid)} key={rsid}>
```
dlrice, line 100: `ExternalLink` used here however pubmed is using just a link. Not sure if there is an intention for pubmed to exceptionally not use `ExternalLink`.

dlrice: Disregard, I forgot that pubmed links internally.

PR #726, opened 2023-07-19
2023 09 datatable expand button

aurel-l:
## Purpose
Only show the expand button for datatable when there is something to expand to. [TRM-28963](https://www.ebi.ac.uk/panda/jira/browse/TRM-28963)

## Approach
Detect scroll within the Nightingale datatable container and if there is no scroll hide the expand button.
Do the check anytime the content changes (data loaded, class changed when hiding/showing rows) with a MutationObserver.

## Testing
Manual testing as MutationObserver is tricky to test in Jest/jsdom. Better coverage by adding a test about the button itself.

## Checklist

### DIFF
--- a/src/shared/components/views/DatatableWithToggle.tsx
+++ b/src/shared/components/views/DatatableWithToggle.tsx
@@ -22,7 +22,7 @@ const DatatableWithToggle = ({ children }: { children: ReactNode }) => {
   const tableRef = useRef<HTMLElement>(null);
   const firstRenderRef = useRef(true);
 
-  const params = useParams<{ accession: string }>();
+  const params = useParams<{ accession?: string }>();
 
   const datatableElement = useCustomElement(
     /* istanbul ignore next */
@@ -45,7 +45,7 @@ const DatatableWithToggle = ({ children }: { children: ReactNode }) => {
         });
       }
       sendGtagEventFeatureDataTableViewClick(
-        params.accession,
+        params.accession || '',
         expandTable ? 'expanded' : 'collapsed'
       );
     }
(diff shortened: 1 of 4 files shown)


src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -47,6 +55,28 @@ const DatatableWithToggle: FC = ({ children }) => {
     }
   }, [expandTable, params.accession]);
 
+  const sizeCheck = useCallback(() => {
+    setShowButton(
+      tableRef.current?.shadowRoot?.firstElementChild?.scrollHeight !==
+        tableRef.current?.shadowRoot?.firstElementChild?.clientHeight
+    );
+  }, []);
```
aurel-l, line 63: Feels a bit hacky to rely on implementation detail of the datatable in Nightingale, but that works, and the datatable will eventually move away from Nightingale

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -57,18 +87,15 @@ const DatatableWithToggle: FC = ({ children }) => {
         >
           {children}
         </datatableElement.name>
-        {/* TODO show button only when the content exceeds visible area of the table  
-        {tableRef?.current?.rows.length > 7 && (
-          
+        {(showButton || expandTable) && (
```
aurel-l, line 90: `|| expandTable` is because if the button has been triggered the scroll will disappear and `showButton` will become false, so make sure to ignore that here otherwise there would not be a way to collapse the table again.

src/shared/components/views/DatatableWithToggle.tsx
```diff
@@ -8,13 +15,14 @@ import { sendGtagEventFeatureDataTableViewClick } from '../../utils/gtagEvents';
 
 import './styles/datatable-toggle.scss';
 
-const DatatableWithToggle: FC = ({ children }) => {
+const DatatableWithToggle = ({ children }: { children: ReactNode }) => {
```
aurel-l, line 18: Just removing `FC` now because future versions of React won't have this type available

dlrice (approved):

PR #729, opened 2023-08-04
Remove query string pkg

dlrice:
## Purpose
[Create url search param utility fn](https://www.ebi.ac.uk/panda/jira/browse/TRM-29979)

## Approach
- `stringifyQuery`: returns a query string by iterating over the args and creating a combined URLSearchParams instance. If a parameter key already exists it will be overwritten by the new value. If the new value is undefined or null then the parameter will be removed. Arrays will be returned as comma separated strings. This also returns the parameters in sorted order.
- `stringifyUrl`: combines a "base" url with query parameters.
- `splitUrl`: simply returns the "base" url and the query string.
- Always show the branch as I find this helpful as I have several versions of the website served locally and it can be tricky to determine which is being served at which port.
- Cleaned up a few trailing `/`s when joining urls.
- Remove `query-string` package however it's still in the yarn lock file as franklin still uses it. I will create a PR for this too.

## Testing
Updated and tested.

## Checklist

### DIFF
--- a/src/shared/config/ftpUrls.ts
+++ b/src/shared/config/ftpUrls.ts
@@ -21,7 +21,7 @@ const ftpUrls = {
       ftpUniProt,
       `/current_release/knowledgebase/pan_proteomes/${id}.fasta.gz`
     ),
-  embeddings: joinUrl(ftpUniProt, '/current_release/knowledgebase/embeddings/'),
+  embeddings: joinUrl(ftpUniProt, 'current_release/knowledgebase/embeddings'),
 };
 
 const restFormatToFtpFormat = new Map([
--- a/src/shared/config/ftpUrls.ts
+++ b/src/shared/config/ftpUrls.ts
@@ -81,11 +81,10 @@ export const getUniprotkbFtpFilenameAndUrl = (
 ) => {
   const sp = new URLSearchParams(downloadUrl);
   const query = sp.get('query');
-  const q = Array.isArray(query) ? query[0] : query;
-  if (!q) {
+  if (!query) {
     return null;
   }
-  const simplifiedQuery = simplifyQuery(q);
+  const simplifiedQuery = simplifyQuery(query);
   if (!simplifiedQuery) {
     return null;
   }
--- a/src/shared/config/ftpUrls.ts
+++ b/src/shared/config/ftpUrls.ts
@@ -1,6 +1,5 @@
 import { capitalize } from 'lodash-es';
 import joinUrl from 'url-join';
-import queryString from 'query-string';
 
 import { FileFormat } from '../types/resultsDownload';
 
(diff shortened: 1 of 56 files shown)


src/shared/utils/url.ts
```diff
@@ -10,21 +9,40 @@ export const getLocationForPathname = (pathname: string) => {
...
+      }
+    }
   }
-  return parsedWithoutArrayValues;
+  combined.sort();
+  return combined.toString();
+};
+
+export const stringifyUrl = (base: string, ...args: QueryStringArg[]) =>
+  `${base}?${stringifyQuery(...args)}`;
```
aurel-l, line 43: I would maybe stringify the query first, then if it's truthy I would happen both the question mark and the stringified query (instead of having the URL finish with a strange `?` only

src/shared/config/ftpUrls.ts
```diff
@@ -80,8 +79,8 @@ export const getUniprotkbFtpFilenameAndUrl = (
   downloadUrl: string,
   format: FileFormat
 ) => {
-  const parsed = queryString.parseUrl(downloadUrl);
-  const { query } = parsed.query;
+  const sp = new URLSearchParams(downloadUrl);
+  const query = sp.get('query');
   const q = Array.isArray(query) ? query[0] : query;
```
aurel-l, line 84: I'd say it's safe to remove this check now, because `get` will only give you a string value corresponding to the first value in the search string (in contrast with `getAll`)

src/shared/utils/url.ts
```diff
@@ -10,21 +9,46 @@ export const getLocationForPathname = (pathname: string) => {
...
+export const stringifyQuery = (...args: QueryStringArg[]) => {
+  // This returns a query string by iterating over the args and creating a combined
+  // URLSearchParams instance. If a parameter key already exists it will be overwritten
+  // by the new value. If the new value is undefined or null then the parameter will
+  // be removed. Arrays will be returned as comma separated strings.
+  const combined = new URLSearchParams();
+  for (const arg of args) {
+    const iter =
+      // Query string
+      (typeof arg === 'string' && new URLSearchParams(arg)) ||
```
aurel-l, line 27: Not sure what `&& new URLSearchParams(arg)` is for, it will always return a truthy value as it will return an URLSearchParameter object, even if there is nothing in `arg`.

src/shared/utils/url.ts
```diff
@@ -10,21 +9,46 @@ export const getLocationForPathname = (pathname: string) => {
...
+      // URLSearchParams
+      (arg instanceof URLSearchParams && arg) ||
```
aurel-l, line 29: Similar question about the `&& arg`, if the first clause is true then this bit will be truthy too

aurel-l (approved):

PR #731, opened 2023-08-10
Idmapping async download

dlrice:
## Purpose
[Async Download for ID Mapping](https://www.ebi.ac.uk/panda/jira/browse/TRM-29907)

## Approach
- Create form reducer for async download
- Add new file formats with (from/to only) for unenriched results
- Adjust logic in Download component for async download of id mapping. Some of this involved reordering statements.

## Testing
Created and updated.

## Checklist


dlrice:
Refactored the Download component:

- Created a reducer state
- Extracted functions which use the state, props, location and job as arguments. There are duplicate calls being made to the same function but I believe these will be very quick and it helps with encapsulating all of the logic
- Extracted proteomes components downloads into own ComponentsDownload component
- Updated unisave downloads to pass previewSelected in order to prevent large preview downloads from stalling the page

Swaathik (approved):

PR #732, opened 2023-08-22
Show deleted entries count in ID mapping results

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-29997

Example IDs: P12345,P05067,A0A852YP84,A0A008APQ8

## Testing
Added a test

## Checklist


aurel-l (approved):
I've changed the way the links are generated because at the moment they were pointing to a UniProtKB search while they should point to a search within the ID Mapping result set

PR #733, opened 2023-08-24
Covid redirect warning

dlrice:
## Purpose
COVID-19 portal is being shut down so alert users if they have been redirected.

Corresponds to this [traffic-manager MR](https://gitlab.ebi.ac.uk/wp/wp-traffic-manager/-/merge_requests/1175).

## Approach

- Check `window.location.search` for `fromCovid19Portal=true` and if present show a dismissible message.
- Renamed the existing deployment warning for clarity.

## Testing
None

## Checklist

### DIFF
--- a/src/app/components/Covid19RedirectWarning.tsx
+++ b/src/app/components/Covid19RedirectWarning.tsx
@@ -1,0 +1,28 @@
+import { Message } from 'franklin-sites';
+import { useState } from 'react';
+
+import style from './styles/warning-message.module.scss';
+
+const reFromCovidPortal = /fromCovid19Portal=true/i;
+
+const Covid19DeploymentWarning = () => {
+  const [dismissed, setDismissed] = useState(false);
+  return (
+    <>
+      {!window.location.search.match(reFromCovidPortal) &&
+      !LIVE_RELOAD &&
+      !dismissed ? (
+        <Message
+          className={style['warning-message']}
+          level="warning"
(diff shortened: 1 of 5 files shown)


src/app/components/Covid19RedirectWarning.tsx
```diff
@@ -0,0 +1,28 @@
...
+      {window.location.search.match(reFromCovid19Portal) &&
+      !LIVE_RELOAD &&
+      !dismissed ? (
+        <Message
+          className={style['warning-message']}
+          level="warning"
+          onDismiss={() => setDismissed(true)}
+        >
+          You have been redirected from the COVID-19 Portal which has been shut
+          down.
```
aurel-l, line 21: Not sure if there's a smoother verb than "shut down" in English.
I would also add something like "All the entries in the portal are now fully integrated into UniProtKB".

aurel-l (approved):
I just left a comment about the wording of the message.

src/app/components/Covid19RedirectWarning.tsx
dlrice: Discontinued?

dlrice: You have been redirected from the COVID-19 Portal which is now obsolete as all entries from the portal have been fully integrated into UniProtKB.

aurel-l: Yep, that suggestion sounds good

PR #742, opened 2023-09-28
Replace download dropdown with sliding panel in entry pages

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-27653
https://www.ebi.ac.uk/panda/jira/browse/TRM-27977
https://www.ebi.ac.uk/panda/jira/browse/TRM-27654

## Approach

## Testing

## Checklist


src/shared/components/entry/EntryDownload.tsx
```diff
@@ -107,142 +100,179 @@ const DownloadAnchor = ({
...
+  );
+
+  if (showPreview) {
+    extraContentNode =
+      fileFormat !== FileFormat.excel ? (
+        <DownloadPreview
+          previewUrl={downloadUrl}
+          previewFileFormat={fileFormat}
+        />
+      ) : null;
```
dlrice, line 156: Could we just swap excel for tsv as we do for search results download?

dlrice (commented):
I can't see the canonical & isoform option like we have in production:

Screenshot 2023-10-20 at 12 47 09

PR #743, opened 2023-10-02
Stats histograms

dlrice:
## Purpose
[Create sequence size plots](https://www.ebi.ac.uk/panda/jira/browse/TRM-30233)

## Approach
Use D3 to create a line plot. Note that the data set selection will be done via a franklin Tab component in a future PR.

## Testing
None

## Checklist


src/uniprotkb/components/statistics/SequenceLength.tsx
```diff
@@ -0,0 +1,65 @@
+import { useMemo, useState } from 'react';
+import cn from 'classnames';
+
+import SequenceLengthLinePlot from './SequenceLengthLinePlot';
+
+import { StatisticsItem } from './StatisticsPage';
+
+import styles from './styles/sequence-length.module.scss';
+
+export type SequenceLengthCount = [number, number];
```
aurel-l, line 10: Not a requirement at all, but since TypeScript allows us to do this I like to name the values within fixed tuples, that helps when looking at the code again in the future, and also it is picked up by IDEs so that can also be useful when coding ("what was it at index 0? the count or the length?").
In that case it could be something like: `export type SequenceLengthCount = [length: number, count: number];`

aurel-l (approved):

PR #745, opened 2023-10-06
Genomic Coordinates v2

aurel-l:
## Purpose
Genomic Coordinates https://www.ebi.ac.uk/panda/jira/browse/TRM-29924
Change the tab according to received feedback, more focus on genes and on comparison of exons across isoforms

## Approach
Change the shape of the payload returned by the Proteins API in order to match the expected UI structure of gene > isoforms > exons.

Separate some of that data changing logic into a utils file.

When IDs are not ENS... do not render any link, mainly because I couldn't find a reliable way to generate links for the coordinates to the genomic browser.

## Testing
Update tests, add tests for the utility functions

To check behaviour you can use:
- P62805: Multiple genes, multiple locations on multiple chromosomes
- A0A0A0MTS7: Big protein, with lots of exons
- P05067: Multiple isoforms, some of which without genomic coordinates info
- A2VEY9: Multiple isoforms, non-ENS... IDs, multiple translations/transcripts in some positions
- P42283: Maximum number of isoforms I could find, some isoforms don't even have the same accession as the canonical

## Checklist


src/uniprotkb/components/entry/tabs/genomic-coordinates/Entries.tsx
```diff
@@ -0,0 +1,346 @@
...
+                  </td>
+                  {mappedIsoforms.map((isoformID) => {
+                    const exon = exons.find(
+                      (exon) => exon.accession === isoformID
+                    );
+                    if (!exon) {
+                      return <td key={isoformID}>-</td>;
+                    }
+                    return (
+                      <td key={isoformID}>
```
dlrice, line 302: I'm wondering if there is any way to make it more explicit to the user that this column is the protein coordinates? We have a header for the genomic coordinates but not for the proteins perhaps because it's a bit awkward as there can be multiple. I thought of a few ideas:

1. Use `title="protein coordinates for P05067-1 which map to 12:2323-24343"`
2. Have a header which spans all of the isoforms but left align the text to avoid the situation where a middle align disappears for wide tables/narrow screens.

src/uniprotkb/components/entry/tabs/genomic-coordinates/Entries.tsx
```diff
@@ -0,0 +1,346 @@
...
+            : index}
+        </h3>
+      }
+    >
+      <InfoList infoData={infoData} columns />
+      {accessionEntriesPairs.map(([accession, entries]) => (
+        <Entry key={accession} entries={entries} xrefInfo={xrefInfo} />
+      ))}
+      <DatatableWrapper alwaysExpanded>
+        <table className={helper['no-wrap']}>
```
dlrice, line 220: Is it possible to increase the column padding to help distinguish the content. It's a bit tricky in the extreme cases to know where one column starts and finishes:

Screenshot 2023-10-20 at 12 24 43

src/uniprotkb/components/entry/tabs/genomic-coordinates/Entries.tsx
aurel-l: In the end, after the discussion in the UX thread, I've implemented a variant of you suggestion 1, on the column header.
We can always interate on that later on I guess

src/uniprotkb/components/entry/tabs/genomic-coordinates/Entries.tsx
aurel-l: I've added an inline padding of 1ch on each side of each cell, so there should be a bit more breathing room now

dlrice (approved):

PR #748, opened 2023-10-25
Stats updates

dlrice:
## Purpose
Update uniprotkb statistics page based on UX feedback.

## Approach

- [x] Extract some components into own files
- [x] Exclude "AND *" from queries
- [x] Create expand table after eg 10 rows
- [x] Add links to sequences ... AND fragment:false
- [x] Use tabs for reviewed/unreviewed
- [x] Place Publication table into Statistics for some line types
- [x] Move UNIQUE_AUTHOR to caption of Publication table
- [x] Miscellaneous statistics > Remove UNIQUE_AUTHOR, UNIQUE_CITATION_ID, SEQUENCE_CORRECTION from table
- [x] Viz before tables
- [x] Use tabs throughout for un/reviewed switching
- [x] Drop percentage everywhere except for amino acid count
- [x] Amino acid composition
    - [x] always expand table
    - [x] remove Entry count
    - [x] Name --> Amino Acid
    - [x] use all AAs
    - [x] Make jira for api to add labels for all
- [x] Top journal table > rename Count to Citations
- [x] Drop Sequence Count table
- [x] Entries count --> Entries with <nameLabel>
- [x] Remove UniProtKB from Unique references
- [x] Remove margin for statistics page table
- [x] Header > tabs > table
- [x] Remove "Table of" from all table titles
- [x] Make a jira for adding average per uniprotkb entry

## Testing
None

## Checklist


src/uniprotkb/components/statistics/AminoAcidBarPlot.tsx
```diff
@@ -5,9 +5,9 @@ import { sum } from 'lodash-es';
 import { StatisticsCategory } from './StatisticsPage';
 
 // Specify the chart’s dimensions.
-const width = 500;
+const width = 340;
 const height = 400;
-const margin = { top: 10, right: 30, bottom: 40, left: 40 };
+const margin = { top: 20, right: 30, bottom: 60, left: 40 };
```
aurel-l, line 10: Looking at l. 104, the total width will end up being this width plus the 2 inline margins (410), so either the grid definition needs to be adjusted to create a new row at 410px instead of 400, or here the sum needs to stay at 400px maximum

aurel-l (approved):

PR #750, opened 2023-10-30
Distribute pie chart

dlrice:
## Purpose
Redistribute the pie chart data to avoid squished labels.

## Approach
Used @aurel-l's algorithm outline. I found that sorting ascended worked better.

## Testing
Unit tests.

## Checklist

### DIFF
--- a/src/uniprotkb/components/graphs/PieChart.tsx
+++ b/src/uniprotkb/components/graphs/PieChart.tsx
@@ -31,7 +31,7 @@ const radius = Math.min(width, height) / 2 - margin;
 
 // Function to distributre pie chart data to hopefully avoid label collisions
 export const distributeByEntryCount = (data: StatisticsGraphItem[]) => {
-  const sorted = data.sort((a, b) => a.entryCount - b.entryCount);
+  const sorted = new Array(data).sort((a, b) => a.entryCount - b.entryCount);
   const n = data.length;
   const a: StatisticsGraphItem[] = new Array(n);
   const middle = Math.floor(n / 2);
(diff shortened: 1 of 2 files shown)


src/uniprotkb/components/graphs/PieChart.tsx
```diff
@@ -29,6 +29,19 @@ const margin = 45;
 
 const radius = Math.min(width, height) / 2 - margin;
 
+// Function to distributre pie chart data to hopefully avoid label collisions
+export const distributeByEntryCount = (data: StatisticsGraphItem[]) => {
+  const sorted = data.sort((a, b) => a.entryCount - b.entryCount);
```
aurel-l, line 34:
```suggestion
  const sorted = new Array(data).sort((a, b) => a.entryCount - b.entryCount);
```
Data is coming from the props, in order to avoid unexpected issues it's better to make a copy of the array as sort will mutate the original one

dlrice: That behaviour surprises me! Good to know!

dlrice: It makes sense actually as it's an instance method. I noticed the tests failed so I used a spread instead.

aurel-l: Sorry I've just used the GitHub suggestion and didn't realise it wasn't the proper thing to use, I meant to use `Array.from()` which is essentially the same as what you used with the array spreading

aurel-l (approved):

PR #756, opened 2023-12-13
Proteins API entry download

Swaathik:
## Purpose
Addressing the latest feedback comments

## Approach

## Testing

## Checklist


src/shared/components/entry/styles/entry-download.module.scss
```diff
@@ -0,0 +1,3 @@
+.select-group-label {
+  font-style: italic;
+}
```
aurel-l, line 3: I don't think this is actually applying to anything at the moment, the select dropdown is not showing anything italicised within it for me (nor do I think this is supported by any browser)

Swaathik: Yes, it was supposed to be removed. I will remove

aurel-l (approved):

PR #765, opened 2024-01-03
Schema updates

dlrice:
## Purpose
[Fix schema issues and add more organization attributes.](https://www.ebi.ac.uk/panda/jira/browse/TRM-30601)

## Approach

- Fix course instance by adding the properties that google requires
- Create some config files: `description` and `socialUrls`
- Create `isCourseOnline` and `isCourseOnsite` util fns

## Testing
Copied `chrome > devtools > Elements > <html>` element and pasted into https://search.google.com/test/rich-results

## Checklist

### DIFF
--- a/src/shared/components/layouts/UniProtHeader.tsx
+++ b/src/shared/components/layouts/UniProtHeader.tsx
@@ -132,6 +132,10 @@ const organizationSchema: WithContext<Organization> = {
   name: 'UniProt',
   description,
   sameAs: Object.values(socialUrls),
+  contactPoint: {
+    '@type': 'ContactPoint',
+    url: 'https://www.uniprot.org/contact',
+  },
 };
 
 const UniProtHeader = () => {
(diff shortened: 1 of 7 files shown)


src/shared/components/layouts/UniProtHeader.tsx
```diff
@@ -127,6 +129,9 @@ const organizationSchema: WithContext<Organization> = {
   '@type': 'Organization',
   url: 'https://www.uniprot.org',
   logo: ['https://www.uniprot.org/android-chrome-512x512.png', Logo],
+  name: 'UniProt',
+  description,
+  sameAs: Object.values(socialUrls),
```
aurel-l, line 134:
```suggestion
  sameAs: Object.values(socialUrls),
  email: 'https://www.uniprot.org/contact',
```
Have you managed to get this bit detected in the Google rich-results tool?
For me it wouldn't appear at all, but when I added an email field it would appear as "Organization", although with a warning about invalid email. I guess it's better than nothing, we could have this and see if it starts appearing in the Google Search Console.
When adding a phone number it would additionally appear as a Local Business, but that's not really what we want to be detected as.

dlrice: Good catch, didn't even notice it organization wasn't being picked up. I can get it to work without a warning using:
```
+  contactPoint: {
+    '@type': 'ContactPoint',
+    url: 'https://www.uniprot.org/contact',
+  },
```

aurel-l: Ah, that's way nicer!

aurel-l (approved):

PR #767, opened 2024-01-05
Align sequence info mix

dlrice:
## Purpose
[Align sequence validation for non-UniProtKB / mixed namespaces / isoforms](https://www.ebi.ac.uk/panda/jira/browse/TRM-26756)

## Approach

- Create new id regexps: uniprotkb with isoform, uniparc and uniref and use these within existing `extractAccession` fn.
- Fetch uniprotkb, uniparc and uniref data to ensure sequence is the same

## Testing
Adapted and added `extractAccession` unit tests.

## Checklist

### DIFF
--- a/src/tools/align/utils/useSequenceInfo.ts
+++ b/src/tools/align/utils/useSequenceInfo.ts
@@ -11,9 +11,13 @@ import { getAccessionsURL } from '../../../shared/config/apiUrls';
 import { FeatureDatum } from '../../../uniprotkb/components/protein-data-views/UniProtKBFeaturesView';
 import { UniProtkbAPIModel } from '../../../uniprotkb/adapters/uniProtkbConverter';
 import { SearchResults } from '../../../shared/types/results';
+import { Namespace } from '../../../shared/types/namespaces';
+import { UniRefAPIModel } from '../../../uniref/adapters/uniRefConverter';
+import { UniParcAPIModel } from '../../../uniparc/adapters/uniParcConverter';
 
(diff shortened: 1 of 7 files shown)


src/tools/align/utils/useSequenceInfo.ts
```diff
@@ -22,63 +26,127 @@ export type SequenceInfo = {
...
+    }
+    for (const result of unirefResults?.data?.results || []) {
+      idToSequenceAndFeatures.set(result.id, {
+        sequence: result.representativeMember.sequence.value,
+      });
+    }
+    for (const result of uniparcResults?.data?.results || []) {
+      idToSequenceAndFeatures.set(result.uniParcId, {
+        sequence: result.sequence.value,
+        // features: result.sequenceFeatures,
```
dlrice, line 95: Is it expected to adapt the features as part of this task?

aurel-l: That was not in my mind when creating the jira originally, but it would make a lot of sense to do that I believe, even though it's a bit more work.

1. UniProtKB and UniProtKB isoforms should be the same, so not a big problem here.
2. UniParc should be simple too as we have InterPro features in the data, although the shape might be different as the features in UniProtKB.
3. UniRef might be a bit more complex. The sequence is that of the representative, and the representative is either a UniProtKB (canonical or isoform), or a UniParc. Maybe the logic could be that of fetching the features from the existing logic in points 1 and 2, extracting the representative from the UniRef ID (UniRef<identity>_<representative>.

Swaathik (approved):

PR #768, opened 2024-01-08
Include full version of cross reference on download

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-30429
Mockups:
https://www.ebi.ac.uk/panda/jira/browse/TRM-30428

## Approach
Updated Accordion in Franklin-sites for updating the label (https://github.com/ebi-uniprot/franklin-sites/pull/262)
Added a new prop 'isDownload' to indicate if the 'isMultiCrossReference' field from API should be taken into account when preparing the fields for accordion.

## Testing

## Checklist


src/shared/components/download/Download.tsx
```diff
@@ -102,6 +116,47 @@ const Download = (props: DownloadProps<JobTypes>) => {
...
+  );
+
+  const getFieldsWithMultipleValueXref = useCallback(
+    (fieldData: FieldData | FieldDatum, id: string): boolean | undefined => {
+      if (Array.isArray(fieldData)) {
+        for (const item of fieldData) {
+          const isFull = getFieldsWithMultipleValueXref(item, id);
+          if (isFull) {
+            return true;
+          }
```
dlrice, line 135:
```suggestion
          return getFieldsWithMultipleValueXref(item, id);
```

src/shared/components/download/Download.tsx
```diff
@@ -102,6 +116,47 @@ const Download = (props: DownloadProps<JobTypes>) => {
...
+  const { data } = useDataApi<ReceivedFieldData>(
+    apiUrls.resultsFields(namespace)
+  );
+
+  const fieldData = useMemo(
+    () => prepareFieldData(data, undefined, true),
+    [data]
+  );
+
+  const getFieldsWithMultipleValueXref = useCallback(
```
dlrice, line 128: It seems this function returns a boolean rather than the actual fields so what do you think about using an `is` for the function prefix eg `isMultiValueCrossReferenceField`. Thinking about it `isMultiValueCrossReference` is a bit confusing to me (I know it's coming from the config file) but basically it's saying the xref has a full option available? If that's true what do you think then about `isXrefWithFullOption`? Just an idea! I'm trying to imagine reading this in a years time and trying to remind myself what `isMultiValueCrossReference` means.

src/shared/components/download/Download.tsx
```diff
@@ -102,6 +116,47 @@ const Download = (props: DownloadProps<JobTypes>) => {
...
+          if (isFull) {
+            return true;
+          }
+        }
+      } else if (fieldData.items) {
+        return getFieldsWithMultipleValueXref(fieldData.items, id);
+      } else if (fieldData.id === id) {
+        return fieldData.addAsterisk;
+      }
+      return undefined;
```
dlrice, line 142:
```suggestion
      return true;
```
I think this would allow the function to be a fully boolean return.

src/shared/components/download/Download.tsx
```diff
@@ -102,6 +116,47 @@ const Download = (props: DownloadProps<JobTypes>) => {
...
+    },
+    []
+  );
+
+  useEffect(() => {
+    const multiValueXrefColumns = [];
+    for (const column of state.selectedColumns) {
+      const isPresent = getFieldsWithMultipleValueXref(
```
dlrice, line 150:
```suggestion
      const hasFullOption = getFieldsWithMultipleValueXref(
```

src/shared/components/download/Download.tsx
```diff
@@ -102,6 +116,47 @@ const Download = (props: DownloadProps<JobTypes>) => {
...
+        return fieldData.addAsterisk;
+      }
+      return undefined;
+    },
+    []
+  );
+
+  useEffect(() => {
+    const multiValueXrefColumns = [];
+    for (const column of state.selectedColumns) {
```
dlrice, line 149: Could a `filter` call be used here instead as it's just a flat array?

src/shared/components/download/Download.tsx
```diff
@@ -102,6 +116,47 @@ const Download = (props: DownloadProps<JobTypes>) => {
...
+      const isPresent = getFieldsWithMultipleValueXref(
+        fieldData,
+        column.includes('_full') ? column.replace('_full', '') : column
```
dlrice, line 152: I noticed this logic in two places - what do you think about extracting a util fn and calling it in those two places?

dlrice (approved):

PR #771, opened 2024-01-11
Redirect by UniProtKB ID

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-27101

## Approach
If any of the results from first fetch(first 25) matches, it is redirected to the entry page

## Testing
Added a test case

## Checklist

### DIFF
--- a/src/shared/components/results/ResultsData.tsx
+++ b/src/shared/components/results/ResultsData.tsx
@@ -127,15 +127,20 @@ const ResultsData = ({
       ) {
         history.replace(getEntryPathForEntry(uniqueItem));
       }
-    } else if (allResults.length && 'uniProtkbId' in allResults[0]) {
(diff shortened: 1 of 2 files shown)


src/shared/components/results/ResultsData.tsx
```diff
@@ -39,6 +39,7 @@ import { PaginatedResults } from '../../hooks/usePagination';
 import { Basket } from '../../hooks/useBasket';
 
 import styles from './styles/results-data.module.scss';
+import { UniProtkbAPIModel } from '../../../uniprotkb/adapters/uniProtkbConverter';
```
dlrice, line 42: Just needs to be move a bit further up in the import list.

dlrice (approved):

PR #773, opened 2024-01-18
Api urls refactor

dlrice:
## Purpose
Refactor api urls

## Approach

- `shared/apiUrls.ts` turned into a directory `shared/apiUrls` with structured contents:
```
apiPrefix.ts
apiUrls.ts
configure.ts
entry.ts
proteinsApi.ts
releaseNotes.ts
results.ts
search.ts
suggester.ts
```
- `apiUrls.ts` now imports from these files and re-exports
- Moved anything from there that doesn't return a url
- If a fn was only used in part of the site move it to its own `apiUrls` directory or file

## Testing
Updated

## Checklist


src/shared/config/apiUrls/apiUrls.ts
```diff
@@ -0,0 +1,17 @@
+import * as configure from './configure';
```
aurel-l, line 1: Have you thought about naming this file `index.ts`? That way an import of this file would just be ...../apiUrls instead of ....../apiUrls/apiUrls
That being said, it's node module resolution logic, that is handled through tooling (Webpack/TypeScript) and we're going further than the standard browser module resolution logic that expects full file names (with extensions)

dlrice: I did consider this but my thinking was to keep the same `apiURls.ts` file name throughout the codebase (some namespaces apiUrls are just files directly under /config).

aurel-l (approved):

PR #781, opened 2024-02-13
2023 02 entry consistency

aurel-l:
## Purpose
Have the top part of entries be more consistent across the 4 main namespaces. [TRM-30784](https://www.ebi.ac.uk/panda/jira/browse/TRM-30784)

## Approach
Use the InfoList component, outside of a card, above any possible tab.
Specific cases:
- In UniRef, show the common taxon in the new overview section (was in the data but not displayed before)
- For UniParc, no change so far: Not enough data to show anything. I've tried length and number of crossrefs but the InfoList with 2 bits of data looked strange. In the future when we get the common taxon data then we might have enough data to display there.

Also:
- Fix incorrect heading level on UniProtKB entry page (the title was in H3 instead of H1)
- Clean some logic that I found related to the isoform-specific entry pages that were discontinued
- Transform some CSS files to use CSS modules
- Make some Sequence component render the sequence or not based on:
  - In UniProtKB, hide the canonical by default if the sequence is huge for performance reasons.
  - In UniParc and UniRef display the sequence if it is small enough not to hide below the fold the rest of the page. 2 hardcoded thresholds based on screen size.

## Testing
Updated snapshots + manual comparisons
Specific cases:
- UniRef UniRef50_P00395: the common taxon is the root (no link because there's no corresponding page)
- Proteome UP000094546: no description
- Proteome UP000242668: redundant proteome, show link for redundant to which other proteome

## Checklist


src/uniref/components/entry/SequenceSection.tsx
```diff
@@ -1,12 +1,16 @@
 import { Card, LongNumber, Sequence } from 'franklin-sites';
```
Swaathik, line 1: The SequenceSection.tsx is the same for both UniParc and UniRef. We have them one each instead of a shared one because of the different data type?

Swaathik (approved):

src/uniref/components/entry/SequenceSection.tsx
aurel-l: I created a shared component wrapping the call to Franklin's sequence component that generates the correct infolist and also handles the default rendering of sequence depending on sequence length and screen size.

Also, I've taken the opportunity of reviewing usage of the sequence component to also not render by default the sequence in search results if the length is too big

Swaathik (approved):

PR #782, opened 2024-02-13
Isoform links

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-28186
https://www.ebi.ac.uk/panda/jira/browse/TRM-29442

## Approach
JIRA 1 - The element IDs are written with an underscore replacing the space as per the naming convention.
JIRA 2 - In order to link only the isoforms present in the data, it is passed from the top level where data is converted for relevant sections. to RichText. Because of that. I had to exclude it while checking 'hasContent' in the entry page for disabling section. 

## Testing
Updated the tests

## Checklist


src/uniprotkb/components/entry/FunctionSection.tsx
```diff
@@ -156,9 +156,14 @@ const BioPhysicoChemicalPropertiesView = ({
 type CofactorViewProps = {
   cofactors?: CofactorComment[];
   title?: string;
+  isoforms?: string[] | undefined;
```
aurel-l, line 159:
```suggestion
  isoforms?: string[];
```

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -49,9 +49,18 @@ type RichTextProps = {
    * Just do enrichments that don't involve adding links. Only the visual ones
    */
   noLink?: boolean;
+  /**
+   * Link isoforms that exist in the data
+   */
+  isoforms?: string[] | undefined;
```
aurel-l, line 55:
```suggestion
  isoforms?: string[];
```

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -164,9 +178,15 @@ type TextViewProps = {
   comments: TextWithEvidence[];
   type?: FreeTextType;
   children?: ReactNode;
+  isoforms?: string[] | undefined;
```
aurel-l, line 181:
```suggestion
  isoforms?: string[];
```

src/uniprotkb/components/protein-data-views/FreeTextView.tsx
```diff
@@ -188,13 +210,15 @@ type FreeTextProps = {
   title?: ReactNode;
   articleId?: string;
   showMolecule?: boolean;
+  isoforms?: string[] | undefined;
```
aurel-l, line 213:
```suggestion
  isoforms?: string[];
```

src/uniprotkb/components/protein-data-views/UniProtKBFeaturesView.tsx
```diff
@@ -77,6 +75,7 @@ type FeatureProps = {
   features: FeatureDatum[];
   inResultsTable?: boolean;
   showSourceColumn?: boolean;
+  isoforms?: string[] | undefined;
```
aurel-l, line 78:
```suggestion
  isoforms?: string[];
```

src/shared/utils/utils.tsx
```diff
@@ -33,8 +33,16 @@ export function removeItemFromList<T>(list: T[], index: number) {
...
+export const hasContent = (
+  obj: Record<string | number | symbol, unknown>,
+  exclude?: string
+) => {
+  // If certain property has to be excluded while checking for content (like section specific)
+  const objCopy = { ...obj };
+  if (exclude && objCopy[exclude]) {
+    delete objCopy[exclude];
+  }
+  return Object.values(objCopy).some((val) => {
```
aurel-l, line 45:
```suggestion
  // If certain property has to be excluded while checking for content (like section specific)
  const objWithoutField = { [exclude]: _, ...obj };
  return Object.values(objWithoutField).some((val) => {
```

aurel-l, line 45:
```suggestion
  return Object.entries(obj).some(([key, val]) => {
    // If certain property has to be excluded while checking for content (like section specific)
    if (key === exclude) {
      return false;
    }
```
Maybe, reading again the code, the intent might be clearer filtering this directly in the callback for `.some()`

aurel-l (approved):

PR #783, opened 2024-02-19
Include Community annotation

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-30488

## Approach
A new component is created to show the annotations and it is displayed at the respective sections

## Testing
Added a test

## Checklist


src/uniprotkb/components/entry/CommunityCuration.tsx
```diff
@@ -0,0 +1,111 @@
...
+            variant="tertiary"
+            className={styles['community-curation-button']}
+            onClick={() => setToggleView(!toggleView)}
+          >
+            <CommunityAnnotationIcon />
+            {`Community curation (${communityReferences.length}) `}
+            {toggleView ? (
+              <ChevronUpIcon width="1ch" />
+            ) : (
+              <ChevronDownIcon width="1ch" />
```
aurel-l, line 52: The only place where you'd need a state would be here, for styling, so as a workaround as the open/closed state is on the html details element then you could just render one chevron, and transform it to flip it in CSS with the selector `[open]` for the parent details element

src/uniprotkb/components/entry/CommunityCuration.tsx
```diff
@@ -0,0 +1,132 @@
...
+                      {communityAnnotation?.disease}
+                    </p>
+                  )}
+                  {communityAnnotation?.comment && (
+                    <p>{communityAnnotation?.comment}</p>
+                  )}
+                  <div className={styles['contributor-details']}>
+                    {citationId && (
+                      <span>
+                        Source:&nbsp;&nbsp;
```
aurel-l, line 84: Any reason for the double nbsp here and the same a bit below? Is that just for a bit more spacing?

aurel-l (approved):

src/uniprotkb/components/entry/CommunityCuration.tsx
```diff
@@ -39,81 +39,79 @@ const CommunityCuration = ({
...
-            <hr className={styles.separator} />
-          </summary>
-          <HeroContainer className={styles.content}>
-            {communityReferences.map(
-              ({ source, communityAnnotation, citationId }) => (
-                <div key={source?.id}>
-                  {communityAnnotation?.proteinOrGene && (
-                    <b>{communityAnnotation.proteinOrGene}</b>
+      <details className={styles['community-annotation-details']}>
+        <summary className={styles.header}>
```
aurel-l, line 43: Is that class still needed?

Swaathik: you mean the 'header' class?

aurel-l (approved):

PR #787, opened 2024-03-11
FASTA subsequence download in Basket

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-30836

## Approach
If 'SubsequencetoAccessionMap' is present, the accessions are either taken with or without subsequence depending on the selected file format.

The download/preview count and the number of entries in the basket page are all in sync now. 
Removed `notCustomisable` from BasketMiniView and used the 'inBasketMini' to determine the position of sliding panel in BasketMiniView. This also fixes the count of `selected` and `all` when downloaded from basketMiniView instead of fixed total number which is wrong.

## Testing
Added a test for subsequence download from basket

## Checklist

### DIFF
--- a/src/shared/components/download/downloadUtils.ts
+++ b/src/shared/components/download/downloadUtils.ts
@@ -98,10 +98,9 @@ export const getPreviewFileFormat = (
 
 export const getCountForCustomisableSet = (
   state: DownloadState,
-  props: DownloadProps<JobTypes>,
-  totalNumberResults: number
+  props: DownloadProps<JobTypes>
 ) => {
-  let totalCount = totalNumberResults;
+  let totalCount = props.totalNumberResults;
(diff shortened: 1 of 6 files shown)


src/shared/components/download/downloadUtils.ts
```diff
@@ -91,13 +96,55 @@ export const getPreviewFileFormat = (
   return state.selectedFileFormat;
 };
 
+export const getCountForCustomisableSet = (
+  state: DownloadState,
+  props: DownloadProps<JobTypes>,
+  totalNumberResults: number
```
dlrice, line 102:
```suggestion
```
`totalNumberResults` should be available from props

dlrice (approved):

PR #794, opened 2024-04-22
GA4 tracking for job submission

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-31093

## Approach
Added a new tag and passing the relevant parameters if necessary

## Testing

## Checklist

### DIFF
--- a/src/shared/utils/gtagEvents.ts
+++ b/src/shared/utils/gtagEvents.ts
@@ -3,6 +3,7 @@ import { SlidingPanel } from 'franklin-sites';
 
 import { Column } from '../config/columns';
 import { ViewMode } from '../hooks/useViewMode';
+import { JobTypes } from '../../tools/types/toolsJobTypes';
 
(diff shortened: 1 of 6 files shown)


dlrice (approved):

src/shared/utils/gtagEvents.ts
```diff
@@ -194,3 +196,10 @@ export const sendGtagEventCacheUpdate = (
     cache_name: cacheName,
   });
 };
+
+export const sendGtagEventJobSubmit = (
+  job: JobTypes,
+  parameters?: Record<string, string>
+) => {
+  sendGtagEvent('job_submit', parameters);
```
aurel-l, line 204: I notice now that `job` is not used here, it might be good to add it to the parameters so that we could separate events based on this dimension.
It might be good to call this `job_type` when sending to GA

PR #795, opened 2024-05-08
More feature tab downloads

dlrice:
## Purpose
[Add more download sources for the feature viewer download panel](https://www.ebi.ac.uk/panda/jira/browse/TRM-31104)

## Approach
- Show interpro representative domains if present
- Download alphafold from prediction endpoint eg https://alphafold.ebi.ac.uk/api/prediction/P05067
- Grab the urls from the json
- Modify the cif url to get confidence url
- Pass uniprot sequence to entry download to compare to alphafold's and if different don't offer downloads
- Display alphafold files in both main, variant viewer and feature viewer tab (as we do with the other additional file formats)

## Testing
None

## Checklist

### DIFF
--- a/src/shared/components/entry/EntryDownload.tsx
+++ b/src/shared/components/entry/EntryDownload.tsx
@@ -431,8 +431,8 @@ const EntryDownload = ({
       : undefined;
 
   if (alphaFoldUrls) {
-    availableDatasets.push(Dataset.alphaFoldConfidence);
     availableDatasets.push(Dataset.alphaFoldCoordinates);
+    availableDatasets.push(Dataset.alphaFoldConfidence);
   }
 
   if (!entryFeatures.loading && entryFeatures.data) {
--- a/src/shared/components/entry/EntryDownload.tsx
+++ b/src/shared/components/entry/EntryDownload.tsx
@@ -298,6 +298,7 @@ export type EntryDownloadProps = {
   columns?: Column[];
   dataset?: Dataset;
   featureTypes?: string[];
+  sequence?: string;
 };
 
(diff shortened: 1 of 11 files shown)


aurel-l (approved):

src/shared/components/entry/EntryDownload.tsx
```diff
@@ -279,46 +379,72 @@ const EntryDownload = ({
...
+  const alphaFoldUrls =
+    // As there can be a build mismatch only use AlphaFold predictions if the sequence is the same as the entry's
+    sequence && sequence === alphaFoldPrediction?.data?.[0].uniprotSequence
+      ? getAlphaFoldUrls(alphaFoldPrediction?.data)
+      : undefined;
+
+  if (alphaFoldUrls) {
+    availableDatasets.push(Dataset.alphaFoldCoordinates);
+    availableDatasets.push(Dataset.alphaFoldConfidence);
+    availableDatasets.push(Dataset.alphaMissenseAnnotations);
```
aurel-l, line 445: There's an example that someone from Ensembl found _with_ AF, but _without_ AM: https://alphafold.ebi.ac.uk/api/prediction/A0A2U3U0J3
Have a look if you could add an extra check for that. Not sure if it's expected or not, PDBe will look into it, but now we see that it _could_ happen

dlrice: That's a good example because it also has these other fields I'm not considering:
```
"amAnnotationsHg19Url": null,
"amAnnotationsHg38Url": null
```
Add these as options too you think?

aurel-l (approved):

PR #800, opened 2024-06-13
Xrefs with colon suggestion

dlrice:
## Purpose
[Website suggestions when user searches for IDs with colon](https://www.ebi.ac.uk/panda/jira/browse/TRM-31343)

## Approach
Detect when all of these are true in the results:

1. There is a colon in the query
2. The API returned an error message like `'PTHR34313' is not a valid search field`
3. In uniprotkb

and suggest to the other an escaped version of their query.

## Testing
Yes

## Checklist


src/shared/components/results/Results.tsx
```diff
@@ -72,6 +77,36 @@ const Results = () => {
 
   const [params] = getParamsFromURL(search);
 
+  // Some cross references legitimately have colons in their ids eg PTHR34313:SF2
+  // The API returns "'PTHR34313' is not a valid search field" on those occassions
+  // so to assist the users, escape the colon PTHR34313\:SF2 and try a head request.
```
aurel-l, line 82: You would also need to check the number of results.
Even for a search returning 0 results the API returns 200 and an empty payload, see search for "PTHR:34313SF2" for example

src/shared/components/results/Results.tsx
```diff
@@ -72,6 +77,36 @@ const Results = () => {
...
+        }
+      : undefined
+  );
+  const escapedColonQueryData = useDataApi(
+    invalidSearchFieldQueryWithColon ? escapedColonQueryUrl : null,
+    {
+      method: 'HEAD',
+    }
+  );
+  const escapedColonQueryTotal =
```
aurel-l, line 107: Need to parse this as a number, the header will be string values, so you could get "0" and lower in l.161 this would be truthy

aurel-l (approved):

PR #801, opened 2024-06-21
uniparc landing page

aurel-l:
## Purpose
Add a landing page for UniParc [TRM-31489](https://www.ebi.ac.uk/panda/jira/browse/TRM-31489)

## Approach
Copy and adapt the UniProtKB landing page

Some liberties taken wrt mockup in order to be more consistent with casing and with the naming of sections across landing pages as well as only displaying info available at the moment

Table of data expandible into a "responsive" (mix CSS and JS logic) multi-column table

## Testing
Manual tests

## Checklist


src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,291 @@
...
+              </ExternalLink>
+              <ExternalLink url="https://ftp.uniprot.org/pub/databases/uniprot/current_release/uniparc/README">
+                readme
+              </ExternalLink>
+            </p>
+            <p>
+              XML
+              <br />
+              <ExternalLink url="https://ftp.uniprot.org/pub/databases/uniprot/current_release/uniparc/uniparc.xsd">
+                xml schema
```
dlrice, line 246: I'm not sure if the inclusion of the XML schema is needed here? At first glance it confused me a bit as I was under the impression that it was the actual XML file download which is actually listed above. We have quite a lot of space so not sure what you think about a traditional list? Eg

```
UniParc sequence archive
- xml (schema)
- fasta
- readme

Explore more FTP
```

I know it breaks with the uniprotkb landing page but for me it organizes the links a bit better.

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,291 @@
...
+          className={cn(
+            'uniprot-grid-cell--small-span-12',
+            `uniprot-grid-cell--medium-span-${expanded ? '9' : '5'}`
+          )}
+        >
+          <h2>Searching the Database</h2>
+          <br />
+          <p>
+            UniParc is available for text- and sequence-based searches.
+            Performing a similarity search against UniParc is equivalent to
```
dlrice, line 205:
```suggestion
            Performing a sequence similarity search against UniParc is equivalent to
```

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,291 @@
...
+        <section
+          className={cn(
+            'uniprot-grid-cell--small-span-12',
+            `uniprot-grid-cell--medium-span-${expanded ? '9' : '5'}`
+          )}
+        >
+          <h2>Searching the Database</h2>
+          <br />
+          <p>
+            UniParc is available for text- and sequence-based searches.
```
dlrice, line 204:
```suggestion
            UniParc provides text- and sequence-based searches.
```

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,291 @@
...
+            Performing a similarity search against UniParc is equivalent to
+            performing the same search against all databases cross-referenced in
+            UniParc, as UniParc contains all proteins from its source databases.
+          </p>
+          <p>
+            Sequence similarity searches can be done using{' '}
```
dlrice, line 210: Just an idea here. The landing page seems like a place to list activities the user can try. The `»` in "Start searching in UniParc »" makes me think "go off and try this out". What about here we change the wording a bit:
```
Start a sequence similarity search in UniParc using BLAST »
```

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,291 @@
...
+      start += countByColumn;
+    }
+  }
+
+  return (
+    <div className={styles['landing-page']}>
+      <HTMLHead title="UniProt sequence archive (UniParc)">
+        <meta name="description" content={metaDescription} />
+      </HTMLHead>
+      <section className="uniprot-grid">
```
dlrice, line 94: As we have a decent amount of whitespace on the page what do you think about increase the grid gap to 2rem just to make clearer the different sections of the page?

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,291 @@
...
+        </div>
+
+        {/* Table */}
+        <section
+          className={cn(
+            'uniprot-grid-cell--small-span-12',
+            `uniprot-grid-cell--medium-span-${expanded ? '12' : '4'}`
+          )}
+        >
+          <h2>Top Databases</h2>
```
dlrice, line 145: This seems to be all of the databases. Maybe just `Databases`? Or the more verbose `Cross-referenced sequence databases` to try to make it as clear as possible what they are?

src/uniparc/components/landing-page/styles/landing-page.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  th,
+  td {
+    padding: $global-padding * 0.25 $global-padding * 0.5;
+    border: 1px solid $colour-sea-blue;
+    border-collapse: collapse;
+  }
+}
+
+.table-container {
+  display: grid;
```
dlrice, line 47:
```suggestion
  margin-top: 2rem;
  display: grid;
```
To line up the table with the text in the `Searching the Database` and `Download` text to the right.

dlrice (commented):
Looks good - just some UX suggestions.

src/uniparc/components/landing-page/LandingPage.tsx
aurel-l: For that I'm not too keen, because it's using the franklin-defined grid and it would mean the grid is inconsistent here compared to the rest of the website.
Hopefully we get more content in the future in order to fill this page

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,292 @@
...
+                </tbody>
+              </table>
+            ))}
+          </div>
+
+          <Button
+            variant="tertiary"
+            onClick={() => setExpanded((expanded) => !expanded)}
+          >
+            {expanded ? 'View less' : 'View more'} databases
```
dlrice, line 191:
```suggestion
            {expanded ? 'View fewer' : 'View more'} databases
```
I only learnt recently about less vs fewer the other day. Less is for uncountable things while fewer is for countable things. "Less milk, fewer cookies."

src/uniparc/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,292 @@
...
+            </Link>
+          </p>
+        </section>
+
+        {/* Downloads */}
+        <section className="uniprot-grid-cell--small-span-12 uniprot-grid-cell--medium-span-3">
+          <h2>Download</h2>
+          <div className={styles.download}>
+            <br />
+            <p>
```
dlrice, line 229: Seeing a `<ul> cannot appear as a descendant of <p>.` error so I guess just changing this to `<div>` would solve it but there is a little spacing difference seen in the second screenshot.
Screenshot 2024-07-23 at 15 14 16
Screenshot 2024-07-23 at 15 14 07

dlrice (approved):
Two minor suggestions. Good to go for merging once resolved.

src/uniparc/components/landing-page/LandingPage.tsx
aurel-l: Ah, right, so we should also probably do that change to the other places where we have that more/less pattern (I think in the ExpandableMessage in Franklin)

src/uniparc/components/landing-page/LandingPage.tsx
aurel-l: In the end I just replaced with the div, it took me a bit to see the difference so I'm assuming a user wouldn't be too bothered

src/uniparc/components/landing-page/LandingPage.tsx
dlrice: Good point. Basically:
```
diff --git a/src/components/expandable-list.tsx b/src/components/expandable-list.tsx
index ab60aa40..9fb07d53 100644
--- a/src/components/expandable-list.tsx
+++ b/src/components/expandable-list.tsx
@@ -20,7 +20,7 @@ export const ExpandableMessage: FC<ExpandableMessageProps> = ({
   showHideWording,
   nHiddenItems,
 }) => {
-  let message = `${showHideWording ? 'Hide' : 'Less'} ${descriptionString}`;
+  let message = `${showHideWording ? 'Hide' : 'Fewer'} ${descriptionString}`;
   if (!expanded) {
     if (nHiddenItems === undefined) {
       message = showHideWording ? 'Show' : 'More';
```
I'm happy to push and release if this works for you?

aurel-l: Yep, thanks!

dlrice: Done!

PR #807, opened 2024-07-29
Remove protvista-track from busco viz

dlrice:
## Purpose
[Remove protvista-track from busco viz](https://www.ebi.ac.uk/panda/jira/browse/TRM-31653)

## Approach
Use css gradients and custom properties

## Testing
Updated mocks

## Checklist


src/proteomes/components/styles/busco-colors.scss
```diff
@@ -0,0 +1,4 @@
+$s-color: #bce875; // completeSingle
+$d-color: #0ba5a6; // completeDuplicated
+$f-color: #eecba4; // fragmented
+$m-color: #314bb4; // missing
```
aurel-l, line 4: To reduce dependency on SASS it might be good to use CSS Custom Properties here.
You might put them into the `:root` selector and make sure to import this file from both CSS files where these are used.
Or you could merge both CSS files and define them there. These could be merged because although labels and graph might be used independently in the page, they're appearing on the sames pages so it won't be loading extra CSS unnecessarily.

aurel-l:
Could we make the busco wrapper (`.busco`) take `calc(100% - 1rem)` or `calc(100% - #{$global-margin})`  in order to match the alignement of the card below?

dlrice:
- Addressed comments
- Removed `useMemo`
- Slightly refactored percentage calculations

src/proteomes/components/styles/busco.module.scss
```diff
@@ -0,0 +1,67 @@
+:root {
+  --s-color: #bce875; // completeSingle
+  --d-color: #0ba5a6; // completeDuplicated
+  --f-color: #eecba4; // fragmented
+  --m-color: #314bb4; // missing
```
aurel-l, line 5: Ah, just a note, if this is set on `:root` maybe use a more specific name, containing `busco` for example, as this will be global for the whole document and it might be used elsewhere.
Otherwise if it's all in the same file you could do:
```css
.busco-view, .busco-legend-item {
  --s-color: #bce875; // completeSingle
  --d-color: #0ba5a6; // completeDuplicated
  --f-color: #eecba4; // fragmented
  --m-color: #314bb4; // missing
}
```
and this way they will be narrowly scoped

PR #814, opened 2024-09-24
Proteome landing page

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-31490

## Approach

## Testing

## Checklist


src/proteomes/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,365 @@
...
+import { FacetsEnum } from '../../config/ProteomesFacetConfiguration';
+
+import styles from './styles/landing-page.module.scss';
+
+import SpeciesIllustration from '../../../images/species_illustration.img.svg';
+
+const documentationLinks = [
+  {
+    label: 'Proteome',
+    id: 'roteome',
```
aurel-l, line 33:
```suggestion
    id: 'proteome',
```

src/proteomes/components/landing-page/LandingPage.tsx
```diff
@@ -0,0 +1,365 @@
...
+              <br />
+              <p
+                className={styles['stats-count']}
+                onPointerEnter={() => setRefProtHovered(true)}
+                onFocus={() => setRefProtHovered(true)}
+                onPointerLeave={() => setRefProtHovered(false)}
+                onBlur={() => setRefProtHovered(false)}
+              >
+                <ReferenceProteomeIcon
+                  width="2em"
```
aurel-l, line 206: This appears to be a different size. Maybe because the underlying SVG has a different size so might need to adjust the viewbox, or maybe try a different unit for width (ch, rem?), or adjust one of the values in order to make them visually the same

src/proteomes/components/landing-page/styles/landing-page.module.scss
```diff
@@ -0,0 +1,68 @@
...
+  flex-direction: column;
+
+  p.stats-count {
+    display: flex;
+    gap: $global-margin;
+  }
+}
+
+.ref-prot-icon {
+  color: $colour-reference-proteome;
```
aurel-l, line 39: Try to use custom properties instead of SASS variables from now on. At the moment we have colours available, you can see them here https://github.com/ebi-uniprot/franklin-sites/blob/main/src/styles/_colours.scss#L58-L110

```suggestion
  color: var(--fr--color-reference-proteome);
```

src/proteomes/components/landing-page/styles/landing-page.module.scss
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.unreviewed-icon {
+  color: $colour-unreviewed;
```
aurel-l, line 43:
```suggestion
  color: var(--fr--color-unreviewed);
```

src/proteomes/components/landing-page/styles/landing-page.module.scss
```diff
@@ -0,0 +1,68 @@
+@import 'franklin-sites/src/styles/settings';
+@import 'franklin-sites/src/styles/colours';
```
aurel-l, line 2: Following replacing SASS variables with CSS Custom Properties, we can remove this import as the colours are set globally

aurel-l (changes requested):

src/proteomes/components/landing-page/styles/landing-page.module.scss
Swaathik: I am still using $global-margin from settings. Shall I keep it as it is or set the size?

aurel-l: For now we don't have Custom Properties for sizing (yet), so you can only remove `colours`, but you'll need to keep `settings` because of that global margin variable

aurel-l (approved):

PR #818, opened 2024-10-10
Update feedback button and re-do scroll bar fix

minjoonkim:
## Purpose
HotJar has combined feedbacks into surveys, which has changed the button for the website feedback. 
This needs a different approach to moving the button to not cover the sidebar, and to also trigger the fallback survey if the hotjar button is not present.
https://www.ebi.ac.uk/panda/jira/browse/TRM-31062

## Approach
The change in hotjar button requires a different way to target the survey and also trigger the fallback google form.

## Testing
Checked on website to confirm changes, with multiple sessions

## Checklist


src/help/components/contextual/SideButtons.tsx
```diff
@@ -17,40 +17,36 @@ const SideButtons = ({ displayHelp, onClick }: Props) => {
   const [displayFeedback, setDisplayFeedback] = useState(false);
 
   useEffect(() => {
-    // Checking if there is a scroll bar
+    // // Checking if there is a scroll bar
```
dlrice, line 20:
```suggestion
    // Checking if there is a scroll bar
```

src/help/components/contextual/SideButtons.tsx
```diff
@@ -17,40 +17,36 @@ const SideButtons = ({ displayHelp, onClick }: Props) => {
...
-      }
-      if (sideButtonHelp) {
-        sideButtonHelp.style.right = `${scrollBarWidth.toString()}px`;
-      }
     }
 
     sleep(3000).then(() => {
       // If there's already Hotjar's feedback, don't do anything
-      if (document.querySelector('._hj_feedback_container')) {
+      if (document.querySelector('#survey_1067707')) {
```
dlrice, line 37: As `#survey_1067707` is located in a few places, it could be easier to update in the future by creating a variable and using this instead eg:

```
const surveyId = '#survey_1067707'
```

src/help/components/contextual/SideButtons.tsx
```diff
@@ -17,40 +17,36 @@ const SideButtons = ({ displayHelp, onClick }: Props) => {
...
         if (scrollBarWidth) {
           const hjButton = document.querySelector<HTMLElement>(
-            '#_hj_feedback_container div button'
+            '#survey_1067707 > div > div > div'
```
dlrice, line 40:
```suggestion
            `${surveyId} > div > div > div`
```
If the other suggestion is used then this would line would want to change to something similar to this.

PR #819, opened 2024-10-14
Use the new endpoint for combined UniProtKB stats

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-30436

## Approach
Pass the new data instead of reviewed and unreviewed sum

## Testing

## Checklist

### DIFF
--- a/src/uniprotkb/components/statistics/utils.ts
+++ b/src/uniprotkb/components/statistics/utils.ts
@@ -55,20 +55,12 @@ export const mergeToMap = (
     accumulator.set(item.name, { uniprotkb: item });
   }
   for (const item of reviewed) {
-    const stats = accumulator.get(item.name);
-    if (stats) {
-      stats.reviewed = item;
-    } else {
-      accumulator.set(item.name, { reviewed: item });
-    }
+    const stats = accumulator.get(item.name) || {};
+    accumulator.set(item.name, { ...stats, reviewed: item });
   }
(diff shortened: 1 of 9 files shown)


src/uniprotkb/components/statistics/utils.ts
```diff
@@ -38,19 +38,29 @@ export type MergedStatisticsItem = {
...
   reviewed: StatisticsItem[],
   unreviewed: StatisticsItem[]
 ) => {
   const accumulator = new Map<string, MergedStatistics>();
+  for (const item of uniprotkb) {
+    accumulator.set(item.name, { uniprotkb: item });
+  }
   for (const item of reviewed) {
-    accumulator.set(item.name, { reviewed: item });
+    const stats = accumulator.get(item.name);
```
dlrice, line 58:
```
  for (const item of reviewed) {
    const stats = accumulator.get(item.name) || {};
    accumulator.set(item.name, { ...stats, reviewed: item });
  }
  for (const item of unreviewed) {
    const stats = accumulator.get(item.name) || {};
    accumulator.set(item.name, { ...stats, unreviewed: item });
  }
```
I had to think about what is happening when we do `stats.reviewed = item`. I find the above just a little clearer as I know where the `stats.reviewed` will end up. What do you think?

dlrice (approved):

PR #820, opened 2024-10-21
Nightingale upgrade

dlrice:
## Notes

- Apologies but this PR includes changes from the tooltip PR which as already been approved. Somehow I can't disregard those changes here.
- Some delay on the MSA loading in alignment overview. I will continue to look at this as I assume the change will want to be made in the nightingale codebase.

## Purpose
Upgrade to the latest nightingale v5

## Approach

- Upgrade protvista packages to the latest nightingale
  - Feature.accession now acts as the protvista feature ID
  - displaystart --> display-start / displayend --> display-end
- Replace `protvista-datatable` with new React component `TableFromData`.
  - This has the consumer provide a column configuration similar to our results tables.
  - Table rows which are in range or are active are now managed using props `markBackground` and `markBorder`
- Incorporate tooltips everywhere using the new `tooltip.ts` functions which are based on floating-ui. 
- Update to latest protvista-uniprot and adapt filtering logic
- Saw repeated attempts to reregister nightingale components so adapted the global `customElements.define` to make sure this doesn't happen. I did spend some time trying to figure out why this was happening but couldn't find the source.

## Follow on tasks
- [Extract 3D structures table from protvista-uniprot into uniprot website](https://www.ebi.ac.uk/panda/jira/browse/TRM-31965)
- [useDeferredValue for nightingale + table interaction](https://www.ebi.ac.uk/panda/jira/browse/TRM-31984)
- [Reimplement filter-scroll](https://www.ebi.ac.uk/panda/jira/browse/TRM-32095)

## Testing
Added and updated.

## Checklist


Swaathik:
Observed few things while testing the code

- Out of Nightingale context, there is a `<div>` injected in the tooltip content
Screenshot 2024-10-22 at 13 29 30

- This is most likely from the past. There is inconsistency in the line gap between 'Description' and 'Evidence' and also Peptidoform and its content. Accession referred is P05067. Attaching screenshots for reference
Screenshot 2024-10-22 at 13 16 29
Screenshot 2024-10-22 at 13 16 04

- I come across this runtime error due to resize observer. Is it something that was recently addressed in Nightingale?
Screenshot 2024-10-22 at 13 17 03.

src/uniprotkb/components/protein-data-views/VisualVariationView.tsx
```diff
@@ -20,96 +24,50 @@ type VariationViewProps = {
...
+      <NightingaleNavigationComponent
+        ref={navigationRef}
+        length={sequence.length}
+        height={40}
+      />
+      <NightingaleSequenceComponent
+        length={sequence.length}
+        sequence={sequence}
+        height={20}
+        filter-scroll
```
Swaathik, line 55: `filter-scroll` behaviour is not captured in the new version - https://github.com/ebi-webcomponents/nightingale/blob/d7b2f1a01554087e840a53465bc2310c2756ce6e/packages/nightingale-new-core/src/mixins/withZoom/index.ts#L92.

What is it supposed to do?

src/uniprotkb/components/protein-data-views/VisualVariationView.tsx
```diff
@@ -20,96 +24,50 @@ type VariationViewProps = {
...
+      <NightingaleFilterComponent
+        for="variation-component"
+        filters={filterConfig}
+      />
+      <NightingaleVariationComponent
+        id="variation-component"
+        length={sequence.length}
+        height={430}
+        ref={setNightingaleVariation}
+        no-scroll
```
Swaathik, line 68: What is the purpose of `no-scroll`? There is no usage of it in Nightingale v5

src/uniprotkb/components/entry/tabs/FeatureViewer.tsx
```diff
@@ -119,9 +118,12 @@ const FeatureViewer = ({
       )}
       {data?.features && (
         <>
-          {shouldRender && (
-            <NightingaleZoomTool length={sequence.length} onZoom={handleZoom} />
-          )}
+          {/* {shouldRender && (
+            <NightingaleZoomTool
```
Swaathik, line 122: Is it removed temporarily?

src/app/components/App.tsx
```diff
@@ -43,6 +43,20 @@ import pkg from '../../../package.json';
 import 'franklin-sites/franklin.css';
 import './styles/app.scss';
 
+// This is hackery is to prevent define being repeatedly called for the same
```
Swaathik, line 46: I was curious to try without the hack and I didn't run into any problem of defining the custom element twice

dlrice: Oh that's interesting - no issue in the feature viewer tab either?

Swaathik: When I switched from Feature viewer to overview/variant viewer tab, it happens and crashes too. It is needed after all

src/shared/components/table/styles/table.module.scss
```diff
@@ -10,12 +20,13 @@
   margin: 1em 0;
   border-spacing: 0;
   border-block-end: 1px solid #c2c4c4;
+  border-collapse: collapse;
 
   & th {
```
Swaathik, line 25: In the current version, the table header is sticky. 

  & thead {
    position: sticky;
    top: 0;
    z-index: 1;
  }
  
With the border-collapse: separate - it looks neat if we have it sticky.

src/shared/utils/tooltip.ts
```diff
@@ -163,18 +166,22 @@ export const showTooltipAtCoordinates = (
...
 
-  function hideTooltip() {
-    tooltip?.remove();
-    cleanup?.();
-    document.body.removeEventListener('click', onClick, true);
-    document.body.removeEventListener('scroll', hideTooltip, true);
-    document.body.removeEventListener('wheel', hideTooltip, true);
+  function hideTooltip(e?: Event) {
+    const target = e?.target as Node;
+    // f scroll/wheel is within the tooltip element, do not remove the tooltip
```
dlrice, line 171: *If

src/shared/utils/tooltip.ts
```diff
@@ -163,18 +166,22 @@ export const showTooltipAtCoordinates = (
...
   document.body.addEventListener('wheel', hideTooltip, true);
 
-  function hideTooltip() {
-    tooltip?.remove();
-    cleanup?.();
-    document.body.removeEventListener('click', onClick, true);
-    document.body.removeEventListener('scroll', hideTooltip, true);
-    document.body.removeEventListener('wheel', hideTooltip, true);
+  function hideTooltip(e?: Event) {
+    const target = e?.target as Node;
```
dlrice, line 170:
```
const target = e?.target as Node | null;
```
This doesn't really change as you're already checking for `target`'s existence later. Suggesting just for consistency.

src/shared/utils/tooltip.ts
```diff
@@ -163,18 +166,22 @@ export const showTooltipAtCoordinates = (
...
+    // f scroll/wheel is within the tooltip element, do not remove the tooltip
+    if (target && !tooltip.contains(target)) {
```
dlrice, line 172: This seems duplicated by L183?

Swaathik: that' right. There is no need for two separate functions. hideTooltip is complete. Do you have any suggestion for a name of the event handler?

PR #823, opened 2024-10-24
Go cam viz

dlrice:
## Note

- Delay merge until start of February when new `wc-gocam-viz` released which addresses many issues.
- Once merged delete this out of date manual file: [gene_ontology](https://github.com/ebi-uniprot/uniprot-manual/blob/main/help/gene_ontology.md)

## Purpose
[Add wc-gocam-viz to uniprotkb entry page.](https://www.ebi.ac.uk/panda/jira/browse/TRM-31564)

## Approach

- Request gocam ids from api.geneontology.org
- If we have some then show a tab with Go-Cam viz
- Use `TreeSelect` for user selection of go-cam id
- Also use `TreeSelect` for Complex Viewer ui

## Testing
Updated existing test.

## Checklist


src/uniprotkb/components/entry/FunctionSection.tsx
```diff
@@ -376,6 +376,13 @@ const FunctionSection = ({
           />
         </Suspense>
       </ErrorBoundary>
+      <ErrorBoundary>
+        <Tabs>
```
aurel-l, line 380: I think you have Tabs here, but also again within GoCam where you have Tabs and a Tab

src/uniprotkb/components/entry/FunctionSection.tsx
```diff
@@ -376,6 +376,13 @@ const FunctionSection = ({
...
+          <Suspense fallback={<Loader />}>
+            <GoCam primaryAccession={primaryAccession} />
```
aurel-l, line 382: Not sure that you'd need a Suspense here if GoCam won't be lazy loaded

src/uniprotkb/components/entry/GoCam.tsx
```diff
@@ -0,0 +1,104 @@
...
+            data={goCamItems}
+            label="Select GO-CAM model"
+            onSelect={(item) => {
+              setSelectedId(item.id);
+            }}
+            defaultActiveNodes={[selectedId]}
+          />
+          <LazyComponent>
+            <GoCamViz id={selectedId} />
+          </LazyComponent>
```
aurel-l, line 92: LazyComponent will prevent rendering until the user gets there, but having GoCamViz's import be a dynamic import within a `lazy()` would allow to have a code split and not load any of that logic until it's actually needed

aurel-l (changes requested):
Maybe some things to check regarding nesting of components and lazy-loading of logic, and to keep it consistent with the ComplexViewer in the interaction section.

src/uniprotkb/components/entry/GoCam.tsx
```diff
@@ -0,0 +1,104 @@
...
+    return null;
+  }
+
+  return (
+    <Tabs>
+      <Tab title="GO-CAM">
+        <div className={styles['go-cam-container']}>
+          <TreeSelect
+            data={goCamItems}
+            label="Select GO-CAM model"
```
aurel-l, line 84: Here's it's inconsistent with the ComplexViewer. What do we prefer?
- Display name in the label of the TreeSelect:
  - ComplexViewer: no change
  - GoCam: Display name in label here + hide the `gocam-title` part through CSS
- or, Display name below the TreeSelect:
  - ComplexViewer: Add some kind of heading/text below the TreeSelect + have a fixed label "Select IntAct complex"
  - GoCam: no change

dlrice: Had a chat with @minjoonkim and went with relying on the select to show which model/complex is selected in both instances.

dlrice:
I've fixed the lazy loading/duplicate tabs issues and also pushed as go ribbon and go-cam sharing tabs for now. The tabs changes are all in one commit so easy to revert if needed.

src/uniprotkb/components/entry/GoRibbon.tsx
```diff
@@ -255,8 +255,7 @@ const GoRibbon = ({
   );
 
   return (
-    <div className="GoRibbon">
-      <h3 data-article-id="gene_ontology">GO annotations</h3>
+    <div className="GoRibbon" data-article-id="gene_ontology">
```
aurel-l, line 258: I think this is this here causing the extra "i" below the whole section, as the "i" will be added inline after the content, and here the content is a block section containing the whole go-ribbon section.
We'd need to find somewhere else to attach it to, that would be inline and not already clickable (so maybe not the tab label itself). Maybe a bit of extra text at the top of the section?

dlrice: I followed your suggestion added a bit of a preamble to both of the GO tabs and attached the `data-article-id` to this.

aurel-l (approved):

src/uniprotkb/components/entry/__tests__/InteractionSection.spec.tsx
```diff
@@ -4,7 +4,8 @@ import customRender from '../../../../shared/__test-helpers__/customRender';
 
 import data, { dataWithNoXref } from './__mocks__/interactionComplexesData';
 
-describe('InteractionSection', () => {
+// TODO: unskip
```
aurel-l, line 7: Is it meant to stay around still?

dlrice: Good catch, it needs to be fixed.

dlrice:
- [x] Fixed message when no models are available
- [x] If we don't have any data for the GO annotations tab, don't display GO-CAM tab
- [x] Update interaction section tests as we now use native select

src/uniprotkb/components/entry/__tests__/InteractionSection.spec.tsx
dlrice: Updated

aurel-l (approved):

src/uniprotkb/components/protein-data-views/GoCamViz.tsx
```diff
@@ -0,0 +1,27 @@
...
+type Props = {
+  id: string;
+};
+
+const GoCamViz = ({ id }: Props) => {
+  const [defined, setDefined] = useState(false);
+  useEffect(() => {
+    goCamVizLoader.defineCustomElements();
+    customElements.whenDefined('wc-gocam-viz').then(() => {
+      setDefined(true);
```
aurel-l, line 16: Async state-setting logic within a useEffect is something to watch out for. Their could be the case of the promise resolving after the component has unmounted, and then it'll try to set a state for a component that doesn't exist anymore.
Maybe use `useSafeState` instead of `useState`.
Have a look at the logic in `useCustomElement`, which internally uses `useSafeState` to do similar things that you're doing here. `useSafeState` is basically a wrapper to make sure that there is no set state when the component is unmounted.

src/uniprotkb/components/entry/GoCam.tsx
```diff
@@ -0,0 +1,162 @@
...
+          concurrency: heuristic.concurrency,
+        });
+        setUniprotGoCamIds(
+          results
+            .filter(({ data }) => isUniprotCurated(data))
+            .map(({ id }) => id)
+        );
+      }
+    }
+    fetchGoCamModels();
```
aurel-l, line 83: Same comment as in `GoCamViz.tsx` regarding state setting in async logic. use `useSafeState`.

Also use one single AbortController both for each fetch and for p-map (look for `signal` in the documentation https://www.npmjs.com/package/p-map ). Look in the codebase for `controller.abort()` to see example usage.
This will also to stop current and subsequent fetches if the user changes page or switch the tab back

src/uniprotkb/components/entry/GoCam.tsx
```diff
@@ -0,0 +1,162 @@
...
+  }, [goCamIdToItem]);
+
+  useEffect(() => {
+    if (uniprotGoCamIds?.[0]) {
+      setSelectedId(uniprotGoCamIds[0]);
+    }
+  }, [uniprotGoCamIds]);
```
aurel-l, line 90: This bit of logic might be able to live directly in the previous useEffect, at the same time as `setUniprotGoCamIds`, to avoid one re-render. Would require using `useSafeState` for `selectedId` though

aurel-l (changes requested):

aurel-l:
Playing with the UI I've just realised that most of the content in the panel on the right is links, and most of those links are to other websites. I don't really like that we have a fairly big section of UI that behaves completely differently than the rest of the website

Mouse cursor but no UI indicator of a link make the user think it might be a button with an action on the current page. It would be better that all link use the bold/blue style that we use for links elsewhere. That would probably require a change of style for the clicked section otherwise the links would blend into the selected background (change background colour? Just have a border to show selection?).

Also, external links not marked as such is a big problem, especially when some of the concepts links do already exist in UniProt (e.g. "part of" or "occurs in" values that look like they could be UniProt subcell locations or keywords).
We do use `mask-image` to inject the external link image in some places, so maybe that could be reused here?

I'm not entirely sure what is doable or not though, given that it's all within custom elements that will be a bit more complicated to style.

src/uniprotkb/components/entry/GoCam.tsx
```diff
@@ -0,0 +1,226 @@
...
+    () => getGoCamStructures(allGoCamIdsResponse.data),
+    [allGoCamIdsResponse.data]
+  );
+
+  useEffect(() => {
+    // eslint-disable-next-line import/no-named-as-default-member
+    const cancelTokenSource = axios.CancelToken.source();
+    const abortController = new AbortController();
+    const { signal } = abortController;
+    signal.addEventListener('abort', () => {
```
aurel-l, line 93: I was wondering why that was needed, but I see the axios cancel token logic above. It's one of these cases where using plain fetch would be easier as AbortController fits well into the native fetch logic.
No need to change here, just a comment

aurel-l (approved):

src/uniprotkb/components/entry/GoCam.tsx
```diff
@@ -135,12 +135,28 @@ const GoCam = ({ primaryAccession }: Props) => {
...
     };
   }, [goCamIdToItem, primaryAccession, setGoCamIdToNode]);
+
   useEffect(() => {
     if (uniprotGoCamIds?.[0]) {
       setSelectedId(uniprotGoCamIds[0]);
     }
   }, [uniprotGoCamIds]);
 
+  const showGoCamViz =
```
aurel-l, line 145: A quick visual check made me find out that there's an issue where if the screen is small and deactivates this boolean we will see the "UniProt curated GO-CAM models have not been found for this entry within the Gene Ontology knowledgebase." sentence. I will stop reviewing this for today, just letting you know if you want to have a look tomorrow morning.
Not sure what should happen, we have multiple options:
- Display useful fallback without the visualisation (just a link to GO-Cam website with the models?)
- Don't display the tab at all
- Display a click-to-load button (see structure section logic below in the page)
- Something else?

aurel-l (approved):

PR #825, opened 2024-10-29
Entries plot

dlrice:
## Note
This should only deployed once backend have populated the table with all of the historical releases.

## Purpose

[Create plot of growth of swiss-prot / trembl](https://www.ebi.ac.uk/panda/jira/browse/TRM-30433)

## Approach

- Use D3 and copy approach used in the other stats plots.
- Created new `DefaultDict` util fn à la [python's](https://docs.python.org/3/library/collections.html#collections.defaultdict) to facilitate accumulating values when using a js object.
- Updated the PR checklist.

## Testing

Wrote a unit test for the new `DefaultDict` util fn.

## Checklist


src/uniprotkb/components/statistics/styles/historical-release-entries-line-plot.module.scss
```diff
@@ -0,0 +1,9 @@
+@import 'franklin-sites/src/styles/colours';
+
+.line-plot {
+  path.line {
+    fill: none;
+    stroke: $colour-weldon-blue;
```
aurel-l, line 6: Would be good to use the global custom properties for colours here, instead of having to import from franklin.

src/uniprotkb/components/statistics/HistoricalReleasesEntriesLinePlot.tsx
```diff
@@ -0,0 +1,113 @@
+import { memo, useCallback, useEffect, useRef } from 'react';
+import {
+  select,
+  scaleLinear,
+  axisBottom,
+  axisLeft,
+  line,
+  scaleTime,
+  format,
+  timeFormat,
```
aurel-l, line 10:
```suggestion
  timeFormat,
  timeYear,
```

src/uniprotkb/components/statistics/HistoricalReleasesEntriesLinePlot.tsx
```diff
@@ -0,0 +1,113 @@
...
+type Props = {
+  dateCounts?: DateCount[];
+  bounds?: Bounds;
+};
+
+const HistoricalReleasesEntriesLinePlot = ({ dateCounts, bounds }: Props) => {
+  const svgRef = useRef<SVGSVGElement>(null);
+
+  const renderPlot = useCallback((dateCounts: DateCount[], bounds: Bounds) => {
+    const chart = select(svgRef.current).select('g');
```
aurel-l, line 35:
```suggestion
    const chart = select(svgRef.current).select('g');
    
    const startYear = new Date(bounds.date[0].getFullYear(), 0, 1);
    const endYear = new Date(bounds.date[1].getFullYear() + 1, 0, 1);
```

Modifying the bounds here and not in the original logic that is finding the data bounds because we're within the rendering component and this change of bounds is purely for visual effect

src/uniprotkb/components/statistics/HistoricalReleasesEntriesLinePlot.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    // x-axis
+    const xScale = scaleTime()
+      .domain(bounds.date) // units: Date
```
aurel-l, line 39:
```suggestion
      .domain([startYear, endYear]) // units: Date
```
Corresponding change in order to get the bounds to include the beginning of the first year of data, and the tick for the upcoming year after the current last data point

src/uniprotkb/components/statistics/HistoricalReleasesEntriesLinePlot.tsx
```diff
@@ -0,0 +1,113 @@
...
+      .range([0, width]); // units: pixels
+    chart
+      .select<SVGGElement>('.x-axis')
+      .transition()
+      .duration(1_000)
+      .call(
+        axisBottom(xScale).tickFormat((d: Date | { valueOf(): number }) =>
+          timeFormat('%Y')(new Date(d.valueOf()))
+        )
```
aurel-l, line 48:
```suggestion
        axisBottom(xScale)
          .ticks(timeYear)
          .tickFormat((d: Date | { valueOf(): number }) =>
            timeFormat('%Y')(new Date(d.valueOf()))
          )
```
Here, make sure that we get only 1 tick every year on January 1st

src/uniprotkb/components/statistics/HistoricalReleasesEntriesLinePlot.tsx
```diff
@@ -0,0 +1,113 @@
...
+      );
+
+    // y-axis
+    const yScale = scaleLinear()
+      .domain(bounds.count) // units: count
+      .range([height, 0]); // units: pixels
```
aurel-l, line 54:
```suggestion
      .range([height, 0]) // units: pixels
      .nice();
```
Make sure that we have a bit of margin at the top so that the highest number is not too close to the top and that the last tick (at, or higher than, the highest value) will be displayed.

aurel-l (commented):
Small changes here and there to generate a view that I prefer visually, but it's up to you to decide if you want it or not
Screenshot 2024-11-11 at 18 38 56
This is the resulting view with the changes in this review.
Feel free to just pick and choose the changes that you prefer

src/uniprotkb/components/statistics/HistoricalReleasesEntriesLinePlot.tsx
aurel-l: Also, maybe with the full data we'll have to adjust the ticks even more and do `timeYear.every(2)` for example

dlrice: image

Current plot has every 4 years - shall we just go for that?

package.json
```diff
@@ -111,7 +111,7 @@
     "core-js": "3.38.1",
     "d3": "5.16.0",
     "deep-freeze": "0.0.1",
-    "franklin-sites": "0.0.247",
+    "franklin-sites": "file:.yalc/franklin-sites",
```
aurel-l, line 114: I'm guessing this is not meant to go in the final code for this PR

dlrice: Not sure how that got in there!

aurel-l (approved):

aurel-l (approved):

PR #826, opened 2024-11-06
Add warning message in Async download for heavy queries

Swaathik:
## Purpose
https://www.ebi.ac.uk/panda/jira/browse/TRM-31837

## Approach
Added warning in the second step of Async download.
Added contextual help links for column customisation, download, advance search wherever applicable.

## Testing

## Checklist

### DIFF
--- a/src/tools/async-download/components/AsyncDownloadConfirmation.tsx
+++ b/src/tools/async-download/components/AsyncDownloadConfirmation.tsx
@@ -20,6 +20,7 @@ import { FormParameters } from '../types/asyncDownloadFormParameters';
 import { APIModel } from '../../../shared/types/apiModel';
 import { FacetObject, SearchResults } from '../../../shared/types/results';
 import { SelectedFacet } from '../../../uniprotkb/types/resultsTypes';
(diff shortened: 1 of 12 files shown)


src/tools/async-download/components/AsyncDownloadConfirmation.tsx
```diff
@@ -100,11 +105,36 @@ const AsyncDownloadConfirmation = ({
...
+      {jobParameters.query === '*' ? (
+        <Message className={styles['warning-message']} level="warning">
+          This job will take longer to finish, as your search is against the
+          whole of {namespaceAndToolsLabels[jobParameters.namespace]}, making it
+          a highly resource-intensive process. We kindly request that you
+          download the file once it becomes available.
+        </Message>
+      ) : null}
+      {jobParameters.query !== '*' &&
+      jobParameters.query?.split(' ').length === 1 &&
```
aurel-l, line 123: I think l.123 might not be needed, the most important is l.124 (checking for the absence of ":"). Search for "homo sapiens" is also a free text search that might not return what the user expected.
Unless I'm missing  a use case that this covers?

aurel-l (approved):

PR #828, opened 2024-11-14
Community annotation

dlrice:
## Purpose

[Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-31447)

- [x] Display the submission date with the community annotation info shown on the page, use it to display the most recent first when there are multiple in the same section.
- [x] Point to the publications tab with the correct facets activated for the link at the top of the entry tab with the count of community annotations
- [x] Merge multiple submission into one when the content is the same, make sure to keep displaying all the other information and that visually they are linked together (pubmed id, user orcid, submission date, and link to submission)
- [x] Don't display protein name annotations if the exact same name is already in the entry

## Approach

- Exclude community names already existing in the curated names
- Group all community annotations by annotation
- Use a table to display community annotations, copy existing Table styles

## Testing

Updated and added unit test for new util fn.

## Checklist

### DIFF
--- a/src/shared/components/table/styles/table.module.scss
+++ b/src/shared/components/table/styles/table.module.scss
@@ -1,3 +1,11 @@
+$table__header-text: #393b42;
+$table__header-background: #fff;
+$table__even: #fff;
+$table__odd: #e4e8eb;
+$table__header-text: #393b42;
+$table_border: #c2c4c4;
+$table_extra-content-background: #f1f1f1;
+
 .container {
   width: calc(100% + 2em);
   padding-inline: 1em;
@@ -9,13 +17,16 @@
   padding: 0;
   margin: 1em 0;
   border-spacing: 0;
-  border-block-end: 1px solid #c2c4c4;
+  border-block-end: 1px solid $table_border;
 
   & th {
     text-align: left;
     vertical-align: middle;
-    background-color: var(--protvista-datable__header-background, #fff);
-    color: var(--protvista-datable__header-text, #393b42);
+    background-color: var(
+      --table__header-background,
+      $table__header-background
+    );
+    color: var(--table__header-text, $table__header-text);
     text-overflow: ellipsis;
     text-transform: uppercase;
   }
@@ -23,7 +34,7 @@
   & td,
   & th {
     padding: 0.2rem 1ch;
-    border-block-start: 1px solid #c2c4c4;
+    border-block-start: 1px solid $table_border;
   }
 
   & thead :global(.dropdown) > button {
@@ -36,10 +47,10 @@
 }
 
 .row {
-  background-color: var(--protvista-datatable__even, #fff);
+  background-color: var(--table__even, #table__even);
 
   &.odd {
-    background-color: var(--protvista-datatable__odd, #e4e8eb);
+    background-color: var(--table__odd, $table__odd);
   }
 
   &:hover {
@@ -47,7 +58,7 @@
     /* => keep groups of rows highlighted in sync */
     &,
     & + .extra-content {
-      background-color: var(--protvista-datatable__hover, #f1f1f1);
+      background-color: var(--table__hover, $table_extra-content-background);
     }
   }
 
@@ -69,7 +80,7 @@
   }
 
   &.extra-content td {
-    border-color: #c2c4c455;
+    border-color: $table_border;
   }
 }
 
@@ -78,5 +89,5 @@
 /* Select a row that has a next .extra-content sibling that is hovered */
 /* => keep groups of rows highlighted in sync */
 .row:has(+ .extra-content:hover) {
-  background-color: var(--protvista-datatable__hover, #f1f1f1);
+  background-color: var(--table__hover, $table_extra-content-background);
 }
--- a/src/shared/components/table/styles/table.module.scss
+++ b/src/shared/components/table/styles/table.module.scss
@@ -47,7 +47,7 @@ $table_extra-content-background: #f1f1f1;
 }
 
 .row {
-  background-color: var(--table__even, #table__even);
+  background-color: var(--table__even, $table__even);
 
   &.odd {
     background-color: var(--table__odd, $table__odd);
--- a/src/shared/components/table/styles/table.module.scss
+++ b/src/shared/components/table/styles/table.module.scss
@@ -4,7 +4,7 @@ $table__even: #fff;
 $table__odd: #e4e8eb;
 $table__header-text: #393b42;
 $table_border: #c2c4c4;
-$table_extra-content-background: #f1f1f1;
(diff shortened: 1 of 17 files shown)


src/shared/components/table/styles/table.module.scss
```diff
@@ -1,3 +1,11 @@
+$table__header-text: #393b42;
+$table__header-background: #fff;
+$table__even: #fff;
+$table__odd: #e4e8eb;
+$table__header-text: #393b42;
+$table_border: #c2c4c4;
+$table_hover: #f1f1f1;
```
aurel-l, line 7: Do we really want to keep using SASS variables on new or touched code?

If the point is to reuse the same colours throughout this file in different places these could be defined as CSS custom properties on `:root` or on a containing class (for example here `.container`) and used later on.
If we want to define them as internal variables with defaults and then allow them to be overridden that would still work as long as whoever overrides them does so on a lower component within the DOM tree (so it'd be better to define on `:root` to avoid issues)

Also, I see that `$table__header-text` is defined twice

src/shared/utils/utils.tsx
```diff
@@ -108,3 +110,8 @@ export function keysToLowerCase<T>(o: { [k: string]: T } = {}): {
     Object.entries(o).map(([k, v]) => [k.toLowerCase(), v])
   );
 }
+
+export const excludeKeys = (o: Record<Key, unknown>, omit: Key[]) => {
+  const setOmit = new Set(omit);
+  return Object.fromEntries(Object.entries(o).filter(([k]) => !setOmit.has(k)));
+};
```
aurel-l, line 117: We could also use lodash's omit to do this, and it would have the same API as your code here.
Was this on purpose to avoid using a library? I'd say we already use that one throughout the code so it's not like we'll want to remove it anytime soon, and also its types are quite good.

src/shared/components/table/styles/table.module.scss
dlrice: I did this to reuse `$table__header-text` in the community curation style sheet for consistency. Can we do this without sass variables?

src/shared/utils/utils.tsx
dlrice: I thought we wanted to move to implementing simple utils for ourselves in the case we want to do away with lodash one day. I see lodash's `omit` used once in the codebase so I could easily replace with my `excludeKeys` or I could just remove my function - happy with either.

src/uniprotkb/components/entry/CommunityCuration.tsx
```diff
@@ -4,21 +4,171 @@ import {
...
+  section: EntrySection;
+  databaseInfoMaps: DatabaseInfoMaps | null;
+}) => (
+  <Card className={styles['reference-card']}>
+    <h4>
+      {section === EntrySection.NamesAndTaxonomy
+        ? 'Community suggested name: '
+        : ''}
+      {annotation}
+    </h4>
```
aurel-l, line 106: I'm wondering if we should keep the actual annotation out of the heading.
For the example of names it's fine, but when looking at functions it can get quite big and feel a bit weird. See examples (O94966 and A6NCS4):
Screenshot 2024-11-15 at 18 04 38
Screenshot 2024-11-15 at 18 04 55

Should we have "Community suggested name:" for the names and taxonomy section and "Community annotation:" for anything else? And then have the actual annotation in a paragraph

src/uniprotkb/components/entry/CommunityCuration.tsx
```diff
@@ -4,21 +4,171 @@ import {
...
+            if (
+              !a.communityAnnotation?.submissionDate ||
+              !b.communityAnnotation?.submissionDate
+            ) {
+              return 1;
+            }
+            return a.communityAnnotation.submissionDate.localeCompare(
+              b.communityAnnotation.submissionDate
+            );
+          })
```
aurel-l, line 133: Are we sure this works as expected? (See first screenshot in this other comment https://github.com/ebi-uniprot/uniprot-website/pull/828/files#r1844271314).

It might be good to extract the function to:
- Give it a name (and potentially comment the function) that conveys the intent
- Test it
- (Marginally) improve performance by defining it only once instead of at every render

Also, this is a mutation of the props passed to the component, so a bit of a red flag in React world. Instead I'd usually do `Array.from(list).sort()` to first copy then mutate-sort the array, but we could also start using the fairly recently added (ES2023) non-mutating methods and in this case [use `toSorted` instead](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted).

src/shared/config/externalUrls.ts
```diff
@@ -4,6 +4,7 @@ import { stringifyUrl } from '../utils/url';
 import { FileFormat } from '../types/resultsDownload';
 
 const IntActBase = '//www.ebi.ac.uk/intact/';
+const CommunityAnnotation = 'https://community.uniprot.org/bbsub';
```
aurel-l, line 7: Changing the pathname here from `cgi-bin` to `bbsub` makes the logic checking the number of results fail, or in this case think that there are no annotation so display no link, because the custom response headers with the count are not returned.

It's a bit inconsistent from the server, because this one redirects to bbsub when following the link but would not work for `CommunityCurationAdd` 🤷🏽

src/shared/utils/utils.tsx
aurel-l: Yep, if that's made on purpose it's fine, and the logic is fairly encapsulated so that's alright.

src/uniprotkb/components/entry/CommunityCuration.tsx
dlrice: Nice suggestions - implemented and now looks like this:
Screenshot 2024-11-18 at 11 40 26
Screenshot 2024-11-18 at 11 40 12
Screenshot 2024-11-18 at 11 39 45

src/shared/utils/utils.tsx
dlrice: Replaced the other instance of `omit` with the new util fn too.

src/uniprotkb/components/entry/CommunityCuration.tsx
dlrice: Good call to extract into a separate function as wasn't working as expected when looking at the tests.

src/shared/config/externalUrls.ts
dlrice: Reverted the currently used urls.

src/shared/components/table/styles/table.module.scss
dlrice: Done as suggested

aurel-l (approved):

PR #839, opened 2025-01-23
ESLint upgrade

Swaathik:
## Purpose
Upgrade ESLint to v9

## Approach
Create flat config 
Remove airbnb plugin as it is not updated to v9 yet
Use the recommended ones 
Abandoned eslint-custom-rules (still in the repo though)

## Testing

## Checklist


src/uniprotkb/components/entry/Entry.tsx
```diff
@@ -514,7 +514,7 @@ const Entry = () => {
                   >
                     Add a publication
                   </a>
-                  {/* eslint-disable-next-line react/jsx-no-target-blank */}
+                  {}
```
dlrice, line 517:
```suggestion
```

src/uniprotkb/components/entry/tabs/genomic-coordinates/CoordinateTable.tsx
```diff
@@ -303,7 +304,6 @@ const CoordinateExtraContent = ({
             <th>Exon ID</th>
             <th>UniProtKB Isoform</th>
             <th>Position(s)</th>
-            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dlrice, line 306: I think we could just keep the exception at the exact line rather than for the whole file.

dlrice (approved):

PR #842, opened 2025-01-30
Use dnd kit for column select

dlrice:
## Purpose
The drag-and-drop UI for customise columns was not working as expected:

- vertical layout instead of horizontal
- weird behaviour when drag-and-dropping

[Jira](https://www.ebi.ac.uk/panda/jira/browse/TRM-32258)

## Approach

- Replace @hello-pangea/dnd with @dnd-kit
- Adapt code for new api
- A little refactoring here and there

## Testing
Updated

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "jest-coverage-ratchet": "jest-coverage-ratchet",
     "coverage-bump": "npm-run-all test:unit jest-coverage-ratchet",
     "prettier": "prettier src --write",
-    "test": "npm-run-all --continue-on-error test:lint test:types test:unit",
+    "test": "npm-run-all test:lint test:types test:unit",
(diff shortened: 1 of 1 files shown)


src/shared/utils/utils.tsx
```diff
@@ -20,15 +20,14 @@ export const pluralise = (singular: string, count: number, plural?: string) => {
...
+  const newArray = array.slice();
+  newArray.splice(
+    to < 0 ? newArray.length + to : to,
+    0,
+    newArray.splice(from, 1)[0]
+  );
+  return newArray;
 }
 
 export function removeItemFromList<T>(list: T[], index: number) {
```
Swaathik, line 33: Shall we rename this function from list to array for consistency?

Swaathik (approved):

PR #857, opened 2025-04-24
Fix FTP links on UniProtKB landing page

matthiasblum:
## Purpose

The FTP links on the [UniProtKB landing page](https://www.uniprot.org/uniprotkb) are broken because they point to the `/pub/databases/uniprot/knowledgebase/complete` directory on the EMBL-EBI FTP server, but `complete` doesn't exist (files are directly under `knowledgebase`.

## Approach

The URL config module is updated to remove the `/complete` leaf.

## Testing

Locally build the project. FTP links are working as expected.

## Checklist

### DIFF
--- a/src/shared/config/ftpUrls.ts
+++ b/src/shared/config/ftpUrls.ts
@@ -8,14 +8,14 @@ const ftpUniProt = 'https://ftp.ebi.ac.uk/pub/databases/uniprot/';
 
 const ftpUrls = {
   uniprot: ftpUniProt,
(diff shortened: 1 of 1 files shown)


Swaathik:
Thanks for the PR, @matthiasblum. There is a temporary redirection in place due to delay. We will revert the FTP URLs back to UniProt's FTP soon instead of EBI's . When we do that, we will have to revert your change as well, making the merge unnecessary. We will make sure to check the links next time when we have to do something similar. Thank you

PR #870, opened 2025-06-27
ABB links

aurel-l:
## Purpose
[TRM-32844](https://embl.atlassian.net/browse/TRM-32844) add missing links to new ABB data

## Approach
Refactor existing code to be easier to add new logic.
Add new exceptions and link patterns.

## Testing
Added new tests, including checking link presence or absence and also snapshot testing

Examples:
- MGI link (fixed from previous links) /uniprotkb/P0DW27/publications?facets=types%3A1
- IEDB link (new) /uniprotkb/P05067/publications?facets=types%3A1
- IMPC link (new) /uniprotkb/Q9Z2L7/publications?facets=types%3A1
- Community link but from "Anonymous" (rendering unchanged, just to see it's still working with the new logic) /uniprotkb/A0A0D5YNH1/publications?facets=types%3A0

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/tabs/Publications.tsx
+++ b/src/uniprotkb/components/entry/tabs/Publications.tsx
@@ -81,25 +81,21 @@ export const PublicationSource = ({
   }
 
   /* URI link exceptions: */
-  // 'GeneRif' is indexed as 'GeneRIF' in the configuration endpoint
   if (source.name === 'GeneRif') {
+    // 'GeneRif' is indexed as 'GeneRIF' in the configuration endpoint
     uriLink = databaseInfoMaps?.databaseToDatabaseInfo.GeneRIF?.uriLink;
-  }
-  // 'MGI' ID is missing the initial "MGI:"
-  if (source.name === 'MGI') {
+  } else if (source.name === 'MGI') {
+    // 'MGI' ID is missing the initial "MGI:"
     uriLink = uriLink?.replace('%id', 'MGI:%id');
-  }
-  // 'IEDB' is not in the configuration endpoint
-  if (source.name === 'IEDB') {
+  } else if (source.name === 'IEDB') {
+    // 'IEDB' is not in the configuration endpoint
     uriLink = 'https://iedb.org/antigen/UNIPROT:%id';
-  }
-  // 'IMPC' is not in the configuration endpoint
-  if (source.name === 'IMPC') {
+  } else if (source.name === 'IMPC') {
+    // 'IMPC' is not in the configuration endpoint
     uriLink = 'https://www.mousephenotype.org/data/genes/%id';
   }
-
-  // 'GAD' not in the configuration endpoint, but don't inject a link
-  // This DB is dead so we only display the accession, but no link to point to
+  /* Some DBs are dead and not in the configuration endpoint (eg GAD) and so we
+  only display the accession but not the link. No need to handle them here */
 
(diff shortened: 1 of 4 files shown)


src/uniprotkb/components/entry/tabs/Publications.tsx
```diff
@@ -24,22 +24,110 @@ import { processUrlTemplate } from '../../../../shared/utils/xrefs';
...
+  // 'MGI' ID is missing the initial "MGI:"
+  if (source.name === 'MGI') {
+    uriLink = uriLink?.replace('%id', 'MGI:%id');
+  }
+  // 'IEDB' is not in the configuration endpoint
+  if (source.name === 'IEDB') {
+    uriLink = 'https://iedb.org/antigen/UNIPROT:%id';
+  }
+  // 'IMPC' is not in the configuration endpoint
+  if (source.name === 'IMPC') {
```
dlrice, line 97:
```suggestion
  else if (source.name === 'MGI') {
    uriLink = uriLink?.replace('%id', 'MGI:%id');
  }
  // 'IEDB' is not in the configuration endpoint
  else if (source.name === 'IEDB') {
    uriLink = 'https://iedb.org/antigen/UNIPROT:%id';
  }
  // 'IMPC' is not in the configuration endpoint
  else if (source.name === 'IMPC') {
```
After looking at the conditions I understand that it's impossible for any two of these if statements to be true at once, but making them `else if` could improve readability by asserting that this will be the case.

src/uniprotkb/components/entry/tabs/Publications.tsx
```diff
@@ -24,22 +24,110 @@ import { processUrlTemplate } from '../../../../shared/utils/xrefs';
...
+          see community submission
+        </ExternalLink>
+        ).
+      </>
+    );
+  }
+
+  /* URI link exceptions: */
+  // 'GeneRif' is indexed as 'GeneRIF' in the configuration endpoint
+  if (source.name === 'GeneRif') {
```
dlrice, line 85: Is there a misconfiguration somewhere along the way (production/backend) which has caused this case mismatch? Is the proposed code change a temporary fix or a more permanent edge case handler?

src/uniprotkb/components/entry/tabs/Publications.tsx
```diff
@@ -24,22 +24,110 @@ import { processUrlTemplate } from '../../../../shared/utils/xrefs';
...
+  // 'IEDB' is not in the configuration endpoint
+  if (source.name === 'IEDB') {
+    uriLink = 'https://iedb.org/antigen/UNIPROT:%id';
+  }
+  // 'IMPC' is not in the configuration endpoint
+  if (source.name === 'IMPC') {
+    uriLink = 'https://www.mousephenotype.org/data/genes/%id';
+  }
+
+  // 'GAD' not in the configuration endpoint, but don't inject a link
```
dlrice, line 101: Does the `'GAD' not in...` line apply to L104-L110? I was looking for `GAD` to appear somewhere, so it looks like its more general? Maybe something along the lines of "Some DBs are dead and not in the configuration endpoint (eg GAD) and so we only display the accession but not the link."

aurel-l: I've changed a bit this section as suggested

dlrice (approved):

PR #871, opened 2025-07-07
AGR homology

dlrice:
## Purpose

[Integrate AGR orthology data](https://embl.atlassian.net/browse/TRM-32833?atlOrigin=eyJpIjoiODFmNzA2MzAyYzhjNDViYjk2NTI3YzE3NTczZGE1MmQiLCJwIjoiaiJ9)

## Approach

- Place within similar proteins section as own set of tabs, below uniref clusters
- Fetch data from AGR
- A lot of logic has been lifted from the AGR codebase
- Show data with `TableFromData`
- Each method gets its own column
- Create and use `WithTooltip` to facilitate the addition of tooltips on the table headers
- Adapt tooltip logic as this was creating issues with hovering over both the tooltip and the tooltip target (method header which is rotated at an angle)
- `TableFromData` now alphabetically sorts filtering options, if present

## Testing

## Checklist

- [x] If needed, the changes have been previewed (eg on wwwdev) by all interested parties.


aurel-l:
Adding a commit to handle the case where there is no AGR xref, e.g. O42043.

Do you have any other specific examples to check, for example orthologs but no paralogs, or vice-versa?

aurel-l (approved):

PR #880, opened 2025-08-29
Add search space information in BLAST form if taxonomy filter is selected

Swaathik:
## Purpose
https://embl.atlassian.net/browse/TRM-32713

## Approach
Show information about the sequence numbers present if taxonomy filter is selected. In case of no sequences for the selected combination, show error message and disable submission

## Testing

## Checklist

### DIFF
--- a/src/jobs/blast/components/BlastForm.tsx
+++ b/src/jobs/blast/components/BlastForm.tsx
@@ -3,6 +3,7 @@ import '../../styles/ToolsForm.scss';
 import cn from 'classnames';
 import {
   Chip,
+  Message,
   PageIntro,
   sequenceProcessor,
   SequenceSubmission,
@@ -15,9 +16,11 @@ import {
   useEffect,
   useReducer,
   useRef,
+  useState,
 } from 'react';
 import { useHistory } from 'react-router-dom';
 import { sleep } from 'timing-functions';
+import joinUrl from 'url-join';
 
 import { Location, LocationToPath } from '../../../app/config/urls';
 import { addMessage } from '../../../messages/state/messagesActions';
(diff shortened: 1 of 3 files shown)


src/jobs/blast/components/BlastForm.tsx
```diff
@@ -409,6 +460,30 @@ const BlastForm = ({ initialFormValues }: Props) => {
...
+            {searchSpaceTotal !== Infinity &&
+            (formValues[BlastFields.taxons]?.selected as SelectedTaxon[])
+              ?.length ? (
+              <Message
+                level={searchSpaceTotal === 0 ? 'failure' : 'info'}
+                className="taxonomy-selection-message"
+              >
+                {searchSpaceTotal === 0 ? (
+                  <>
+                    <strong>No sequences found for your selection</strong>
```
dlrice, line 472: A suggestion for reporting results to users:

```
Search space: 123,322 protein sequences
This is a filtered subset of the target database UniProtKB reference proteomes + Swiss-Prot based on your taxonomy filters.

--

Search space has no protein sequences
Filtering the target database UniProtKB reference proteomes + Swiss-Prot by the selected 2 taxa returns no protein sequences. Adjust the target database or taxonomy filters to continue.
```
Suggested this to:

- Repeat their DB choice/taxonomy counts to remind them
- "Match" makes me think of BLAST results so I thought best to avoid this so users don't think we've somehow run the job already and we're telling them the results.
- We can use the `pluralise` fn for sequence/s, taxonomy/taxa
- We can use `formatLargeNumber`/`LongNumber` for the sequence count

src/jobs/blast/components/BlastForm.tsx
```diff
@@ -149,6 +157,49 @@ const BlastForm = ({ initialFormValues }: Props) => {
...
+            break;
+          default:
+            break;
+        }
+        const url = stringifyUrl(baseUrl, {
+          format: fileFormatToUrlParameter[FileFormat.fasta],
+          includeIsoform: true,
+          query,
+        });
+        const headResponse = await fetch(url, {
```
dlrice, line 186: We could use `fetchData` and have a clean up for the `useEffect` which aborts eg:

```
    return () => {
      abortController.abort();
    };
```

`fetchGoCamModels` has an example of this.

src/jobs/blast/components/BlastForm.tsx
```diff
@@ -149,6 +157,49 @@ const BlastForm = ({ initialFormValues }: Props) => {
...
   );
 
+  useEffect(() => {
+    const fetchData = async () => {
+      if (!excludeTaxonField && formValues[BlastFields.taxons].selected) {
+        const baseUrl = joinUrl(apiPrefix, 'uniprotkb', 'search');
+        let query = `(${(formValues[BlastFields.taxons].selected as SelectedTaxon[]).map((taxon) => `organism_id:${taxon.id}`).join(' OR ')})`;
+        switch (formValues[BlastFields.database].selected) {
+          case 'uniprotkb_refprotswissprot':
+            query += ' AND (keyword:KW-1185 AND reviewed:true)';
```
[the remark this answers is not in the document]
dlrice: In this case, that is desired as that dataset is the union of those two sets.

PR #931, opened 2026-03-19
Prefill contact form for third-party use

Swaathik:
## Purpose
https://embl.atlassian.net/browse/TRM-33655

## Approach
Listen to URL search params - `subject`, `message` and `history`. 
Expects `history` to be compressed plain text
Decode history and add it to the context textarea

## Testing
Test the new URL params

## Checklist


src/types/pako.d.ts
dlrice, line 1: This can be removed now with your recent change.

dlrice (approved):

PR #951, opened 2026-05-18
Protoeme schema

Swaathik:
## Purpose
https://embl.atlassian.net/browse/TRM-33723

## Approach
Updated the proteome types in the following pages:

- [x] Landing page
- [x] Facets (sort and decorate the labels)
- [x] Entry page
- [x] Fix Pan proteome
- [ ] Use new icon for Non reference proteome

## Testing

## Checklist


src/proteomes/config/ProteomesColumnConfiguration.tsx
```diff
@@ -159,14 +159,14 @@ ProteomesColumnConfiguration.set(ProteomesColumn.genomeRepresentation, {
 ProteomesColumnConfiguration.set(ProteomesColumn.proteinCount, {
   ...getLabelAndTooltip(
     'Protein count',
-    'Number of protein entries associated with this proteome: UniProtKB entries for regular proteomes or UniParc entries for redundant proteomes'
+    'Number of protein entries associated with this proteome: UniProtKB entries for reference proteomes or UniParc entries for non reference proteomes'
```
dlrice: It seems we should mention excluded proteomes here too if we look at the logic of the below

```
    const shouldPointToUniParc =
      proteomeType === 'Excluded' || proteomeType === 'Non Reference proteome';
```

src/shared/components/results/ResultsFacets.tsx
```diff
@@ -66,17 +70,48 @@ const ResultsFacets = memo<Props>(({ dataApiObject, namespaceOverride }) => {
...
+            ?.slice()
+            .sort((a, b) => {
+              if (facet.name !== 'proteome_type') {
+                return 0;
+              }
+              const order = [
+                ProteomeFacetValue.REFERENCE,
+                ProteomeFacetValue.NON_REFERENCE,
+                ProteomeFacetValue.EXCLUDED,
+              ];
```
dlrice: This can come outside of the sort's function arg.

src/shared/components/results/ResultsFacets.tsx
```diff
@@ -66,17 +70,48 @@ const ResultsFacets = memo<Props>(({ dataApiObject, namespaceOverride }) => {
...
-                ? facetValue.value.toUpperCase()
-                : facetValue.value,
-            label:
-              facet.name !== 'proteome'
-                ? getDecoratedFacetLabel(facetValue)
-                : null,
-          })),
+          // Sort proteome_type facet values in a specific order
+          values: facet.values
+            ?.slice()
```
dlrice: I think this could be a little neater if we `map` (which will create a copy, which is what the `slice` is for I think) first, then sort.

src/shared/components/results/ResultsFacets.tsx
```diff
@@ -65,53 +71,42 @@ const ResultsFacets = memo<Props>(({ dataApiObject, namespaceOverride }) => {
...
+                  ],
+              }),
+            }))
+            .sort(
+              (a, b) =>
+                proteomeTypeOrder.indexOf(a.value as ProteomeFacetValue) -
+                proteomeTypeOrder.indexOf(b.value as ProteomeFacetValue)
+            ),
+        };
+      case 'proteome': // uppercase value until backend returns it consistently
```
dlrice, line 103: Wasn't sure what the comment means as it's actually lowercase? Does this need a TODO?

Swaathik: I tried to find the use case but was unsuccessful

dlrice: Ah I see now, the comment applies to L107. Maybe backend would know which endpoint provides that facet.

src/proteomes/components/entry/Overview.tsx
```diff
@@ -51,10 +52,14 @@ const Overview = ({ data }: { data: ProteomesUIModel }) => {
...
             </span>
           ) : null}
-          {proteomeType === 'Non Reference proteome' ? (
+          {proteomeType === 'Non Reference proteome' &&
+          data.relatedProteomes?.length ? (
             <>
               {' ('}
-              <a href="#similar_proteomes">view similar proteomes</a>)
+              <a href={`#${EntrySection.SimilarProteomes}`}>
+                view similar proteomes
```
dlrice: Is this always going to point at reference proteomes (assuming some exist)?

Swaathik: yes, if there are any related proteomes, the link is shown only for `proteomeType === 'Non Reference proteome'`

Is the condition relevant? 
proteomes/UP000053176
This RP has related proteomes. the proteomeType check is not needed

dlrice (approved):

PR #980, opened 2026-06-25
Structure section refactor

Swaathik:
## Purpose
https://embl.atlassian.net/browse/TRM-33957

https://docs.google.com/document/d/1CnaMItT7LTM6wxIYyFOwCq0E7IABOsvBUmNL2p7yVqk/edit?tab=t.0

## Approach
Covered the immediate steps in AF and 3D beacons tabs.
Pending ones are PDB identifiers and the corresponding links and complex, ligand information as we don't have this info in our REST response

Need to link protvista changes via yalc for testing - https://github.com/ebi-webcomponents/protvista/pull/178

## Testing

Updated test
## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "marked": "18.0.3",
     "marked-gfm-heading-id": "4.1.4",
     "p-map": "7.0.4",
-    "protvista-uniprot": "4.9.1",
+    "protvista-uniprot": "file:.yalc/protvista-uniprot",
     "react": "19.2.6",
     "react-diff-viewer-continued": "4.2.2",
     "react-dom": "19.2.6",
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "marked": "18.0.3",
     "marked-gfm-heading-id": "4.1.4",
     "p-map": "7.0.4",
-    "protvista-uniprot": "4.9.2",
+    "protvista-uniprot": "4.9.3",
     "react": "19.2.6",
     "react-diff-viewer-continued": "4.2.2",
     "react-dom": "19.2.6",
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "marked": "18.0.3",
     "marked-gfm-heading-id": "4.1.4",
     "p-map": "7.0.4",
-    "protvista-uniprot": "file:.yalc/protvista-uniprot",
+    "protvista-uniprot": "4.9.2",
     "react": "19.2.6",
     "react-diff-viewer-continued": "4.2.2",
     "react-dom": "19.2.6",
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "marked": "18.0.3",
     "marked-gfm-heading-id": "4.1.4",
     "p-map": "7.0.4",
-    "protvista-uniprot": "4.9.3",
+    "protvista-uniprot": "4.9.4",
     "react": "19.2.6",
     "react-diff-viewer-continued": "4.2.2",
     "react-dom": "19.2.6",
--- a/src/shared/components/table/TableFromData.tsx
+++ b/src/shared/components/table/TableFromData.tsx
@@ -26,10 +26,15 @@ function TableHeaderFromData<T>({
   selectedValue,
   onFilterChange,
 }: TableHeaderFromDataProps<T>) {
+  const filterChoices = column.filterOptions
+    ? column.filterOptions.map((option) => option.label)
+    : options && options.size > 1
+      ? [...options]
+      : undefined;
   return (
     <th>
       {column.label}
-      {options && options.size > 1 && (
+      {filterChoices && (
         <>
           <br />
           <select
@@ -37,7 +42,7 @@ function TableHeaderFromData<T>({
             value={selectedValue}
             onChange={(e) => onFilterChange(column.id, e.target.value)}
           >
-            {[UNFILTERED_OPTION, ...options].map((option) => (
+            {[UNFILTERED_OPTION, ...filterChoices].map((option) => (
               <option key={option} value={option}>
                 {option}
               </option>
@@ -70,7 +75,14 @@ function filterDatum<T>(
 ): boolean {
   return columns.every((column) => {
     const filterValue = filterValues[column.id];
-    if (filterValue === undefined || !column.getValue) {
+    if (filterValue === undefined) {
+      return true;
+    }
+    if (column.filterOptions) {
+      const option = column.filterOptions.find((o) => o.label === filterValue);
+      return option?.predicate(datum) ?? true;
+    }
+    if (!column.getValue) {
       return true;
     }
     return getCellString(column, datum) === filterValue;
@@ -82,6 +94,7 @@ export type TableFromDataColumn<T> = {
   label: ReactNode;
   render: (datum: T) => ReactNode;
   getValue?: (datum: T) => string | number | null | undefined;
+  filterOptions?: { label: string; predicate: (datum: T) => boolean }[];
 };
 
 type Props<T> = HTMLAttributes<HTMLTableElement> & {
--- a/src/uniparc/components/entry/UniParcStructureView.tsx
+++ b/src/uniparc/components/entry/UniParcStructureView.tsx
@@ -4,6 +4,7 @@ import { lazy } from 'react';
 import { type Sequence } from '../../../shared/types/sequence';
 import { entrySectionToLabel } from '../../config/UniParcSubEntrySectionLabels';
 import SubEntrySection from '../../types/subEntrySection';
+import styles from './styles/uniparc-structure-section.module.scss';
 
 const StructureView = lazy(
   () =>
@@ -26,6 +27,7 @@ const UniParcStructureView = ({ data }: Props) =>
       }
       id={SubEntrySection.Structure}
       data-entry-section
+      className={styles['structure-section']}
     >
       <StructureView checksum={data.md5} isUniProtKBAccession={false} />
     </Card>
--- a/src/uniparc/components/entry/XRefsSection.tsx
+++ b/src/uniparc/components/entry/XRefsSection.tsx
@@ -1,4 +1,4 @@
-import './styles/XRefsSection.scss';
+import './styles/XrefsSection.scss';
 
 import { Card, DataTableWithLoader, Loader } from 'franklin-sites';
 import { useMemo } from 'react';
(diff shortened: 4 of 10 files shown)


dlrice (commented):
The tab split reads really well and the declarative `structureTabs` table (`belongs` / `hiddenColumns`) is a clean way to drive the per-tab columns. A few things to sort out before merge, one of which will break the CI/Netlify build. I've grouped them by priority and included exact locations and suggested fixes so they're quick to action.

All line references are against `src/uniprotkb/components/protein-data-views/StructureView.tsx` unless stated otherwise.

### 1\. `XRefsSection.tsx` imports the stylesheet with the wrong case

**File:** `src/uniparc/components/entry/XRefsSection.tsx:1`

The import was changed to a lowercase `r`:

import './styles/XrefsSection.scss';   // current

but the file on disk (and in git) is `styles/XRefsSection.scss` (capital `R`), and `main` imported it with the matching capital `R`. macOS is case-insensitive so it resolves locally, but Webpack on Linux (CI and Netlify) is case-sensitive and will fail with `Module not found: can't resolve './styles/XrefsSection.scss'`.

This change also looks unrelated to the structure work — probably an accidental edit.

**Fix** — revert the import to match the actual filename:

import './styles/XRefsSection.scss';

(Equivalently, `git mv` the file to `XrefsSection.scss`, but reverting the import is simpler since nothing else references the lowercase name.)

### 2\. Reuse the franklin-sites `Tabs` / `Tab` component — and it does fit the controlled case

**Where:** the `<div className="tabs">` block, \~lines 315–347 (switching lets us delete the hardcoded `tabs__header__item*` class strings entirely).

We already have a tab component that's used in the very section this is modelled on:

- `src/uniprotkb/components/protein-data-views/SubcellularLocationWithVizView.tsx:235` — `<Tabs bordered>` / `<Tab>`  
- `src/uniprotkb/components/entry/Entry.tsx:841` — `<Tabs active={…}>` (a **controlled** active tab)  
- also `SimilarProteins.tsx`, `UniProtKBStatsTabs.tsx`, `AgrHomology.tsx`

Two reasons to switch:

- **Convention:** avoid hardcoding global class names in components, but this hardcodes `'tabs__header__item'`, `'tabs__header__item--bordered'`, `'tabs__header__item--active'`. If franklin renames those, this breaks silently.  
- **Accessibility:** the component already sets `role="tablist"` / `role="tab"` / `role="tabpanel"`, `aria-controls`, `aria-disabled`, and a generated panel id — the associations the hand-rolled version is missing.

Having looked at franklin's `Tabs` source, it fits here, with **one important detail about its managed (controlled) mode**. There are two modes:

- **Unmanaged** (no `active` prop): franklin owns the selected tab and wires up click, keyboard (Enter/Space) and `tabIndex` on the headers for you. This is the `SubcellularLocationWithVizView` usage.  
- **Managed** (`active` set): *you* own which tab is active, and franklin renders the markup but **deliberately does not attach click / keyboard / `tabIndex`** — the consumer makes the header interactive. `Entry.tsx` uses this mode and puts a router `<Link>` in each tab's `title` to handle activation.

Our case needs **managed** mode, because the active tab is derived from the viewer's `selectedId`. So set `active={activeTab.id}` and forward our own handlers onto each header — any extra props on `<Tab>` are spread onto the header `<div>`. Keep the protvista viewer *above* `<Tabs>` (it's shared and must not unmount on tab switch):

{/\* viewer stays here, outside \<Tabs\> \*/}

\<protvista-uniprot-structure ref={setStructureEl} … noTable /\>

{\!viewerOnly && loading && \<Loader /\>}

{\!viewerOnly && \!loading && structures.length \=== 0 && (

  \<Message level="info"\>No structure information available…\</Message\>

)}

{\!viewerOnly && \!loading && structures.length \> 0 && activeTab && (

  \<Tabs bordered active={activeTab.id}\>

    {visibleTabs.map((tab) \=\> (

      \<Tab

        key={tab.id}

        id={tab.id}

        title={tab.title}

        tabIndex={0}

        onClick={() \=\> handleTabClick(tab.rows)}

        onKeyDown={(event) \=\> {

          if (event.key \=== 'Enter' || event.key \=== ' ') {

            event.preventDefault();

            handleTabClick(tab.rows);

          }

        }}

      \>

        \<p\>{tab.description}\</p\>

        \<TableFromData

          data={tab.rows}

          columns={columns.filter(

            (col) \=\> \!(tab.hiddenColumns as readonly string\[\]).includes(col.id)

          )}

          getRowId={(row) \=\> row.rowKey}

          onRowClick={handleRowClick}

          markBackground={(row) \=\> row.id \=== selectedId}

        /\>

      \</Tab\>

    ))}

  \</Tabs\>

)}

Things that are easy to miss (all confirmed from the source):

- **Managed mode wires no interactivity** — you must forward `onClick`, and for keyboard users `tabIndex={0}` \+ `onKeyDown` (Enter/Space), or the tabs aren't operable at all. (Alternatively, follow `Entry.tsx` and make `title` an interactive element — e.g. a `<button>` — and let it own click \+ keyboard.)  
- **Always pass an explicit `id`** on each `<Tab>` matching the value given to `active`. If `id` is omitted, franklin falls back to the child index (`"0"`, `"1"`…), which won't match `active="pdb"`.  
- **`active` must always match a rendered tab** — franklin `throw`s (`Could not find a tab with the id…`) if it doesn't. Since we only render `<Tabs>` when `activeTab` is defined and build the list from `visibleTabs`, that holds — keep the guard.

Knock-on cleanups this brings for free (so it's less churn than it looks):

- Covers the a11y roles/associations noted above, and franklin uses `data-target` rather than a page-global header `id`, so the id-collision worry in **item 7** goes away.  
- One `<TableFromData>` per `<Tab>` (no `cache`) means the inactive tab's table unmounts, so its filter state resets on its own — that's **item 5** handled.  
- The viewer no longer sits inside a `.tabs` wrapper, so the style bleed that needed the `overflow: hidden` patch (**item 6**) is avoided.

If you'd rather keep the hand-rolled version for this PR, the minimum is to move the `tabs__header__item*` strings behind the shared constants and add proper keyboard focus — but since `<Tabs>` also folds in items 5–7, switching is the better trade.

### 3\. The download link lost its accessible name

**Where:** the `Links` column, lines 256–260.

const downloadLink \= row.downloadUrl ? (

  \<a href={row.downloadUrl} target="\_blank" rel="noopener noreferrer"\>

    \<DownloadIcon width="1em" /\>

  \</a\>

) : null;

It used to carry the visible text "Source", so it had an accessible name; now it's an icon-only link and screen readers announce an empty/URL-only link. (The test had to stop using `getByRole('link', { name: /Source/i })` and drop to a raw `querySelector` for exactly this reason — a good signal the name is gone.)

**Fix** — give it a label:

const downloadLink \= row.downloadUrl ? (

  \<a

    href={row.downloadUrl}

    target="\_blank"

    rel="noopener noreferrer"

    aria-label="Download structure file"

  \>

    \<DownloadIcon width="1em" /\>

  \</a\>

) : null;

### 4\. Foldseek link can render `accession=undefined` for AlphaFold rows

**Where:** the `Links` column, lines 261–273.

The old code guarded the Foldseek link with `primaryAccession && …`; that guard was dropped. For PDB rows this is fine (accession is `row.id`), but for an `AlphaFold DB` \+ canonical row when `primaryAccession` is undefined, the URL becomes `…?accession=undefined&source=AlphaFoldDB`. That path is real: the UniParc structure section renders `<StructureView>` with no `primaryAccession` (`src/uniparc/components/entry/UniParcStructureView.tsx:32`).

**Fix** — compute the accession first and require it for the AlphaFold branch:

const foldseekAccession \=

  row.source \=== PDB\_SOURCE ? row.id : primaryAccession;

const showFoldseek \=

  row.source \=== PDB\_SOURCE ||

  (row.source \=== ALPHAFOLD\_SOURCE &&

    row.isoformIsCanonical \=== true &&

    Boolean(foldseekAccession));

if (showFoldseek) {

  const foldseekSource \= row.source \=== PDB\_SOURCE ? 'PDB' : 'AlphaFoldDB';

  return (

    \<span className={helper\['no-wrap'\]}\>

      {downloadLink}

      {' · '}

      \<ExternalLink

        url={\`https://search.foldseek.com/search?accession=${foldseekAccession}\&source=${foldseekSource}\`}

      \>

        Foldseek

      \</ExternalLink\>

    \</span\>

  );

}

return downloadLink;

While here: this function mixes the string literals `'PDB'` / `'AlphaFold DB'` (e.g. the identifier column, lines 172 & 262\) with the `PDB_SOURCE` / `ALPHAFOLD_SOURCE` constants defined at the top. Worth using the constants throughout for consistency.

### 5\. Table filter state leaks across tabs

`TableFromData` keeps its filter selection in its own state, and the same component instance is reused when switching tabs (no `key`), so a resolution/oligomeric filter set in one tab is still applied when you come back to it. It can also select a `tab.rows[0]` that a stale filter has hidden.

**Fix** — key the table per tab so its filter state resets (\~line 363):

\<TableFromData

  key={activeTab.id}

  data={activeTab.rows}

  …

/\>

(If you adopt franklin `<Tabs>` per item 2, this resolves itself — a per-tab `<TableFromData>` unmounts when you leave the tab, resetting its filter state.)

### 6\. The `.tabs` wrapper renders unconditionally

`<div className="tabs">` (line 315\) wraps the viewer even during loading and in `viewerOnly` mode (visible in the updated `EntryMain` snapshot). Pulling franklin's global `.tabs` layout in unconditionally is what required the `overflow: hidden` patch in `src/uniparc/components/entry/styles/uniparc-structure-section.module.scss`. Only applying the wrapper when the tab header actually renders (or adopting `<Tabs>` per item 2\) removes the need for that workaround.

### 7\. Small correctness / hygiene notes

- **Resolution filter boundary** (lines 228–234): the predicate uses strict `<`, so "Below 4 Å" excludes structures at exactly 4.0 Å. Confirm that matches the curation intent — "below 4" from the curation manual sometimes means ≤ 4\.  
- **Generic DOM ids** (line 321): `id={tab.id}` emits page-global ids like `id="pdb"` and `id="others"` that nothing references (`aria-controls` isn't set) and could collide with other elements or hash anchors. Drop them, or namespace them (e.g. `structure-tab-pdb`).

## Tests

The updated tests are correct and the tab-switch comment in the row-click test is a nice touch, but the new behaviour is mostly uncovered. Suggested additions:

- **Resolution filter:** select "Below 2.5 Å" and assert rows are filtered.  
- **Non-canonical AlphaFold row:** assert **no** Foldseek link appears (the inverse of the new `isoformIsCanonical` branch).  
- **Oligomeric State column** renders its value in the AlphaFoldDB tab.  
- **Tab grouping:** PDB vs AlphaFoldDB vs Others rows land in the right tab, and each tab's description text renders.  
- **UniParc path** (no `primaryAccession`): a canonical AlphaFold row does not produce a Foldseek URL containing `accession=undefined` (covers item 4).

PR #984, opened 2026-07-09
Add zoom hint to feature viewer

Swaathik:
## Purpose

https://embl.atlassian.net/browse/TRM-30968

## Approach

- Move the zoom buttons to the right (closer to the viewer)
- Show key shortcut (ctrl/cmd + scroll) hint on first scroll and discard after 3s in bot feature viewer and variation viewer

## Testing

## Checklist

### DIFF
--- a/src/uniprotkb/components/entry/tabs/ZoomHint.tsx
+++ b/src/uniprotkb/components/entry/tabs/ZoomHint.tsx
@@ -48,7 +48,7 @@ const ZoomHint = ({ children }: { children: ReactNode }) => {
       {children}
       {showZoomHint && (
         <div className={styles['zoom-hint']}>
-          <span>Use [CTRL] + scroll to zoom</span>
+          <span>Use [CTRL/CMD] + scroll to zoom</span>
         </div>
       )}
     </div>
--- a/src/uniprotkb/components/entry/tabs/ZoomHint.tsx
+++ b/src/uniprotkb/components/entry/tabs/ZoomHint.tsx
@@ -1,0 +1,58 @@
+import {
+  type ReactNode,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
+
+import styles from './styles/zoom-hint.module.scss';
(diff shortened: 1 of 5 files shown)


minjoonkim:
couple comments: 
1. the hint only shows `ctrl`, we should also include `/cmd` for mac users
2. should we only trigger the zoom hint when the user doesn't use ctrl/cmd + scroll? So if a user correctly uses it, we don't expose the hint at all. 

Accessibility feature:
role="status"/aria-live="polite" on the badge would announce it to screen-reader users

src/uniprotkb/components/entry/tabs/ZoomHint.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <div
+      className={styles['zoom-hint-container']}
+      onScroll={onScrollOrWheel}
+      onWheel={onScrollOrWheel}
+    >
+      {children}
+      {showZoomHint && (
+        <div className={styles['zoom-hint']}>
+          <span>Use [CTRL] + scroll to zoom</span>
```
minjoonkim: also add in CMD for mac users

src/uniprotkb/components/entry/tabs/ZoomHint.tsx
```diff
@@ -0,0 +1,58 @@
...
+        clearTimeout(zoomHintTimeout.current);
+      }
+    },
+    []
+  );
+
+  return (
+    <div
+      className={styles['zoom-hint-container']}
+      onScroll={onScrollOrWheel}
```
minjoonkim: from claude:
3. onScroll is likely dead code. The scroll event doesn't bubble, and the wrapper div has no overflow, so it never scrolls itself — inner scrolls in the web component won't reach it. onWheel is what actually drives this. Not harmful, but onScroll can probably be dropped to avoid implying it does something.

not sure if this is correct but something to check

minjoonkim (approved):

PR #1001, opened 2026-08-06
Proteome component uniprotkb

Swaathik:
## Purpose

https://embl.atlassian.net/browse/TRM-34258

Need to use dev API to test

## Approach
The logic for query builder and field initialiser was already there for UniParc; removed the namespace === uniparc check so that it is common for all.
Updated the query link in components table for UniProtKB too.

## Testing

Updated snapshots

## Checklist

### DIFF
--- a/src/proteomes/components/entry/Components.tsx
+++ b/src/proteomes/components/entry/Components.tsx
@@ -128,9 +128,7 @@ const Components = ({
                       : Location.UniProtKBResults
                   ],
                 search: stringifyQuery({
-                  query: shouldPointToUniParc
-                    ? `(proteomecomponent:"${id}:${name}")`
-                    : `(proteome:${id}) AND (proteomecomponent:"${name}")`,
+                  query: `(proteomecomponent:"${id}:${name}")`,
                 }),
               }}
             >
--- a/src/proteomes/components/entry/__tests__/__snapshots__/Components.spec.tsx.snap
+++ b/src/proteomes/components/entry/__tests__/__snapshots__/Components.spec.tsx.snap
@@ -113,7 +113,7 @@ exports[`Components view should render 1`] = `
               </td>
               <td>
                 <a
-                  href="/uniprotkb?query=%28proteome%3AUP000005640%29+AND+%28proteomecomponent%3A%22Chromosome+1%22%29"
+                  href="/uniprotkb?query=%28proteomecomponent%3A%22UP000005640%3AChromosome+1%22%29"
                 >
                   13,926
                 </a>
@@ -179,7 +179,7 @@ exports[`Components view should render 1`] = `
               </td>
               <td>
                 <a
-                  href="/uniprotkb?query=%28proteome%3AUP000005640%29+AND+%28proteomecomponent%3A%22Chromosome+2%22%29"
+                  href="/uniprotkb?query=%28proteomecomponent%3A%22UP000005640%3AChromosome+2%22%29"
                 >
                   10,399
                 </a>
@@ -232,7 +232,7 @@ exports[`Components view should render 1`] = `
               </td>
               <td>
                 <a
-                  href="/uniprotkb?query=%28proteome%3AUP000005640%29+AND+%28proteomecomponent%3A%22Chromosome+3%22%29"
+                  href="/uniprotkb?query=%28proteomecomponent%3A%22UP000005640%3AChromosome+3%22%29"
                 >
                   9,040
                 </a>
@@ -285,7 +285,7 @@ exports[`Components view should render 1`] = `
               </td>
               <td>
                 <a
-                  href="/uniprotkb?query=%28proteome%3AUP000005640%29+AND+%28proteomecomponent%3A%22Chromosome+4%22%29"
+                  href="/uniprotkb?query=%28proteomecomponent%3A%22UP000005640%3AChromosome+4%22%29"
                 >
                   5,517
                 </a>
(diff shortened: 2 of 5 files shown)


dlrice (commented):
## Must fix

### 1. `ComponentsButtons.tsx` still emits the legacy query form, which returns 0 on 2026_03

`src/proteomes/components/entry/ComponentsButtons.tsx:67` (and the download panel + the `x-total-results` HEAD request it feeds) still builds `(proteome:ID) AND (proteomecomponent:"name")`. Verified on dev: that form returns **0** for UniProtKB on 2026_03 (the fused form returns 13,931). It also already returns 0 on production today for UniParc-routed proteomes (where the fused form returns 15,487). So once 2026_03 ships, "View proteins", "Download selected (N)" and the download itself break for reference proteomes, while the per-row protein-count links in the same card work — two contradictory query dialects on one page.

Note `ComponentsButtons.spec.tsx:17` and `Download.spec.tsx:134` currently pin the legacy form, so CI actively asserts the behaviour this PR moves away from — they'll need updating together with the fix.

### 2. The orphan-drop deletes the new fused clause in non-builder `stringify` callers

`parse()` stores the fused value as a single `proteomecomponent` query bit with no `proteome` bit, so `stringify(parse(q))` hits the orphan-drop and returns `''` for the PR's own canonical form:

```js
stringify(parse('(proteomecomponent:"UP000005640:chromosome")')) === ''
```

The query builder is rescued by the sibling inputs re-populating both bits on mount, but two callers use `stringify(parse(...))` directly and now silently rewrite user queries (they previously passed no `namespace`, so the drop never applied to them):

- `src/shared/components/results/TaxonomyFacet.tsx:54` — on `?query=(proteomecomponent:"UP000005640:Chromosome 1") AND (organism_id:9606)`, clicking the taxonomy chip to remove just the organism filter navigates to an **empty query** (both filters lost).
- `src/shared/utils/searchSuggestions.ts:41` — suggestion links (e.g. `gene_exact`, taxonomy levels) drop the component scope entirely and display hit counts for the unscoped query.

Suggested fix: in the drop branch of `queryStringProcessor.ts`, recognise a value already in fused form (`/^UP\d{9}:/i`) and emit it unchanged — or teach `parse` to split the fused value into both query bits. Then please add the round-trip test above to `queryStringProcessor.spec.ts` (the existing fusion tests only cover a clause that already holds both bits, so this whole class of regression is invisible to CI).

### 3. Wildcards and pre-quoted values break in the fusion branch

`queryStringProcessor.ts:63` quotes the combined value unconditionally and skips the `needsQuotes` handling the normal branch has. The builder UI explicitly tells users "Type * in the search box to search for all values", but verified on dev: `(proteomecomponent:"UP000005640:*")` → **0**, while unquoted `proteomecomponent:UP000005640*` → 147,520. A user pasting an already-quoted `"chromosome 1"` produces the malformed `(proteomecomponent:"UP000005640:"chromosome 1"")`. Please route the combined value through the same quoting logic as the else-branch.

## Should fix

### 4. Colon in component names truncates on round-trip

`src/query-builder/utils/fieldInitializer.ts:42` does `const [proteomeId, component] = value.split(':')`, keeping only the first two segments, while stringify joins with `:` unconditionally. A name like `Chromosome 1: segment 2` initialises the field to `Chromosome 1`, and TextField's mount effect writes the truncated value back into `queryBits`, so re-submitting queries a different component with no warning. Colons in component names are real (e.g. `src/uniparc/utils/subEntry.ts:67` joins sources with `:`). An `indexOf`-based split fixes it.

### 5. No escaping of interpolated values, and a third copy of the routing predicate

`src/proteomes/components/entry/Components.tsx:131` interpolates `id` and `name` into a quoted Lucene phrase unescaped — a name containing `)` can't be re-parsed by `clauseMatcher` (it degrades to a free-text `All` clause in the builder), and a `"` closes the phrase early and produces a query the API rejects. Since the value is now always quoted, this is the right moment to escape.

Relatedly, `proteomeType === 'Non Reference proteome' || proteomeType === 'Excluded'` now exists in three copies under two names (`Components.tsx:118`, `ComponentsButtons.tsx:62`, `ProteomesColumnConfiguration.tsx:169`), and the fused query template is hand-built in multiple places. Suggest extracting `isUniParcProteome(proteomeType)` next to the `ProteomeType` union, plus a single `proteomeComponentQuery(id, names)` helper used by all three files — that also gives fix #1 a natural home.

### 6. Duplicated orphan predicate (and a redundant re-test)

The orphan predicate in `QueryBuilder.tsx:265` is a character-for-character copy of the drop condition in `queryStringProcessor.ts:68`, so the warning toast and the actual drop can silently drift apart (the warning has no test). Exporting one `isOrphanProteomeComponent(clause)` — or having `stringify` return dropped clauses as diagnostics — keeps them in lockstep. Also, inside the final `else` the `!(proteome && reProteomeIdValue.test(proteome))` re-test is already implied by falling past the `else if` above, so the condition reduces to `if (clause.queryBits.proteomecomponent)`.

## Worth considering

- **Legacy query migration**: two-clause URLs `(proteome:UP…) AND (proteomecomponent:"…")` (bookmarks, help docs, anything pre-2026_03) now return 0 on the new backend, and opening the builder on one rewrites it to `(proteome:UP…)` — the whole proteome — with a warning claiming no proteome ID was supplied. Migrating them in `parse` (fold a proteome clause into an adjacent component clause) would preserve intent. Also note a `NOT`-ed component clause dropped via `continue` loses its negation silently.
- **Test mock is stale**: `src/query-builder/components/__tests__/__mocks__/configureSearchTerms.ts:4383-4401` still has the old non-sibling shape for uniprotkb, so the suite exercises a config that no longer matches 2026_03 — worth regenerating from dev.
- **Deploy ordering**: production is still 2026_02, where the fused form returns 0 (legacy returns 13,926) — this needs to land with, not before, the API release. Probably already the plan given the Jira, but flagging since there's no feature flag.
- Minor: `shouldPointToUniParc` and the resolved pathname in `Components.tsx:118` depend only on `proteomeType` and could be hoisted out of the per-row render.

dlrice (approved):

